PR #357, opened 2015-01-20
Source link to instance definition (fixes #299)

### LINKED ISSUE #299: Add links to source code of typeclass instances implementations
**Original reporter**: *justnika@\<hidden\>*

JPMoresmau:
We get the source information from the class instance Name, and pass it down the chain to the XHTML Layout class. All feedback welcome!

I see the automated tests do not generate the source links, so how can I add a unit test?

### DIFF
--- a/haddock-api/resources/html/Ocean.std-theme/ocean.css
+++ b/haddock-api/resources/html/Ocean.std-theme/ocean.css
@@ -378,6 +378,15 @@ div#style-menu-holder {
   margin: 0 -0.5em 0 0.5em;
 }
 
+#interface td.src .link {
+  float: right;
+  color: #919191;
+  border-left: 1px solid #919191;
+  background: #f0f0f0;
+  padding: 0 0.5em 0.2em;
+  margin: 0 -0.5em 0 0.5em;
+}
+
 #interface span.fixity {
   color: #919191;
   border-left: 1px solid #919191;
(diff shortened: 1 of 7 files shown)


Fuuzetsu:
This is very welcome to see! I would appreciate if you just squashed your two commits into one and gave it a proper description: ‘initial version that works‘ does not look nice in git log and we often cherry-pick quite a bit ;)

>All feedback welcome!

The changes are fairly simple and I don't have any objections. I don't have an answer to your FIXME without investigating.

>I see the automated tests do not generate the source links, so how can I add a unit test?

If you want to do this then you will have to change test runner to accomodate for this. There is currently no mechanism to do so. I see that Travis fails (most likely old cabal, certainly not the fault of the patch) so if you'd like to fix that up and it becomes green with these patches then I'd be much more confident to take it, but if you don't want to fix Travis then that's OK too, it will just have to wait slightly.

Unfortunately I can not test the changes myself for few more days. I can most likely merge Thursday evening unless @sol beats me to it. If you want to make changes to the PR then anytime before then is great and it will make it into GHC 7.10 RC 2 (planned Friday) and GHC 7.10 release but if it's later than that then it's unlikely.

JPMoresmau:
OK, I've squashed the commits and edited the comment.
The FIXME was already in the source, I've just kept it even if its location has changed a bit.

JPMoresmau:
Yes, the travis build uses Cabal 1.18, which causes a warning with GHC 7.10, which causes the build to fail because of -WError. Should we force an install of Cabal 1.22 in travis.yml?

Fuuzetsu:
1.22 sounds fine though if it's a warning only then it may be worthwhile to be more lenient with 1.18 and compile with both

Fuuzetsu:
In 279a662adc83dba2e24bd0b99f7da9d63455f840

bitemyapp:
This is just lovely, thank you for adding this!

:+1: :tada:

PR #358, opened 2015-01-21
Build on Travis

JPMoresmau:
The travis file now installs Cabal 1.22 and ignore the warnings. Also an ambiguity on dataDir has been solved. Note that the build still doesn't pass because the tests fail, but at least it compiles. The test seem to generate "Safe" instead of "Safe-inferred". I have no idea if this is normal in GHC 7.10. Please advise, and I can proceed fixing the tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,14 +10,18 @@ before_install:
  - sudo apt-get update
  - sudo apt-get install ghc-$GHCVER
  - export PATH=/opt/ghc/$GHCVER/bin:$PATH
+ - sudo apt-get install cabal-install-1.22
+ - export PATH=/opt/cabal/1.22/bin:$PATH
+ - cabal --version
  - cd haddock-library
  - cabal install --only-dependencies --enable-tests
  - cabal install doctest
- - cabal configure --enable-tests --ghc-options=-Werror && cabal build && cabal test
+ # --ghc-options=-Werror
+ - cabal configure --enable-tests && cabal build && cabal test
  - doctest -isrc -i$(echo vendor/attoparsec-*) -optP-include -optPdist/build/autogen/cabal_macros.h src/Documentation/Haddock/Parser.hs
  - cabal install
  - cd ..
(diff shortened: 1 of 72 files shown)


JPMoresmau:
Actually, I see that SafeInferred is no longer a constructor of SafeHaskellMode in DynFlags, so I suppose I can replace Safe-Inferred by Safe everywhere.

JPMoresmau:
OK, done that, but there are still failures in the test. The behavior seems to have change in Test.hs, no idea if it's normal or not!

Fuuzetsu:
Now the Travis failure looks legitimate: in first diff it seems that it now renders both signatures in line. This may be due to GHC change but these two:

```
907,917c903
< 			> :: Int</dt
< 		      ><dd class="doc"
< 		      ><p
< 			>This comment applies to both <code
< 			  >r</code
< 			  > and <code
< 			  >s</code
< 			  ></p
< 			></dd
< 		      ><dt class="src"
< 		      ><a name="v:s" class="def"
---
> 			>, <a name="v:s" class="def"
961,965c947
< 			> :: Int</dt
< 		      ><dd class="doc empty"
< 		      >&nbsp;</dd
< 		      ><dt class="src"
< 		      ><a name="v:v" class="def"
---
```

Certainly look worrying. I wonder what caused these. I will merge this PR now but those failures definitely will need investigating.

Fuuzetsu:
Oh, actually, I lie, I will only merge after we resolve the -Werror thing ;)

JPMoresmau:
Mmhh, I can put the -Werror again, but I fear... I'll try tonight and let you know

JPMoresmau:
With -Werror, Attoparsec doesn't compile. So I don't see how we can enable it, until all the libraries we depend on are clean.

Fuuzetsu:
Aha, noted.

JPMoresmau:
Of course now the commit I used to test on Travis as found its way onto this pull request (-:.

JPMoresmau:
Do you want me to try to delete that last commit, then?

Fuuzetsu:
I can cherry pick, it's ok.

Fuuzetsu:
Picked the first 3 commits and I updated the failing tests. Fortunately they were nothing serious, couple of GHC changes hitting us.

Issue #356, opened 2015-01-17
Haddock on windows double the endline on code snippet


Twinside:
With GHC 7.8.4 and Haddock 2.14.3, generating the documentation with code snippets yield double line ending, like for tha package [Rasterific](http://hackage.haskell.org/package/Rasterific-0.4.1/docs/Graphics-Rasterific.html).

The version generated by stackage (on linux) doesn't show these problems: [here](http://www.stackage.org/haddock/nightly-2015-01-16/Rasterific-0.4.1/Graphics-Rasterific.html)

Fuuzetsu:
Why do you think it's the problem of the platform it was built on? I would be much more inclined to suspect a difference because one was built with Haddock 2.16.0 while the other with 2.14.3.

Consider the following:

```
module Foo where

-- |
-- > a
-- > b
-- > c
foo :: ()
foo = ()
```

If you can show that there is a difference between Linux and Windows *both with version 2.16.0* then I am much more inclined to accept this as a platform problem. Also note that I have no access to a Windows machine (and even less so to one with recent GHC and Haddock) so it is a bit difficult to fix and/or test for me even if that does turn out to be the problem.

Twinside:
I thought the problem was associated with windows because I'm mainly on windows and I don't see the problem appearing when the doc is built on other platforms. I was not making a diagnosis here, it might not be related to the platform, but it's still a bug.

Is something related to code snippets has been changed from the last revision available on hackage and the one in the repository? I've tested 2.15.0.2 which gave the same result.

Fuuzetsu:
To the best of my knowledge, the spacing should not happen at all, be it in 2.14, 2.15 or 2.16. It may or may not be a Windows-related problem however, which is why I'm looking for a confirmation. If you are able to give me a simple test case (like the one I posted) and say with confidence that you get different result with same Haddock version on Linux and Windows then that's something to go on. Understandably I would not like to chase after a bug that may or may not be caused by a platform I don't have access too ;)

Just to confirm, the `Foo` module above gives you double spacing with 2.15.0.2 on Windows, right?

Twinside:
I've got more information, I tried the snippet at work, with a GHC 7.6.3 and the last Haddock available on hackage, no problem. Everything is well displayed. Back at home with Haddock 2.15.0.2 and the following snippet:

~~~~{haskell}
module Foo where

-- |
-- > a
-- > b
-- > c
foob :: ()
foob = ()

-- | Bar
--
-- > a
-- > b
[...]

PR #359, opened 2015-01-21
Filtering '\r' on windows, was causing double endline on code blocks.

Twinside:
Related to haskell/haddock#356


Fuuzetsu:
Looks OK, I wonder if we could get away with just one filter in appropriate place. This does confirm that `\r` is a problem at the very least. I will merge it closer to the evening of my time when I can run it.

Thanks for the PR!

Fuuzetsu:
Pulled in with a slightly better (IMHO) commit message, thanks.

Issue #361, opened 2015-01-27
Hoogle export with GHC 7.10 has excessive square brackets.


ndmitchell:
In particular, I see the line:

    [Nothing] :: Maybe a

The square brackets are really odd. This is taken from the GHC 7.10 RC2 file.

Fuuzetsu:
I minimum test case would be very welcome as usual.

ndmitchell:
Grabbing GHC 7.10.0.20150123, I made the file `Foo.hs`:

    module Foo where
    data MaybeD a = NothingD | JustD a

And ran `haddock Foo.hs --hoogle`, but that failed with:

    haddock: internal error: expectJust getPackageDetails

I then tried:

    haddock Foo.hs --hoogle --package-name=foo --package-version=1

Which worked, but produced the Hoogle file containing the two lines:

    [NothingD] :: MaybeD a
    [JustD] :: a -> MaybeD a

This reports itself to be:

    $ haddock --version
    Haddock version 2.16.0, (c) Simon Marlow 2006
    Ported to use the GHC API by David Waern 2006-2008

Which is certainly an incorrect copyright year...

ndmitchell:
Note this is a regression introduced after GHC 7.8.

snoyberg:
I've run into the `expectJust` error message myself, and it looks like it should be a blocker for the GHC 7.10 release. @Fuuzetsu Am I overstating the severity of this issue? Do you have any thoughts on what's causing this bug?

ndmitchell:
@snoyberg - how did you hit that issue? Is there some tool like Cabal than runs it? I was doing that for debugging, so not sure if anything "real" does it.

snoyberg:
I ran `stackage nightly` with GHC 7.10, which runs `cabal haddock --hoogle` (and a few other options). I'll go and see if there's a minimal example of it.

ndmitchell:
Does it fail if you omit the `--hoogle`? That's about the only thing you can take out of my example above.

snoyberg:
--hoogle is the trigger to cause it to fail:

```
$ cabal haddock --hoogle
Running Haddock for acme-missiles-0.3...
Preprocessing library acme-missiles-0.3...
Warning: The documentation for the following packages are not installed. No
links will be generated to these packages: stm-2.4.4
Haddock coverage:
 100% (  2 /  2) in 'Acme.Missiles'
  75% (  3 /  4) in 'Acme.Missiles.STM'
  Missing documentation for:
[...]

Issue #353, opened 2014-12-23
Panic with package keys and --hoogle


luite:
Haddock needs the package version and name to generate a hoogle index. Before package keys, this could be (de)constructed from the passed in package name. The lookup code has since been changed to get the version from the package db.

This works, as long as the package key can be found in one of the package databases, which is not always the case for the package currently being compiled. Since cabal can now generate documentation for executables and test suites, we can't rely on the inplace package database either.

To reproduce the panic:

```
module Test where
test = test
```

```
$ haddock --hoogle Test.hs
```

The same panic can be triggered with `cabal haddock` when `--hoogle` is enabled. Make sure the package hasn't been installed before, otherwise the packagekey can be found in the user package db.

I think we need a new command line option to pass in the package name and version, since it can't be determined from the package key anymore. `-package-name` is deprecated in GHC 7.10 too, and relying on a `--optghc` option for this isn't a great solution in any case. Cabal also needs to be updated, since it's still using the deprecated `-package-name` flag with haddock.

A quick fix: rely on Cabal passing in `--optghc=-package-name --optghc=packagename-version` and use the name/version pair as a fallback when looking up the package with `GhcUtils.modulePackageInfo` fails.

Proper fix? Add command line options `--package-name` and `--package-version` to haddock, update Cabal to pass these in.

PR #369, opened 2015-02-25
Make the error encountered when a package can't be found more user-friendly

bgamari:
Previously we used `getPackageDetails` to fetch information about the package providing a module. This started failing regularly when GHC moved to package keys, leading to errors of the form `haddock: internal error: expectJust getPackageDetails` (see https://github.com/haskell/cabal/issues/2297, https://github.com/haskell/haddock/issues/363, https://github.com/haskell/haddock/issues/361, https://github.com/haskell/haddock/issues/353).

Now we use `lookupPackage` instead, which provides the means for us to gracefully handle failure. The user should now see the slightly more helpful error,
```
haddock: Unable to find a package providing module "++moduleName pkgMod++ ", skipping Hoogle."

         Perhaps try specifying the desired package explicitly using the --package-name"
         and --package-version arguments."        
```
Still not great, but hopefully along with a fixed Cabal release this error should occur rather infrequently.

### DIFF
--- a/haddock-api/src/Haddock.hs
+++ b/haddock-api/src/Haddock.hs
@@ -39,6 +39,7 @@ import Haddock.Options
 import Haddock.Utils
 
 import Control.Monad hiding (forM_)
+import Control.Applicative
 import Data.Foldable (forM_)
 import Data.List (isPrefixOf)
 import Control.Exception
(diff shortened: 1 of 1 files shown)


bgamari:
Still waiting for my test builds to finish.

Fuuzetsu:
Totally forgot about this. Making some minor changes and pushing it in.

Fuuzetsu:
Sorry for the hold up. I made some minor changes to the commit (style only) and pushed in.

Issue #151, opened 2014-05-08
long lines in the synopsis


ghc-mirror:
**Original reporter**: *s.menge@\<hidden\>*

the synopsis widget looks bad for very long lines (which is bad style anyway).

e.g. http://hackage.haskell.org/packages/archive/gdiff/1.0/doc/html/Data-Generic-Diff.html#t:EditScriptL

PR #373, opened 2015-03-16
Prevent Synopsis from using up too much horizontal space

Rufflewind:
When long type signatures occur in the Synopsis, the element is stretched beyond the width of the window.  Scrollbars don't appear, so it's impossible to read anything when this happens.


Fuuzetsu:
Can you show a screenshot of before and after?

Rufflewind:
screenshot 2015-03-17 07 40 57

Fuuzetsu:
Thanks! It actually fixed an open issue (#151). I updated the changelog too.

PR #386, opened 2015-04-26
Change ModuleTree to fix #385

### LINKED ISSUE #385: Contents page has broken (local) links for modules.
This is with ghc 7.10.1 on OSX 10.10.3 compiled from source (although the ghcformacosx version seemed to behave the same).
Cabal and cabal-install 1.22.3.0
Haddock from ghc, hackage (2.16.0) and git behave similarly.

I have `documentation: True` and `hyperlink-source: True` in my `~/.cabal/config`

The links to modules on the contents page (`~/.cabal/share/doc/x86_64-osx-ghc-7.10.1/index.html`) are relative and therefore broken as the actual page referred to is in a subdir named after its package.

It seems that commit `﻿6e95f42907970c9cbb9f6279509a4bf5542ffae4` modified the `ModuleTree` so that the nodes refer to a `SourcePackageId` instead of a `PackageKey`.  Unfortunately when `Haddock.Backends.Xhtml.Names.ppModule` calls `moduleUrl` on this to get the anchor href it looks for the `Module`, constructed with the `SourcePackageId`, in `html_xrefs` where it was stored with a `PackageKey`. Thus no match is found and a local link is created.

Extending the `ModuleTree.Node` to carry an extra parameter so that it contains both the `PackageKey` and `SourcePackageId` and modifying `Haddock.Backends.Xhtml.mkNode` so that `htmlModule` uses the `PackageKey` and `htmlPkg` uses the `SourcePackageId` seems to work.  If that makes sense I could make a pull request but I'm not certain I fully understand all the ramifications.

amurrayc:
This is a pretty mechanical addition of an extra parameter to `ModuleTree.Node` so that it carries both the `PackageKey` (required so that the call to `ppModule` in `mkNode` can find the entry in `html_xrefs`) and the `SourcePackageId` (required so that `htmlPkg` in `mkNode` refers to a nice package name).

### DIFF
--- a/haddock-api/src/Haddock/Backends/Xhtml.hs
+++ b/haddock-api/src/Haddock/Backends/Xhtml.hs
@@ -288,7 +288,7 @@ mkNodeList qual ss p ts = case ts of
 
 
 mkNode :: Qualification -> [String] -> String -> ModuleTree -> Html
-mkNode qual ss p (Node s leaf pkg short ts) =
+mkNode qual ss p (Node s leaf pkg srcPkg short ts) =
   htmlModule <+> shortDescr +++ htmlPkg +++ subtree
   where
     modAttrs = case (ts, leaf) of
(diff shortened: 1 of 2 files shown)


Fuuzetsu:
@thoughtpolice do you want to test if this fixes what you wanted? I don't have the time to for few more days.

cstrahan:
Hey, looks like this was pushed in https://github.com/haskell/haddock/commit/bf31846b9f7280b5e75f09e91ca18c4ced37af08, as noted in #385. Perhaps this can be closed?

Fuuzetsu:
You're right.

PR #387, opened 2015-04-26
Do not insert anchor for section headings in contents box

watashi:
Haddock used to insert anchor for section headings both in contexts box and the actual headings. This makes a link `#section1` jump to the element in contexts box instead of the actual heading, because they have duplicate anchor name. See test for more details. The diff of `Bug387.html` is:
```diff
45c45,47
< 	  >Section1</li
---
> 	  >Section1<a name="a:section1"
> 	      ></a
> 	    ></li
47c49,51
< 	  >Section2</li
---
> 	  >Section2<a name="a:section2"
> 	      ></a
> 	    ></li
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@
 /doc/configure
 tags
 TAGS
+
+.cabal-sandbox
+cabal.sandbox.config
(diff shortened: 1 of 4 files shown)


Fuuzetsu:
Thanks!

PR #391, opened 2015-05-06
Ignore doc/haddock.{ps,pdf}

bgamari:
These have frequently clogged up my workflow as arcanist insists that there be no unstaged changes (including untracked files) before creating a differential.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,8 @@
 /latex-test/out/
 
 /doc/haddock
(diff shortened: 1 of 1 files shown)


Fuuzetsu:
In 319bf20d44589545e6850db3e676e30216dd3cc2

If you're working with GHC source as you seem to be, you'll need to arrange for it to make it into `ghc-head` branch.

Issue #382, opened 2015-04-08
Proposal: render math formulas


solatis:
Hello,

As I understand it, the topic of Haddock being able to render math formulas in a pretty way comes up sometimes. There are several ways to tackle this issue:

 * use graphviz (like doxygen)
 * write a pure haskell implementation of mathtex;
 * use a client side library like [MathJax](https://www.mathjax.org/)

I would like to start a discussion whether there is any interest in tackling this issue. Personally, I think the MathJax way might be the most pragmatic since it renders everything purely client side: you will not have any additional dependencies for rendering the HTML, and the only thing that is required is including some javascript library and generating proper javascript when math expressions are encountered in haddock markup.

I would be interested in the point of view of the maintainers of haddock on this issue.

hvr:
As MathJax is well-established by now, I think that'd be the fastest/easiest way to get us nicely rendered math-formulas... What markup-syntax on the Haddock side would you suggest for dropping into math-mode? Maybe simply `$f(x) = x$` like in TeX?

Fuuzetsu:
So yes, it does come up frequently. Straight away I think we can abandon the idea that we'll render the formulas ourselves due to extra dependencies and whatnot.

As usual with these proposals, there's a question ‘can every backend do something reasonable with it?’. This question is mostly why we don't allow inlining HTML.

But in this case yes, I think we can do reasonable things. We can render in both HTML (somehow) and LaTeX, and just leave as-is in Hoogle.

I think mathjax is the way to go here. Downside is that it requires JavaScript to be enabled but in that case I think it's OK for the users to just see the formula.

So in the end I'd have no problem with such a feature.

Regarding syntax, I'd like to use `$foo$` but I fear the clash between named blocks. What do you think?

hvr:
Then maybe `$$f(x) = x$$` instead?

idontgetoutmuch:
$$...$$ actually means something in latex. E.g.

```
$$
\begin{aligned}
\mathbb{C}[X_u, X_t \,\vert\, X_0 = x_0] &=
\mathbb{E}\Bigg[\Bigg( \sigma\int_0^u e^{-\beta(u - s)}\mathrm{d}W_s  \Bigg)
                \Bigg( \sigma\int_0^t e^{-\beta(t - s)}\mathrm{d}W_s  \Bigg)\Bigg] \\
&=
\sigma^2e^{-\beta(u + t)}
[...]

PR #397, opened 2015-05-16
Proposal for Mathjax

idontgetoutmuch:
Consider this a proposal for how we support math in haddock. I've only done inline math so far but display math shouldn't be much harder. See https://github.com/haskell/haddock/issues/382 for more discussion.


cartazio:
The one concern I have is how the rendering will look when the viewer doesn't have in working Internet or if the cdn url changes.  But maybe those are none issues

bkaes:
I don't think that using a valid operator name `$$` is appropriate for inline math:

1. `$$` is used for block math

idontgetoutmuch:
@cartazio it will look like this so not too disastrous.

mathjaxunavail

idontgetoutmuch:
@bkaes very interesting - I am clearly a TeX dinosaur ;-) No more $$...$$ for me.

hvr:
@idontgetoutmuch one could make the matjhax js url (optionally) configurable via a CLI option, that way ppl who need it could build local haddock docs which don't require internet access...

ocharles:
I do think that is pretty disastrous if you don't have a working internet connection, but I like the idea of being able to bundle in a local Mathjax source.

idontgetoutmuch:
Ok so my proposal is this: we use \(...\) for inline, \[...\] for display and allow the mathjax url to be configured via a CLI option. I will try and finish this off next weekend.

For info I am adding a linking to the reddit comments: http://www.reddit.com/r/haskell/comments/366fgo/proposal_for_mathjax_pr_haskellhaddock397/

bkaes:
@idontgetoutmuch Did you mean `\( … \)` and `\[…\]`?

idontgetoutmuch:
@bkaes yes I didn't realise that github wanted me to escape \\

Fuuzetsu:
Do we need two different syntax constructs for this? Can't it work like `@` does?

idontgetoutmuch:
I think using latex / mathjax syntax makes it easier for authors. I think most people would write the mathematics in another document e.g. a tex file and would want to cut and paste without having to re-format. Here's an example of what I mean. I can take this out of an existing document and put it straight into the proposed haddock (with each line being prefixed by "-- "). Ok I'd have to reformat the bold font but apart from that my job as an author is pretty straightforward. I think there is also a benefit to sticking to an existing convention rather than inventing our own.

```
Recall that a {\bf{transition kernel}} is a mapping \(\mu : X \times
{\cal{Y}} \rightarrow \overline{\mathbb{R}}_{+}\) where \((X,
{\cal{X}})\) and \((Y, {\cal{Y}})\)are two measurable spaces such that
\(\mu(s, \cdot)\) is a probability measure on \({\cal{Y}}\) for all
\(s \in X\) and such that \(\mu(\cdot, A)\) is a measurable function
on \(X\) for all \(A \in {\cal{Y}}\).

For example, we could have \(X = Y = \{a, b\}\) and \({\cal{X}} =
{\cal{Y}} = \{\emptyset, \{a\}, \{b\}, \{a,b\}\}\) and \(\mu(a,\{a\}) =
0.4, \mu(a,\{b\}) = 0.6, \mu(b,\{a\}) = 0.6, \mu(b,\{b\}) =
0.4\). Hopefully this should remind you of the transition matrix of a
Markov chain.

Recall further that a family of such transitions \(\{\mu_t\}_{t \in T}\)
where \(T\) is some index set satisfying

\[
\mu_{t+s}(x, A) = \int_{Y} \mu_s(x, {\mathrm{d}y})\mu_t(y, A)
\]
```

which is rendered via pdflatex as
avoidreformat

idontgetoutmuch:
You can now say

{code}
haddock --mathjax=file:///Users/whatever/MathJax/MathJax.js?config=TeX-AMS-MML_HTMLorMML
{code}

Fuuzetsu:
Left a few comments. You should aim to squash some of the spurious commits you have as they serve no purpose on their own and save git tree clutter.

idontgetoutmuch:
Thanks.

1. I  just email Neil.

Fuuzetsu:
Been looking at this so it may potentially go into 2.16.1. You don't seem to have updated the `InterfaceFile.hs` as a warning during compilation helpfully points out. That bit is kind of important or we end up losing the maths the moment multiple packages are involved. It's probably too late for 2.16.1 (due to be out with 7.10.2 which is due into RC very soon; my fault for being so tardy but that's how it is) but we may release another version later (which breaks interface file too, woo) independently or if 7.10.3 is happening then with that. We have a GSOC student doing some stuff for us so a bigger release will probably be in order anyway.

Branch also needs cleaning a bit (don't really need 5 fixup commits) and rebasing on master (1 new test added since that needs updating). I did these things to a basic degree at https://github.com/Fuuzetsu/haddock/tree/mathjax-merge if you want a reference.

idontgetoutmuch:
I have some time today so I will try my best.

idontgetoutmuch:
Ok I have squashed the commits but the required instances in `InterfaceFile.hs` seem to be generated by DrIFT. I don't know how to invoke DrIFT to do the necessary.

Fuuzetsu:
Just pointing out that this PR has not been forgotten nor is being rejected. I am delaying it due to interface file changes required are an ugly update. But it only adds to the end of the structure (rather than changing existing types) so maybe it won't be too bad. I may include it in the next release which I hope comes fairly soon, including changes from GSOC.

@idontgetoutmuch 

You can ignore the DrIFT comment as you did, I think it's just a very old mechanism that generated instances for you.  Should just remove the comment.

idontgetoutmuch:
Thanks for keeping me updated. I removed the reference to DrIFT already I think.

Fuuzetsu:
That's great. We're keeping this PR on hold due to interface change, I want to see if I can't bundle some other similar changes with it instead of breaking things twice. So it may be a bit still before I merge and release it.

idontgetoutmuch:
@Fuuzetsu any news on this? Will you make a release coincide with gcc 8? That would be in January I think?

bgamari:
@idontgetoutmuch I don't mind merging this for GHC 8.0. @idontgetoutmuch perhaps you could rebase on my [ghc-head](https://github.com/bgamari/haddock/tree/ghc-head) branch?

idontgetoutmuch:
Ok I'll try my best - my git skills are not great.

recursion-ninja:
Very excited, will begin using the feature as soon as it is released!

bgamari:
This has been merged in #465!

PR #398, opened 2015-05-17
attach to instance location the name that has the same location file.…

### LINKED ISSUE #383: Typeclass instance links incorrect for base-4.8.0.0
I'm not sure if this is Haddock's fault (it may be similar to #372), but I'll file an issue just in case. Currently, several typeclass instance hyperlinks that Haddock generated for [`base-4.8.0.0`](http://hackage.haskell.org/package/base-4.8.0.0) go to the wrong modules. As an example, the instance link for `Functor STM` in [`Data.Functor`](http://hackage.haskell.org/package/base-4.8.0.0/docs/Data-Functor.html#t:Functor) currently links to [`GHC.Base`](http://hackage.haskell.org/package/base-4.8.0.0/docs/src/GHC-Base.html#line-628) (where `Functor` is defined); however, the actual instance is located in [`GHC.Conc.Sync`](http://hackage.haskell.org/package/base-4.8.0.0/docs/src/GHC-Conc-Sync.html#line-628) on line 628 (the line number, at least, seems to be correct, but the module is wrong).

JPMoresmau:
… Fixes #383

We carry over the source location of the instance AND a name that is in the same file (either the class name or the instance name, depending). This allows us to generate source links in the case of having the instance definition colocated with the type, or with the class. this won't work for orphan instances, but for orphan instances we have no names to attach to anyway.

### DIFF
--- a/haddock-api/src/Haddock/Backends/LaTeX.hs
+++ b/haddock-api/src/Haddock/Backends/LaTeX.hs
@@ -544,14 +544,14 @@ ppDocInstances unicode (i : rest)
     (is, rest') = spanWith isUndocdInstance rest
 
 isUndocdInstance :: DocInstance a -> Maybe (InstHead a)
-isUndocdInstance (L _ i,Nothing) = Just i
+isUndocdInstance (i,Nothing,_) = Just i
 isUndocdInstance _ = Nothing
 
 -- | Print a possibly commented instance. The instance header is printed inside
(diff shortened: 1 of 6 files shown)


Fuuzetsu:
@JPMoresmau I can't merge this because it breaks stuff, see Travis for example.

The module looks like
 and those `ImplicitParams` should not be there.

JPMoresmau:
Damn, forgot to remove some debug info. Sorry about that.

PR #400, opened 2015-05-23
Build haddock executable with '-threaded' (fixes #399)

### LINKED ISSUE #399: Build haddock executable with `-threaded`
In an internal project, there is some Template Haskell that requires the multithreaded RTS to run.  Running haddock on this code fails:

```
Haddock coverage:

src/My/Module.hs:60:3:
    Exception when trying to run compile-time code:
      user error (RTS doesn't support multiple OS threads (use ghc -threaded when linking))
    Code: template-haskell:Language.Haskell.TH.Quote.quoteExp
            z "foo"
```

Building the `haddock` executable with `--ghc-options=-threaded` seems to fix this.

Would there be any disadvantage to adding `-threaded` to haddock.cabal's ghc-options?  If a simple example would be helpful, let me know and I'll put one together.

(this is with GHC 7.8.4 on Linux x86_64, in case it's relevant)

manny-fp:
This includes a test which verifies that the threaded RTS is being used.

### DIFF
--- a/haddock.cabal
+++ b/haddock.cabal
@@ -42,7 +42,7 @@ executable haddock
   default-language:     Haskell2010
   main-is:              Main.hs
   hs-source-dirs:       driver
-  ghc-options:          -funbox-strict-fields -Wall -fwarn-tabs -O2
+  ghc-options:          -funbox-strict-fields -Wall -fwarn-tabs -O2 -threaded
 
   build-depends:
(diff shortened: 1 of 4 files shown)


manny-fp:
Is there a chance of getting this merged before the next GHC release?

Fuuzetsu:
yes, sorry, was busy

PR #401, opened 2015-05-27
List nesting improvements (fixes #278)

### LINKED ISSUE #278: Nested lists/paragraphs are not relative
**Original reporter**: *fuuzetsu@\<hidden\>*

Consider

```haskell 
{-|
Foo

      - bar

      - baz
-}
```

The ‘baz’ list is nested under ‘bar’. A saner behaviour would be to start counting indentation from the first list.

mrhania:
It hopefully fixes #278, although there may be some corner cases I haven't thought about. It also slightly changed (fixed?) behaviour in case `k` of `Nesting.hs` test module. Before, *level 3 header* was parsed as a child of the *[c]* item. But they are separated by an empty line so the header should be just the element of paragraph, not a child of a list. I might be wrong of course, but then it would be erratic with the way the example item is treated couple lines above.

### DIFF
--- a/haddock-library/src/Documentation/Haddock/Parser.hs
+++ b/haddock-library/src/Documentation/Haddock/Parser.hs
@@ -231,18 +231,20 @@ markdownImage = fromHyperlink <$> ("!" *> linkParser)
 
 -- | Paragraph parser, called by 'parseParas'.
 paragraph :: Parser (DocH mod Identifier)
-paragraph = examples <|> skipSpace *> (
-      since
-  <|> unorderedList
-  <|> orderedList
-  <|> birdtracks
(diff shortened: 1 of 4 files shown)


Fuuzetsu:
Very nice, I have no issues with what's here. Can you add a spec test for the parser too, under `test` in `haddock-library`? Those tests can be ran much more quickly and independently of the HTML backend tests which makes them very nice when hacking on the parser.

mrhania:
That `test` folder in `haddock-library` was well hidden, I must have missed it. But as suggested, I have added simple spec test for this improvement.

Fuuzetsu:
Thanks.

Issue #207, opened 2014-05-08
haddock: internal error: spliceURL UnhelpfulSpan


ghc-mirror:
**Original reporter**: *selinger@\<hidden\>*

With Haddock 2.8, I got this error message occasionally; with Haddock 2.9.4, I get it *all the time*. Minimal example attached.

The problem seems to arise when one has some function without a type signature, and the --source-entity option is used. The old behavior (2.8) was to ignore such a function. The new behavior is to generate the type signature, but then fail with the above error message.

To reproduce: 

haddock -h HaddockBug.hs --source-entity=HaddockBug.html#line-%L

Obviously, the above error message was not intended for end users; "Missing type signature" or "Can't find this line in your file" would be a better error message. Still better would be not to produce an error at all.

PR #402, opened 2015-06-06
Fix haddock: internal error: spliceURL UnhelpfulSpan (#207)

niteria:
Inferred type signatures don't have SrcSpans, so let's use the one from
the declaration.

I've tested this manually on the test-case from #207, but I got stuck at
trying to run the test-suite.


Fuuzetsu:
Took the liberty to squash these two commits and apply on top of master

PR #407, opened 2015-06-15
Link to the definitions to themselves

Rufflewind:
Currently, the definitions already have an anchor tag that allows URLs with fragment identifiers to locate them, but it is rather inconvenient to obtain such a URL (so-called "permalink") as it would require finding the a link to the corresponding item in the Synopsis or elsewhere.  This commit adds hyperlinks to the definitions themselves, allowing users to obtain links to them easily.

To preserve the original aesthetics of the definitions, we alter the color of the link so as to be identical to what it was, except it now has a hover effect indicating that it is clickable.

Additionally, the anchor now uses the `id` attribute instead of the (obsolete) `name` attribute.

### DIFF
--- a/haddock-api/resources/html/Ocean.std-theme/ocean.css
+++ b/haddock-api/resources/html/Ocean.std-theme/ocean.css
@@ -41,6 +41,9 @@ a[href]:link { color: rgb(196,69,29); }
 a[href]:visited { color: rgb(171,105,84); }
 a[href]:hover { text-decoration:underline; }
 
+a[href].def:link, a[href].def:visited { color: black; }
+a[href].def:hover { color: rgb(78, 98, 114); }
+
 /* @end */
 
 /* @group Fonts & Sizes */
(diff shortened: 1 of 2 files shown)


Rufflewind:
Hmm ... I wonder if it might be better to put the link in a separate text element, like a hash (`#`) or a pilcrow (`¶`).  Currently, having the named item be linked makes it annoying to actually copy-paste the item's name.  What do you think?

Fuuzetsu:
I'd rather not add extra stuff on the line, for now it still looks like valid Haskell and it'd be nice to keep it that way. If anything it'd have to go to the right of the signature. But in what way will a separate element be better?

Rufflewind:
Here's a mockup:

untitled

The `🔗` could also be "hover-only" so it never appears until the user hovers near it.

> But in what way will a separate element be better?

It makes copying the name of the symbol more difficult.

Normally, it'd be double-click then Ctrl+C.  Now you'd have to very carefully drag out the selection, and then Ctrl+C.  Unfortunately, current browsers don't really offer an easy way to select link text.

Fuuzetsu:
If anything, I'd rather have a `#` on the right side, near `Source` button &c.

PR #410, opened 2015-06-30
Source hyperlinker

mrhania:
[First part](https://ghc.haskell.org/trac/summer-of-code/ticket/1663) of my Google Summer of Code project. I am sure that there are many, many issues I am not aware of, so treat is as a beta for now. The real testing will begin once Cabal incorporates this to its pipeline.


haddock-api/haddock-api.cabal
```diff
@@ -89,6 +96,37 @@ library
...
+      base >= 4.3 && < 4.9
+    , bytestring
+    , filepath
+    , directory
+    , containers
+    , deepseq
+    , array
+    , xhtml >= 3000.2 && < 3000.3
+    , Cabal >= 1.10
+    , ghc >= 7.10 && < 7.10.2
```
Fuuzetsu: I think relaxing this upper bound would be good, perhaps to something like `7.10.*`. Unless we have a reason to believe that it won't work with 7.10.2, there's no need to constrain it this tightly. In all likelyhood, this will come out after 7.10.2 is released anyway.

haddock-api/haddock-api.cabal
```diff
@@ -89,6 +96,37 @@ library
...
+
+  build-depends:
+      base >= 4.3 && < 4.9
+    , bytestring
+    , filepath
+    , directory
+    , containers
+    , deepseq
+    , array
+    , xhtml >= 3000.2 && < 3000.3
```
Fuuzetsu: The bound seems pretty strict. Is there a reason?

haddock-api/src/Haddock/Backends/Hyperlinker/Ast.hs
```diff
@@ -0,0 +1,192 @@
...
+        ]
+
+-- | A map containing association between source locations and "details" of
+-- this location.
+--
+-- For the time being, it is just a list of pairs. However, looking up things
+-- in such structure has linear complexity. We cannot use any hashmap-like
+-- stuff because source locations are not ordered. In the future, this should
+-- be replaced with interval tree data structure.
+type DetailsMap = [(GHC.SrcSpan, TokenDetails)]
```
Fuuzetsu: On `ghc-7.10` branch of GHC (7.10.2 candidate) I see

```
data SrcSpan =
    RealSrcSpan !RealSrcSpan
  | UnhelpfulSpan !FastString   -- Just a general indication
                                -- also used to indicate an empty span

  deriving (Eq, Ord, Typeable, Show) -- Show is used by Lexer.x, because we
                                     -- derive Show for Token
```

If I look at the docs for `7.10.1` I see `Ord` there too: https://downloads.haskell.org/~ghc/7.10.1/docs/html/libraries/ghc-7.10.1/SrcLoc.html

Surely we can use a `Map` from `containers` then? `containers` is a core lib so we can use it without a problem.

haddock-api/src/Haddock/Backends/Hyperlinker/Parser.hs
```diff
@@ -0,0 +1,235 @@
...
+    | TkPragma
+    | TkUnknown
+    deriving (Show, Eq)
+
+-- | Turn source code string into a stream of more descriptive tokens.
+--
+-- Result should retain original file layout (including comments, whitespace,
+-- etc.), i.e. the following "law" should hold:
+--
+-- @concat . map 'tkValue' . 'parse' = id@
```
Fuuzetsu: Haddock actually supports `prop>` markup which when combined with `doctest` (and `QuickCheck`) can verify this this law in test phase. Might be worthwhile though not necessarily that high priority, I'm expecting there's a test for it already (haven't read down that far yet). It's similar to `>>>` in that it's code that can be ran as part of a testsuite.

haddock-api/src/Haddock/Backends/Hyperlinker/Parser.hs
```diff
@@ -0,0 +1,235 @@
...
+    , "~#"
+    , "=>"
+    , "-"
+    , "!"
+    ]
+
+special :: [Char]
+special = "()[]{},;`"
+
+-- TODO: Add support for any Unicode symbol or punctuation.
```
Fuuzetsu: This one is a bit tricky as it's a moving target. Every once in a while GHC updates its list of unicode symbols (they grab tables off the web) and logic for how all this works is a bit special.

I'd say supporting unicode stuff should addressed sooner than later but we might have to call out to GHC for that. If it can't do that for us right now, it'll have to be hacked so it can. Not vital for right now though.

haddock-api/src/Haddock/Backends/Hyperlinker/Parser.hs
```diff
@@ -0,0 +1,235 @@
...
+    | c == '"' = TkString
+    | c == '\'' = TkChar
+classify str
+    | str `elem` keywords = TkKeyword
+    | str `elem` glyphs = TkGlyph
+    | all (`elem` symbols) str = TkOperator
+    | isIdentifier str = TkIdentifier
+    | otherwise = TkUnknown
+
+keywords :: [String]
```
Fuuzetsu: We should probably be asking GHC about these eventually

haddock-api/src/Haddock/Backends/Xhtml/Utils.hs
```diff
@@ -38,18 +37,23 @@ import Text.XHtml hiding ( name, title, p, quote )
...
 import Name     ( getOccString, nameOccName, isValOcc )
 
 
 spliceURL :: Maybe FilePath -> Maybe Module -> Maybe GHC.Name ->
              Maybe SrcSpan -> String -> String
-spliceURL maybe_file maybe_mod maybe_name maybe_loc = run
+spliceURL mfile mmod = spliceURL' mfile (moduleName <$> mmod)
+
+
+spliceURL' :: Maybe FilePath -> Maybe ModuleName -> Maybe GHC.Name ->
```
Fuuzetsu: Maybe a quick comment (`-- |`) on what way this splices the URL? If we're looking through docs, we'll no doubt need to check the source to see what this actually does or is supposed to do. Yeah, there was no comment on original `spliceURL`, but that's even more reason to add one.

haddock-api/test/Haddock/Backends/Hyperlinker/ParserSpec.hs
```diff
@@ -0,0 +1,97 @@
...
+import Haddock.Backends.Hyperlinker.Parser
+
+
+main :: IO ()
+main = hspec spec
+
+
+spec :: Spec
+spec = do
+    describe "parse" parseSpec
```
Fuuzetsu: no point for the `do` just for this line

hypsrc-test/Utils.hs
```diff
@@ -0,0 +1,47 @@
...
+    replaceBetween' = replaceBetween pref end val
+
+stripLocalAnchors :: String -> String
+stripLocalAnchors = replaceBetween "<a name=\"local-" '\"' "0"
+
+stripLocalLinks :: String -> String
+stripLocalLinks = replaceBetween "<a href=\"#local-" '\"' "0"
+
+stripLocalReferences :: String -> String
+stripLocalReferences = stripLocalLinks . stripLocalAnchors
```
Fuuzetsu: I really dislike how this works. It relies far too much on the layout of the output we produce. It feels silly that we're using a HTML lib to output stuff and then do string processing to filter stuff out of it. I wonder if we could leverage the library to parse the content and process it that way? I have not checked, have you?

This of course goes for the existing tests we had and is just a general observation but if we're adding new ones, maybe we could do it a bit better.

hypsrc-test/src/Records.hs
```diff
@@ -0,0 +1,25 @@
+{-# LANGUAGE NamedFieldPuns #-}
+
+module Records where
```
Fuuzetsu: Should definitely also test `RecordWildCards` as it's a popular one.

Fuuzetsu:
Left few minor comments on the diff itself.

I am using `yi-rope` package (git master) for testing. I ran
`[nix-shell:~/programming/yi-rope]$ cabal configure && cabal haddock --with-haddock=/tmp/haddock-build/bin/haddock --haddock-options=--pretty-html --haddock-options=--hyperlinked-source`

Few observations/questions:

* Given a multi-part identifier such as `Yi.Rope.empty`, when I hover over it, the whole thing gets highlighted as one would expect. But when I click on it/drag, only the part I click on gets ‘selected’: if I click/drag on `Yi` part, only that gets the selection frame. It's distracting visually and can make the user think that which part they click on matters: after all, they could either have jumped to `empty` or the part they clicked on. Would it be possible for the whole identifier to count as one entity?

* if I zoom in, I notice all the hoverable identifiers are underlined. I wonder if it's by design. I'm fine with it staying this way by the way.

* I think I haven't seen a test for clickable polymorphic vars inside type signatures, i.e. for stuff like `foo :: a -> a`. Highlighting seems to work but I haven't tried in cases like `foo :: a -> (forall. a -> a) -> a`. Should have a test for this.

* I wonder if when we click on identifier, instead of jumping to the binding site and putting it on the top of the page, it could instead put it slightly lower, say in the center. We probably want to see the code a bit before it too. At the same time I'm reluctant towards this: if user is clicking on an identifier, presumably the user wants to be able to find out quickly where it as and I guess we are trained to look at the top after clicking on anchors. What do you think? It's probably best to leave it as-is (going to top) but maybe you have a second opinion.

* highlighting for `@-patterns` just doesn't work: for `foo z@foo = z`, hovering over RHS `z` doesn't highlight the LHS `z`. Can't hover over LHS `z` at all.

* I wonder if there's a plan to do the highlighting trick for type signatures: `foo :: Bar a => a -> a -> b` where you could hover over `a` and `b`. It'd only highlight within that type signature of course. Maybe a bit out of scope but if it's easy to do then it might be cool.

* I wonder if it'd be possible to make identifiers inside comments clickable. We already parse and rename them for the HTML backend, maybe we could benefit from that. I imagine this might be a bit harder so I don't know. It'd certainly be cool.

Of course these are just some quick questions/suggestions, by no means are they blockers. Probably want to look into `@-patterns` and add a test for the type signature clickables now though.

I'm going to try with `lens` now and leave more comments.

Fuuzetsu:
Something seems broken. For `lens` I only get this stuff in `src`:



Seems like a fairly big problem.

Nevertheless, in block

```
fromSet :: (k -> v) -> Set.Set k -> Map.Map k v
#if MIN_VERSION_containers(0,5,0)
fromSet = Map.fromSet
#else
fromSet f x = Map.fromDistinctAscList [ (k,f k) | k <- Set.toAscList x ]
#endif
```

I notice the `f` and `x` identifiers aren't highlighted. I suspect that it's because of the CPP around it and the code only enters first branch but I note that I haven't seen a test with list comprehensions, hint hint :)

Fuuzetsu:
Oh, edit to the above, didn't spot it at first but in the terminal there is:

`haddock: internal error: No match in record selector tcdLName`

tsk tsk tsk

Fuuzetsu:
When `ScopedTypeVariables` is on and we have

```
foo :: forall a. a -> a
foo x = x :: a
```

Would it be possible to have the `a` in definition link to the `a` in signature? That's how it's scoped. Not really a high priority but would be cool to be able to jump like this. A bit like Agda's locally defined types (in `where` blocks and whatnot). I notice that it is scoped properly in class definitions already so maybe it Just Works™. We need tests for this stuff… Current test setup might not be able to test this which is something I have alluded to a while ago before you started your endavour. Might have to consider upgrading it.

--

I am looking at output for `aeson` (git master) now.

Given something like

```
class Foo a where
  someMethod :: a

bar = someMethod
```

clicking on the `someMethod` inside `bar` won't jump to the type signature in the class definition. Presumably this is the same issue as I alluded to earlier where we couldn't jump to type sigs. It works fine for jumping to default definition.

Of course there are few things that can be done when user clicks on a class method:

1. Jump to the type signature in the class
2. Jump to the default definition if present (else do 1)
3. If we know which instance is used, jump to the actual definition of the field at that instance

3 is the hardest and potentially confusing because we sometimes would jump to the class and sometimes to instance. I think we should go with 1 or 2. Currently we do 2 but we don't have 1 as fallback. Seems pretty important to be able to jump to class methods when possible.

--

That's about it from me for now. The highest priority would be fixing the crash when ran on `lens`. We can not make this a default if it crashes, especially on such a popular library.

In general what to take from this is that we need more tests and they have to test that things are anchoring properly. Currently we're only testing the lexing.

You should also see about updating the doc file. I may turn that into something like Markdown tomorrow to encourage this even more as we've been planning to do so for a while now anyway, that should be far easier for you to edit. In this doc there should be how to use the feature, what to expect, what to not expect, what needs JS, what doesn't, what's broken, what's not working by design &c &c. Basically some overview that could answer most of my questions from here instead of having to play ‘try and see’.

I also want to encourage you to add some more comments. You have a fair amount of doc comments on the stuff you have added but there are functions you have added or changed which don't' have one. A brief name might make all the sense to you when you know how to use it but if someone is tracking something down, they don't often want to jump into source of every little function just to find out its purpose.

In general this is fine work. I have not tracked *how* it works yet, only briefly skimmed the diff, left few comments there and gave it a spin. I may still have some input on how something should be done logic-wise though that will have to wait until next time I'm awake. If you can work out the quirks (*especially* the crash) and do a stronger job on the testing side (we need to test anchors for sure), I would have 0 problems making this the default source output. We should still keep the HsColour possibility around for a release so that people have a chance to give feedback. If there's compelling reason in that time to keep using HsColour, we'll just drop anything to do with that and make your work the only officially supported way. I imagine some people might quote performance, something we might need to investigate later.

haddock-api/haddock-api.cabal
mrhania: I just mindlessly copied it from the normal build dependencies. However, I don't think most of these are actually needed for spec so I can safely get rid of them.

haddock-api/haddock-api.cabal
mrhania: Same reason as above: because it was like this in library build dependencies.

haddock-api/src/Haddock/Backends/Hyperlinker/Ast.hs
mrhania: When I wrote that they are not ordered I meant that source spans do not form any sane total order (spans are intervals and intervals are just sets, and sets are only partially ordered). They might be ordered in a Haskell sense (by using ordinary lexical ordering on span coordinates for example), but it is not very helpful as I am looking things up by checking for interval intersection, hence the `Map` has no use here. Suitable data structure would be either [segment tree](https://en.wikipedia.org/wiki/Segment_tree) or [interval tree](https://en.wikipedia.org/wiki/Interval_tree) but these are not available in core as far as I know.

haddock-api/src/Haddock/Backends/Hyperlinker/Parser.hs
mrhania: Yep, this keyword list is essentially just copied from one of the GHC source files so it would be nice to somehow get these automagically.

hypsrc-test/Utils.hs
mrhania: Not a big fan of it either, I simply tried to mirror behaviour of HTML spec test runner. However, it looks like XHTML library Haddock uses doesn't support HTML parsing. Which is not that surprising considering generating HTML is much, much easier than reading it. However, since it generates (hopefully valid) XHTML, maybe there is XML library in core that could be used for that?

mrhania:
> Given a multi-part identifier such as Yi.Rope.empty, when I hover over it, the whole thing gets highlighted as one would expect. But when I click on it/drag, only the part I click on gets ‘selected’: if I click/drag on Yi part, only that gets the selection frame. It's distracting visually and can make the user think that which part they click on matters: after all, they could either have jumped to empty or the part they clicked on. Would it be possible for the whole identifier to count as one entity?

Definitely doable and rather necessary for people with disabled JavaScript. It might require some refactoring though.

> if I zoom in, I notice all the hoverable identifiers are underlined. I wonder if it's by design. I'm fine with it staying this way by the way.

By design. Can be easily tweaked in CSS file.

> I wonder if when we click on identifier, instead of jumping to the binding site and putting it on the top of the page, it could instead put it slightly lower, say in the center. We probably want to see the code a bit before it too. At the same time I'm reluctant towards this: if user is clicking on an identifier, presumably the user wants to be able to find out quickly where it as and I guess we are trained to look at the top after clicking on anchors. What do you think? It's probably best to leave it as-is (going to top) but maybe you have a second opinion.

This is just the default browser behaviour when jumping to anchors. Possibly it could be altered with JavaScript but I am not a fan of overriding default behaviours and I personally hate when websites do this (back button, link right click and scrolling just to name the few most annoying).

> highlighting for @-patterns just doesn't work: for foo z@foo = z, hovering over RHS z doesn't highlight the LHS z. Can't hover over LHS z at all.

An oversight, should be easy to fix.

> I wonder if it'd be possible to make identifiers inside comments clickable. We already parse and rename them for the HTML backend, maybe we could benefit from that. I imagine this might be a bit harder so I don't know. It'd certainly be cool.

Depends how *well coded* should it be ;) *Well coded* solution would reuse already existing Haddock parsing utilities and that could be a bit more complicated. Hacking something that just scans for strings wrapped in apostrophes and tries to match them with identifiers within the scope should be easier (although not 15-minutes-of-work easy).

> I notice the f and x identifiers aren't highlighted. I suspect that it's because of the CPP around it and the code only enters first branch but I note that I haven't seen a test with list comprehensions, hint hint :)

I am afraid it can't be helped because, as you said, GHC takes only one branch when processing file. Thus, AST is not available for opted out part of code and it can't be hyperlinked.

> Oh, edit to the above, didn't spot it at first but in the terminal there is:
>
> `haddock: internal error: No match in record selector tcdLName`

Of course, this should not happen. I will check what is up with `lens` and why I use record accessor where I am not supposed to do so.

hypsrc-test/Utils.hs
Fuuzetsu: For test dependencies we can use whatever we want. Being able to run this stuff inside GHC tree is nice but if it gimps us then screw it, it's more important that I (as Haddock dev) can write, test and amend stuff easily than that people not hacking on the project can run them. We already rely on stuff like hspec for the specs and that definitely is not in core.

Fuuzetsu:
>This is just the default browser behaviour when jumping to anchors.

Let's leave it as-is then.

>Depends how well coded should it be ;)

Probably to do this we'd have to include some location info in the parser. Seems tedious. Other hacks such as parsing the comments introduced issues like preserving whitespace properly and whatnot. I think we should just file this as a feature request later.

>I am afraid it can't be helped because, as you said, GHC takes only one branch when processing file. Thus, AST is not available for opted out part of code and it can't be hyperlinked.

I expected as much. Oh well, does not matter too much as long as the highlighting works which it seems to.

>I will check what is up with lens and why I use record accessor where I am not supposed to do so.

`lens` is pretty good library to check Haddock against, it uses many of Haddock's features and has a mix of language extensions, TH, CPP, all the nitty gritty basically.

mrhania:
Okay, some fixes:

* Polymorphic types are (almost) properly highlighted, clicking upon polymorphic variable jumps to appropriate `forall` binding (if available) and it works also with `ScopedTypeVariables`. There is an issue though. Signature `:: a -> (forall a. a -> a) -> a` is highlighted incorrectly while `:: forall a. a -> (forall a. a -> a) -> a` is highlighted correctly. The problem is with GHC API where the AST node span for leftmost `a` includes everything until the end of the line (the problem does not exist for version with explicit outer `forall`). I suspect that might be a bug in GHC itself.
* `@`-patterns are now properly highlighted.
* Qualified names are now rendered as one entity.

... and other minor code improvements. I have yet to investigate the `lens` issues.

Fuuzetsu:
Presumably 0d0550cdcf3fa7ceff88e2572f7ffb341b9f760d is for lens crash, right?

Could we perhaps make the `typ` helper used there less prone to crashing or more obvious that it may crash? For example we could try to catch the exception when it happens from GHC and do something (even crashing with better information would be an improvement). Probably better solution is to make the patterns explicit instead of using `_` on LHS and blindly calling `tcdLName`. At least we'll then know when GHC moves things around and we need to update.

--

It is very good to see you adding tests. GHC is a moving target and without tests it's pretty much impossible to tell how/when something breaks.

--

What's the situation with unicode? I know it was a TODO and I haven't seen changes. Even if the linking doesn't work optimally, we need to operate as usual. What does HsColour behave like? Notably, I don't want a regression in functionality. If HsColour is able to highlight unicode stuff properly but we aren't then that'd be a regression IMHO.

--

I wonder if you thought about how we could go through migration from HsColour. I would not like to exclude the possibility to use it. We can't update old cabal versions so we should probably make sure that if they call Haddock how they used to, stuff still works (even if our own highlighter is used). I think I will create an issue over at cabal and see how they want to go about it and what needs they have: maybe some stuff on Hackage needs to be flipped too.

If we ship with some grave bug that we don't catch but it becomes unusable for people then the option to use HsColour as usual needs to be there, at least for a release or two. After it has been battle-tested a bit, we could potentially remove any extra stuff we don't need for it (is there any?).

--

It would be good for you to prepare some kind of release notes for this feature. What it does and how to use. If you prefer, you can write a sort of a blog post and publish it somewhere once we merge and I can link to that when announcing new release (though of course you're free to publish early with subject-to-change disclaimer). Eventually we'll also need some words in the documentation, notably updating the flag descriptions and adding any new ones.

For early adopters, we should tell them what flags to use before cabal makes it default (if it has to do any work at all). Luckily there is nowadays a `--hadock-options` flag we can piggy-back on if need be.

--

>The problem is with GHC API where the AST node span for leftmost a includes everything until the end of the line (the problem does not exist for version with explicit outer forall). I suspect that might be a bug in GHC itself.

If it's a bug in GHC then we need to check if it's know or not and report it if necessary. @alanz is this stuff in your domain of reign?

--

Are there any bugs you know of or major features that are missing still? If there's something seriously broken then we need to fix it ahead of time and if not possible, ticket it. If there are non-critical issues missing then we need to open tickets. Assuming this PR works as advertised, I can't think off the top of my head of anything major except cross-package links which are out of scope for this right now or anything I haven't already mentioned/asked about.

I don't want to have Edward K on my case because Haddock is crashing or otherwise misbehaving on one of his libraries again.

--

Lastly, do *you* think what's here is OK for merging and use? Basically, do you have confidence in that it will work? I can only test so much myself before running out of time: if you can add tests for more cases then do! Stuff like the `lens` crash was only discovered because I happened to run on it but `lens` is only one package out of many thousand. If I see a test for something, I don't have to write or find a package that uses that feature saving me a lot of time and saving everyone time in the future.

I would like to merge within few days, maybe Saturday/Sunday if you give a heads up.

mrhania:
> Presumably 0d0550c is for lens crash, right?

Yep.

> Could we perhaps make the typ helper used there less prone to crashing or more obvious that it may crash? For example we could try to catch the exception when it happens from GHC and do something (even crashing with better information would be an improvement). Probably better solution is to make the patterns explicit instead of using _ on LHS and blindly calling tcdLName. At least we'll then know when GHC moves things around and we need to update.

Done the explicit pattern version. The problem is that constructors in the AST are sometimes quite intimidating, like with the `ClassDecl` case with 10 arguments. Fortunately (thanks to your suggestion for a hyperlinker test) I learned about `RecordWildCards` which made it a little bit nicer.

> What's the situation with unicode? I know it was a TODO and I haven't seen changes. Even if the linking doesn't work optimally, we need to operate as usual. What does HsColour behave like? Notably, I don't want a regression in functionality. If HsColour is able to highlight unicode stuff properly but we aren't then that'd be a regression IMHO.

HsColour doesn't recognize unicode either. And lack of it isn't that terrible, just some operators like `≫` being rendered as identifiers instead of whatever was set in CSS file. For token classification is use standard functions like `isUpper`, `isAlpha`, `isSpace` etc. which (allegedly) support Unicode characters.

> I wonder if you thought about how we could go through migration from HsColour. I would not like to exclude the possibility to use it. We can't update old cabal versions so we should probably make sure that if they call Haddock how they used to, stuff still works (even if our own highlighter is used). I think I will create an issue over at cabal and see how they want to go about it and what needs they have: maybe some stuff on Hackage needs to be flipped too.
>
> If we ship with some grave bug that we don't catch but it becomes unusable for people then the option to use HsColour as usual needs to be there, at least for a release or two. After it has been battle-tested a bit, we could potentially remove any extra stuff we don't need for it (is there any?).

As long as you don't use `--hyperlinked-source` flag, everything works as before. There is not much *extra stuff* except for `source-*` flags for specifying anchor format (which are ignored when `--hyperlinked-source` is on). They are not very HsColour-specific though. I think having them does no harm and can come handy in case we decide to make hyperlinker an external tool or something.

> It would be good for you to prepare some kind of release notes for this feature. What it does and how to use. If you prefer, you can write a sort of a blog post and publish it somewhere once we merge and I can link to that when announcing new release (though of course you're free to publish early with subject-to-change disclaimer). Eventually we'll also need some words in the documentation, notably updating the flag descriptions and adding any new ones.
>
> For early adopters, we should tell them what flags to use before cabal makes it default (if it has to do any work at all). Luckily there is nowadays a --hadock-options flag we can piggy-back on if need be.

I updated the documentation file (I hope I did it right as I have never done anything in DocBook before). Essentially, there are only two new flags and that's it. I have also added some information about (previously undocumented) source path option in `--read-interface` flag which I use now to enable cross-package hyperlinking - I am not sure what was the use of that before, though.

> Are there any bugs you know of or major features that are missing still? If there's something seriously broken then we need to fix it ahead of time and if not possible, ticket it. If there are non-critical issues missing then we need to open tickets. Assuming this PR works as advertised, I can't think off the top of my head of anything major except cross-package links which are out of scope for this right now or anything I haven't already mentioned/asked about.
>
> I don't want to have Edward K on my case because Haddock is crashing or otherwise misbehaving on one of his libraries again.

> Lastly, do you think what's here is OK for merging and use? Basically, do you have confidence in that it will work? I can only test so much myself before running out of time: if you can add tests for more cases then do! Stuff like the lens crash was only discovered because I happened to run on it but lens is only one package out of many thousand. If I see a test for something, I don't have to write or find a package that uses that feature saving me a lot of time and saving everyone time in the future.

I checked it on `vector`, `lens` and `Cabal` and it *seems* to work fine. But we can treat it as a beta feature, encourage people to try it out and just stick to HsColour for the time being if something goes wrong.

The only issues I know about:

* no module hyperlinks in imports
* incorrect type variable highlighting in certain cases (with `RankNTypes`)

I will create appropriate issues for them.

That being said, this is Haskell and tons of weird extensions I have never heard about. While it shouldn't crash no matter what (but obviously I can't guarantee that), it may produce wrong hyperlinks or syntax highlighting from time to time. But everything that HsColour does right now, my hyperlinker should be capable of as well.

Fuuzetsu:
>I have also added some information about (previously undocumented) source path option in --read-interface flag which I use now to enable cross-package hyperlinking

Can you elaborate? We don't have cross-package hyperlinking do we?

>But we can treat it as a beta feature, encourage people to try it out and just stick to HsColour for the time being if something goes wrong.

I find that not many people tend to do this. It's easier for us to make it the default and then advise to report issues and use HsColour if there's a problem.

>While it shouldn't crash no matter what (but obviously I can't guarantee that), it may produce wrong hyperlinks or syntax highlighting from time to time. 

Good enough. Logic bugs we can fix later, crashes are immediate problems.

>But everything that HsColour does right now, my hyperlinker should be capable of as well.

Good!

Fuuzetsu:
Oh yeah, and

>The problem is that constructors in the AST are sometimes quite intimidating, like with the ClassDecl case with 10 arguments. Fortunately (thanks to your suggestion for a hyperlinker test) I learned about RecordWildCards which made it a little bit nicer.

Alternatively you could have done something like

```
typ (ClassDecl { decl = decl }) = …
```

or 

```
typ c@(ClassDecl {}) = … decl c …
```

mrhania:
> Can you elaborate? We don't have cross-package hyperlinking do we?

We do if there is interface file available. For example, assume our package depends on `filepath`. So first we generate docs for `filepath`:

    cabal haddock --html --haddock-options="--hyperlinked-source --dump-interface=filepath.iface"

And then you can do something like:

    FILEPATH_ROOT="~/Desktop/filepath"
    FILEPATH_IFACE="${FILEPATH_ROOT}/filepath.iface"
    FILEPATH_HTML="${FILEPATH_ROOT}/dist/doc/html/filepath"
    FILEPATH_SRC="${FILEPATH_HTML}/src"
    cabal haddock --html --haddock-options="--hyperlinked-source --read-interface=${FILEPATH_HTML},${FILEPATH_SRC},${FILEPATH_IFACE}"

... except I just did it and something seems to be broken. `</>`s are hyperlinked but `FilePath`s are not for some reason. It worked before, I will look into this tomorrow.

> Alternatively you could have done something like

Thanks, I wasn't aware you can do partial record unpacking.

mrhania:
> ... except I just did it and something seems to be broken. </>s are hyperlinked but FilePaths are not for some reason. It worked before, I will look into this tomorrow.

Mystery solved: `FilePath`s are not hyperlinked because they originate from `base` (`Prelude`, `System.IO`), not from `filepath`. However, when Haddock generates documentation it is somehow able to associate `FilePath` with `System.FilePath.Windows`. The magic happens in `Haddock.Interface.Rename` module but I have not yet looked at it in detail.

Generally, these cross-package hyperlinking issues should happen only for things that are somewhat duplicated which is quite uncommon for things outside `base` I think.

There are also some good news: when searching for the source of a problem, I did a bit of refactoring and that gave me an idea how to solve problem with imports hyperlinking.

Fuuzetsu:
Awesome.

mrhania:
I have investigated how Haddock is able to hyperlink `FilePath` to `System.FilePath.Windows` of `filepath` instead of `System.IO` of `base` and I think I am not gonna do the same thing in hyperlinker. It creates link environments as a heuristic (sort of?) and that may possibly lead to sometimes incorrect links. Whether it is better to have incorrect link instead of none is of course debatable, but:

* it would add some twisted logic to hyperlinker
* it happens only in pathological cases
* in *real life* it will be quite uncommon to have interface for one dependency and to not have for another

And once you have interface file and hyperlinked source for `base` (by the way, one more library tested!) `FilePath`s work as expected. So, I think we can live with this.

---

I have finally fixed issue with hyperlinking modules in import statements. I suspect AST does not have `Module` reference in these because multiple packages can share a module (maybe, I don't know). If that's the case, then hyperlink will point to arbitrary file declaring this module.

---

Anyway, I think we are ready for a merge. Or maybe there are some new thoughts?

Fuuzetsu:
I'll try to merge today.

Fuuzetsu:
Please open issues/feature requests discussed throughout this PR and any other you can think of.

We will no doubt mess with the code from this PR here and there in the future but you should now focus on the second part of your GSOC proposal.

PR #411, opened 2015-07-07
StrictData: add support for lazyness marks

adamse:
This ones needs to be merged in order for https://phabricator.haskell.org/D1033 to be merged into GHC.

### DIFF
--- a/haddock-api/src/Haddock/Backends/LaTeX.hs
+++ b/haddock-api/src/Haddock/Backends/LaTeX.hs
@@ -823,8 +823,11 @@ pp_hs_context cxt unicode = parenList (map (ppType unicode) cxt)
 
 
 ppBang :: HsBang -> LaTeX
-ppBang HsNoBang = empty
-ppBang _        = char '!' -- Unpacked args is an implementation detail,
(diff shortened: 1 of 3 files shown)


Fuuzetsu:
I can merge it but I don't know if it's desirable, I mean after all the GHC change is still pending.

Normally what I think happens is that you/someone makes a Haddock branch with your desired change and uses that for phabricator and once everything is OK we put it on `ghc-head`.

I made you a branch on this haskell/haddock repository: `adamse-D1033` you should be able to use its reference for phab. If you want to edit it then you might have to get in touch with @hvr or something to arrange it. The situation is a bit messy right now, might have to work something else out in the future. So for now please use that branch and once it lands I'll be happy to take it into `ghc-head` or for someone from GHC team to do it. Closing this PR.

adamse:
I wasn't quite sure what the protocol was, thank for doing the right thing for me @Fuuzetsu !

Fuuzetsu:
I don't know if it's the right thing but it seems at least saner than updating ghc-head for every potential phab diff touching haddock

Issue #419, opened 2015-07-22
Source links for instances do not point to hyperlinked source


mpickering:
[Here](https://hackage.haskell.org/package/ghc-exactprint-0.3/docs/Language-Haskell-GHC-ExactPrint.html) is a package which demonstrates the problem. All other source links link to the hyperlinked source but the source links attached to instances point to the old-style source code.

Fuuzetsu:
cc @mrhania

mpickering:
From digging around.. it seems like this is expected. Rather than falling back to the old source code maybe it would be best to inject the `line-xxx` divs into the new source representation as well as all the hyperlinking stuff.

Fuuzetsu:
No, I really don't thin it is expected. The `instance` source links were a fairly recent addition and I suspect @mrhania just forgot or didn't know about their existence when implementing the code highlighter/hyperlinker. It should use the new source files and that's' that.

mrhania:
> From digging around.. it seems like this is expected. Rather than falling back to the old source code maybe it would be best to inject the line-xxx divs into the new source representation as well as all the hyperlinking stuff.

Falling back to the old source code is certainly not expected. Generally, once source hyperlinker is enabled, flags `--source-module`, `--source-entity` and `--source-entity-line` (previously provided to make linking to code generated by tools like HsColour possible) are ignored or rather overridden by hyperlinker format. However, until just now hyperlinker was not able to generate anchors so `--source-entity-line` flag was left intact - which you must have provided (or Cabal did it for you for some reason). Haddock even should give you a warning about mixing `--source-*` and `--hyperlinked-source` flags.

Anyway, #420 introduces line anchors, so `--source-entity-line` is now overridden and there should no longer be any reference to non-hyperlinker source.

However, this does not solve this issue. I have run Haddock on `base` and source links for instances point to the wrong lines (line anchors are correct).

> The instance source links were a fairly recent addition and I suspect @mrhania just forgot or didn't know about their existence when implementing the code highlighter/hyperlinker.

Well, hyperlinker is rather independent from source links generation (I haven't really changed the way function source links are generated, just reused existing functionality). Does it always work correctly for HsColour sources (for example, for `base` package)?

mrhania:
[...]

PR #420, opened 2015-07-25
Line anchors for hyperlinked source

mrhania:
As @mpickering mentioned in #419, it would be nice to have line anchors in hyperlinked source. So, here they are.

Note that it does *not* solve #419 - documentation for instances still sometimes refers to the wrong line number. It is not bug in hyperlinker itself so this should be a subject for another PR (once I figure out what's wrong).


Fuuzetsu:
on master

Issue #301, opened 2014-05-27
Record field documentation doesn't align with field name


tibbe:
As can be seen in the screenshot below (look for `flushInterval`) the record field documentation doesn't always align with the field name.

![haddock-issue](https://cloud.githubusercontent.com/assets/17277/3091559/385c1992-e59f-11e3-8cf0-59a032d2c078.png)

```
$ haddock --version
Haddock version 2.14.2, (c) Simon Marlow 2006
Ported to use the GHC API by David Waern 2006-2008
```

Rufflewind:
Getting `<dl>` to display like a table involves compromises: you have to fix the width of one of the columns and if the content overflows that width something bad will happen.  In this particular case, the text pushed the other column downwards.

I suggest scrapping `<dl>` entirely and replacing it with something like:

    <ul>
      <li><dfn>host :: !Text</dfn> Server hostname or IP address</li>
      <li><dfn>port :: !Int</dfn> Server port</li>
      …
    </ul>

and then use `display: table-row` and `display: table-cell` to tabularize them.

PR #421, opened 2015-07-27
Fix record field alignment when name is too long

Rufflewind:
Change `<dl>` to `<ul>` and use `display:table` rather than floats to layout the record fields.  This avoids bug #301 that occurs whenever the field name gets too long.  ([Note: this uses a feature that's not available on IE7 or earlier](http://caniuse.com/#feat=css-table))

~~Slight aesthetic change: the entire field name row is now shaded grey rather than just the area where text exists.~~

Do you prefer having the entire row shaded grey (A), or just the text (B), or the entire cell (C)?

screenshot 2015-07-27 06 06 21

### DIFF
--- a/haddock-api/resources/html/Ocean.std-theme/ocean.css
+++ b/haddock-api/resources/html/Ocean.std-theme/ocean.css
@@ -424,30 +424,31 @@ div#style-menu-holder {
   visibility: hidden;
 }
 
-.subs dl {
+.subs ul {
+  list-style: none;
+  display: table;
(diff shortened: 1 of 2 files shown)


Fuuzetsu:
I like option C most. Then A. B is awful.

Thank you for your PRs as usual. I am busy recently but I hope I can merge them tonight… Sorry for tardiness.

Rufflewind:
Okay, rebased the branch to use C.

PR #425, opened 2015-07-31
Support for typeclass specialization

mrhania:
[Second part](https://ghc.haskell.org/trac/summer-of-code/ticket/1670) of my Google Summer of Code project.

Note that specialization does not generally work for associated types (only partially) - that is because [ClsInst](https://downloads.haskell.org/~ghc/7.8.2/docs/html/libraries/ghc-7.8.2/InstEnv.html#t:ClsInst) object from which Haddock obtains all necessary information does not hold relevant type family data.

### DIFF
--- a/haddock-api/src/Haddock/Backends/Xhtml/Decl.hs
+++ b/haddock-api/src/Haddock/Backends/Xhtml/Decl.hs
@@ -239,28 +239,31 @@ ppSimpleSig links splice unicode qual loc names typ =
 --------------------------------------------------------------------------------
 
 
+ppFamilyInfo :: Bool -> FamilyInfo DocName -> Html
+ppFamilyInfo assoc OpenTypeFamily
+    | assoc = keyword "type"
+    | otherwise = keyword "type family"
+ppFamilyInfo assoc DataFamily
+    | assoc = keyword "data"
+    | otherwise = keyword "data family"
+ppFamilyInfo _ (ClosedTypeFamily _) = keyword "type family"
+
+
+ppFamilyKind :: Unicode -> Qualification -> Maybe (LHsKind DocName) -> Html
+ppFamilyKind unicode qual (Just kind) =
+    dcolon unicode <+> ppLKind unicode qual kind
+ppFamilyKind _ _ Nothing = noHtml
+
+
 ppTyFamHeader :: Bool -> Bool -> FamilyDecl DocName
               -> Unicode -> Qualification -> Html
 ppTyFamHeader summary associated d@(FamilyDecl { fdInfo = info
                                                , fdKindSig = mkind })
               unicode qual =
-  (case info of
-     OpenTypeFamily
-       | associated -> keyword "type"
-       | otherwise  -> keyword "type family"
-     DataFamily
-       | associated -> keyword "data"
-       | otherwise  -> keyword "data family"
-     ClosedTypeFamily _
-                    -> keyword "type family"
-  ) <+>
+    ppFamilyInfo associated info <+>
+    ppFamDeclBinderWithVars summary d <+>
+    ppFamilyKind unicode qual mkind
 
-  ppFamDeclBinderWithVars summary d <+>
-
-  (case mkind of
-    Just kind -> dcolon unicode  <+> ppLKind unicode qual kind
-    Nothing   -> noHtml
-  )
 
 ppTyFam :: Bool -> Bool -> LinksInfo -> [DocInstance DocName] ->
            [(DocName, Fixity)] -> SrcSpan -> Documentation DocName ->
--- a/haddock-api/src/Haddock/Backends/Xhtml/Decl.hs
+++ b/haddock-api/src/Haddock/Backends/Xhtml/Decl.hs
@@ -282,7 +282,7 @@ ppTyFam summary associated links instances fixities loc doc decl splice unicode 
       = subEquations qual $ map (ppTyFamEqn . unLoc) eqns
 
       | otherwise
-      = ppInstances links instances Nothing docname splice unicode qual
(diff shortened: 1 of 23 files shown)


haddock-api/src/Haddock/Backends/Xhtml/Decl.hs
```diff
@@ -490,32 +533,123 @@ ppClassDecl summary links instances fixities loc d subdocs
...
+-- refactoring, for now we just generate naive hash for given instance.
+--
+-- Hashing is very, very trivial and turns a list of 'DocName' to 'Int'. Idea
+-- for such simple hash function (djb2) is stolen from
+-- <http://stackoverflow.com/questions/9262879/create-a-unique-integer-for-each-string here>.
+--
+-- Hashing is performed on string representation of `Name`. Why string instead
+-- of 'Unique' of that 'Name'? That would be much faster and nicer, yes.
+-- However, 'Unique' is not very deterministic, so running it on different
+-- configurations would yield different HTML documents. This is not very bad,
```
Fuuzetsu: No, this *is* very bad on any system/distro that does binary caches and wants deterministic builds.

haddock-api/src/Haddock/Types.hs
```diff
@@ -280,7 +284,16 @@ data DocName
   | Undocumented Name
      -- ^ This thing is not part of the (existing or resulting)
      -- documentation, as far as Haddock knows.
-  deriving Eq
+  deriving (Eq, Data)
+
+type instance PostRn DocName NameSet  = PlaceHolder
```
Fuuzetsu: Why do we need this stuff?

Fuuzetsu:
You should create a ticket on GHC Trac asking them to expose the associate types stuff we're missing. Also create a ticket on here to track it.

Fuuzetsu:
If you click on any types in the instance box, it does not take you to that type but it instead collapses/expands the list. Try http://fuuzetsu.co.uk/misc/lensdocspec/Control-Lens-Each.html#t:Each for example. When I click on `ByteString`, it doesn't do what I expect which is to try to take me elsewhere (or at least give me 404 there because the cross-package links point at my local paths).

Further, I notice the `Source` links in the specialised methods all point at the method in the class definition rather than at the instance.

haddock-api/src/Haddock/Backends/Xhtml/Decl.hs
mrhania: I am not sure what you mean. Can you elaborate a bit?

But anyway, yeah, this is terrible. I could do either something like that or introduce some ordinal number as another parameter to already 10+ argument functions and thread it all the way down. But maybe doing so would be a better choice and then wrap it up nicely during refactoring in the future...

haddock-api/src/Haddock/Types.hs
mrhania: Previously it was defined in `Haddock.Interface.Rename` module (not sure what for) but it was needed elsewhere, so I moved it to `Haddock.Types`. And basically, it makes `DocName` an instance of [`DataId`](https://downloads.haskell.org/~ghc/7.10.1/docs/html/libraries/ghc-7.10.1/PlaceHolder.html#t:DataId) which is required to use SYB stuff with `HsType`.

mrhania:
> You should create a ticket on GHC Trac asking them to expose the associate types stuff we're missing. Also create a ticket on here to track it.

Okay. But I will ask about that on IRC first, maybe there is a way to extract that information and I simply missed that.

> If you click on any types in the instance box, it does not take you to that type but it instead collapses/expands the list.

That should be easy to fix.

> When I click on ByteString, it doesn't do what I expect which is to try to take me elsewhere (or at least give me 404 there because the cross-package links point at my local paths).
> Further, I notice the Source links in the specialised methods all point at the method in the class definition rather than at the instance.

I just mindlessly reused methods for printing class methods. That would explain why source links are not working correctly but type hyperlinks not working is definitely weird.

mrhania:
> If you click on any types in the instance box, it does not take you to that type but it instead collapses/expands the list.

Fixed.

> When I click on ByteString, it doesn't do what I expect which is to try to take me elsewhere (or at least give me 404 there because the cross-package links point at my local paths).

This is not related to type specialization. Notice that `ByteString` and `Text` references in documentation text aren't working either. The link is present and clicking it generates following message in Chrome DevTools:

    Not allowed to load local resource: file:///nix/store/5wl7jzdqz6kspfxi8f8jv5w5sgh6n9cv-ghc-7.10.1/share/doc/ghc/html/libraries/bytestring-0.10.6.0/Data-ByteString.html#t:ByteString

> Further, I notice the Source links in the specialised methods all point at the method in the class definition rather than at the instance.

This may be hard to get right. Essentially, this is the same problem as with class instance source links. For now we can either keep it as it is or scrap these links completely.

---

Oh, and I have also figured out how to remove that dreadful type hashing (and it was way easier than I thought it would be).

Fuuzetsu:
I merged the HTML PRs yesterday, would be great if you could rebase and make sure it all still works.

mrhania:
Rebased, looks like it still works.

bgamari:
It might be useful to attach anchors to the specialized definitions so that a type-aware search tool like `hoogle` could bring you to a specialized definition. I'm not sure if I'd want this behavior from hoogle by default, but it should at least be possible.

mpickering:
Copying over from IRC. 

> It would be good if in instances for `(->)` that the operator was placed infix. 
> https://hackage.haskell.org/package/base-4.8.1.0/docs//Control-Arrow.html

mrhania:
@mpickering: `(->)` (and generally any other type operator) is now sugared to infix style.

Fuuzetsu:
On master

Issue #427, opened 2015-08-06
Test suite failure in AdvanceTypes.hs with GHC 7.10.2


peti:
Citing from http://hydra.cryp.to/build/1047031/nixlog/1/raw:

```
Output for AdvanceTypes.hs has changed! Exiting with diff:
Test suite html-test: FAIL
```

Fuuzetsu:
Yeah, due to GHC change that I failed to catch before release. Will fix in next Haddock release (Soon™), for nixpkgs probably best to disable the tests for Haddock for now.

mrhania:
Just for the record, `PatternSyns` test case is also failing when using GHC 7.10.2, diff:

    204,206c204
    <               <code>
    <                 <a href="#">&gt;&lt;</a>
    <               </code>
    ---
    >               <code>&gt;&lt;</code>

Fuuzetsu:
@mrhania feel free to fix it if you have some minutes, it can take me a while to get around to it

alexbiehl:
I think this can b

PR #431, opened 2015-08-06
Fixes for promoted type printing

mrhania:
As described in #427, `AdvanceTypes` test case was not working after upgrading to GHC 7.10.2. IMHO, neither test output nor test reference was correct - promoted lists and tuples in type signatures were printed within backtick quotes instead of single LISP-like quotation (`'[a, b]`, `'(a, b)`) that should be used in code.

This PR fixes that. Additionally, because `(:)` in type signatures in valid only with promoted types and *has* to be quoted I have added the exception for quoting it as well. And I have renamed `AdvancedTypes` to `PromotedTypes` as it better describes what does that case actually check.

Note that my entire knowledge about type promotion comes from [here](https://downloads.haskell.org/~ghc/7.4.1/docs/html/users_guide/kind-polymorphism-and-promotion.html) so it would be nice if someone more acquainted could confirm what I have stated in this post.


mrhania:
Oh, and to make Travis happy GHC version has to be bumped - in pre-7.10.2 versions, `'[]` does not appear as promoted list type in the AST (that's why tests started to "fail").

Fuuzetsu:
See travis

PR #433, opened 2015-08-10
HsBang is split into HsSrcBang and HsImplBang

adamse:
With recent changes in GHC handling of strictness annotations in Haddock
is simplified.

For when [D1069][1] is to be merged.

[1]: https://phabricator.haskell.org/D1069


bgamari:
Thanks for this!

PR #436, opened 2015-08-20
Follow changes in GHC build system

thomie:
Please merge with ghc-head. I'm doing some refactoring of the GHC build system.

### DIFF
--- a/ghc.mk
+++ b/ghc.mk
@@ -59,9 +59,9 @@ endif
 .PHONY: install_utils/haddock_data
 install_utils/haddock_data:
 	$(foreach i,$(sort $(dir $(utils/haddock_dist_DATA_FILES))), \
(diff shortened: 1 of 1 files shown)


Fuuzetsu:
@thomie I wonder if you could properly set up haddock-library haddock-api and haddock packages in the build system then. If you look at the top-level haddock file, we currently have a flag which is effectively a hack for GHC build system: when we're in GHC tree, we pretend there are no packages and that it's all just one source tree, which is not great.

Will merge this in a few.

thomie:
Making that change would probably mean that haddock-api and haddock-library will be installed in the global package database. Would that a problem in any way?

hvr:
@thomie that'd be undesirable IMHO

thomie:
Why?

PR #437, opened 2015-08-24
Deleted code responsible for: 'hypsrc-test/run.hs:1:2: parse error on…

richardPFisk:
Running ```cabal build -j4``` currently fails with the error ```hypsrc-test/run.hs:1:2: parse error on input ‘#!/’```.

The tiny change in this pull request fixes it.


richardPFisk:
Still an issue when using version 1.22.2.0 of the Cabal library:

It seems it is a [known problem](https://github.com/haskell/cabal/issues/223) with a resolution that has not come through.

PR #438, opened 2015-08-25
Testing framework improvements

mrhania:
Short summary:

  * removes a lot of code duplication in test suites
  * makes test runner more configurable
  * replaces hacky HTML stripping with somewhat nicer XML processing
  * introduces Hoogle backend test suite

It works for me both with Cabal sandboxes and with Stack but a lot can depend on local configuration, so it would be nice if others can confirm that.


bgamari:
Thanks @mrhania and sorry that this went so long without any review. This looks like a great improvement.

Issue #379, opened 2015-03-30
response file expansion doesn't allow spaces in arguments


randen:
I have cherry-picked d567a12b2d24bab610cd7e8f8014d40c7615e24d into haddock 2.16.0 so that we can do the Haskell Platform build for GHC 7.10.1.  To this end, I hacked cabal-install & Cabal to take advantage of the response file addition to haddock, by taking each param/switch/flag/argument and putting one per line in the response file.  Reviewing the change, I was worried about this commit that it parses the response file based on the Prelude's words function.

Trying this change from the above-mentioned commit, a particular argument that could contain some spaces is the "title" argument, which caused a failure as so:

> target `code.' is not a module name or a source file

Looking at the response file, in part it appears as so

> ...
> --odir=.\dist\doc\html\hscolour\.\
> --title=hscolour-1.22: Colourise Haskell code.
> --optghc=-fbuilding-cabal-package
> ...

For my hackish purposes, I change the `words` in the `expandResponse` function to `lines` but this would require that the response file be written with each arg on a separate line, which might be too inflexible and not what would normally be expected of a response file.

23Skidoo:
I guess something like `-foo="Bar Baz"` should be supported.

23Skidoo:
Looks like `gcc` [supports single/double quotes](https://github.com/gcc-mirror/gcc/blob/master/libiberty/argv.c#L135) in response files.

randen:
Is there an existing Haskell implementation of an appropriate parser for busting up unix-like command lines, with quote and/or back-slash-style escaping?  I guess this is what makes handling response files a bit ugly, since normally this is the job of the OS (or the calling program), giving us the argv already split into arguments.

randen:
Generally, supporting "response files" seems to mean one can basically take what would have been the full command line directly (except, of course, the command name itself), stuff that into a file, and feed that to the given executable.  This means worrying about OS-specific parsing of the command line arguments by the receiving program.

Is it too ugly (and I would fully understand if you think it is) to support a sub-set of this functionality, by supporting only a partially digested, sort-of "response files", in that each argument in the response file should be on a separate line (i.e., newline-separated) in the response file?  Essentially, the system call to fork the executable already requires this, with the argument list separated into entries in the list.

[...]

PR #467, opened 2015-12-30

randen:
This is intended to fix haskell/haddock#379. The other half is haskell/cabal#3004, which is not actually complete without the version number of haddock which will support this style of response file.


23Skidoo:
The first two commit look unrelated, I think you need to rebase this against `master`.

randen:
close in favor of a cleaner pull request

23Skidoo:
You can just do `push -f` to destructively update this PR.

randen:
To complete the story for archeologists (including myself--I wasn't able to find it immediately), I should have linked the PR that was used: #470

PR #468, opened 2015-12-30

randen:
* driver/Main.hs
  * The specification for the format of response files
    which gcc generates and consumes, seems to be best
    derived from the gcc code itself (libiberty/argv.c),
    so that is what has been done here.
  * This is intended to fix haskell/haddock#379

### DIFF
--- a/driver/Main.hs
+++ b/driver/Main.hs
@@ -2,6 +2,7 @@
 module Main where
 
 import Control.Exception
+import Data.Char (isSpace)
 import Documentation.Haddock (haddock)
 import System.Environment (getArgs)
 import System.Exit (exitFailure)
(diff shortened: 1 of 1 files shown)


ezyang:
LGTM. I don't know if it's easy to test but some small unit tests would be great.

randen:
Yes, I will re-work this, moving the 'expandResponse' code into a separate module to live in the 'driver' directory, add tests (of the existing test-suites in this package, there doesn't seem to be a place for any driver tests to live, so a new test-suite for this), and update this pull request.  [This functionality really needs to live somewhere else in case other programs need it, but where?]

randen:
I really want to base this off of the v2.16 branch, not master.  I think I need to close this PR in favor of a new one based on the v2.16 branch, unless there is some git/github magic?

randen:
Closing this PR in favor of #470 which is built upon the v2.16 branch.

PR #476, opened 2016-01-17
Render */# instead of TYPE 'Lifted/TYPE 'Unlifted (fixes #473)

### LINKED ISSUE #473: Haddock renders * as TYPE Lifted
An example of it happening can be found in the Haddocks for [`Data.Functor.Product`](http://downloads.haskell.org/~ghc/master/libraries/html/base/Data-Functor-Compose.html). I imagine Haddock also probably renders `#` as `TYPE Unlifted` as well (at least, I'd guess so from looking at this [similar GHC bug](https://phabricator.haskell.org/D1757)).

Could we render it as `TYPE Lifted` as `*`, and `TYPE Unlifted` as `#` instead? They're type synonyms in GHC 8.0 now, so I don't think it would be too much trouble to change, and it would certainly be less confusing for beginners.

RyanGlScott:
An attempt at making Haddock print something a bit nicer for one of the most common kinds you'll see.

### DIFF
--- a/haddock-api/src/Haddock/Convert.hs
+++ b/haddock-api/src/Haddock/Convert.hs
@@ -33,9 +33,9 @@ import TcType ( tcSplitSigmaTy )
 import TyCon
 import Type
 import TyCoRep
-import TysPrim ( alphaTyVars )
-import TysWiredIn ( listTyConName, ipTyCon )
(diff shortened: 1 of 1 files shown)


bgamari:
Looks reasonable to me.

bgamari:
Merged. Thanks @RyanGlScott!

PR #477, opened 2016-01-18
Make sure --mathjax affects all written HTML files

### LINKED ISSUE #475: `--mathjax` does not work consistently
In Debian, the files we ship must not rely on external data, so we cannot use mathjax from the CDN. Luckily, haddock provides a `--mathjax` flag that allows to specify the URL to use, which we do (`--mathjax=file:///usr/share/javascript/mathjax/MathJax.js`).

Unfortunately, it does not work everywhere: It has an effect on `index.html`, but not other HTML pages generated by haddock.

The problem is that in `haddock-api/src/Haddock/Backends/Xhtml.hs`, `maybe_mathjax_url` is not passed around consistently, and sometimes `headHtml` is then called with `Nothing`.

It would probably be (type-)safer to pass this as a `String` and not as a `Maybe String`, and set the default in the argument parser. This way, you cannot “forget” to pass it.

nomeata:
This fixes #475.

### DIFF
--- a/haddock-api/src/Haddock.hs
+++ b/haddock-api/src/Haddock.hs
@@ -292,7 +292,7 @@ render dflags flags qual ifaces installedIfaces extSrcMap = do
 
   when (Flag_GenIndex `elem` flags) $ do
     ppHtmlIndex odir title pkgStr
-                themes opt_contents_url sourceUrls' opt_wiki_urls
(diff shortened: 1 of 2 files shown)


bgamari:
Thanks @nomeata!

idontgetoutmuch:
Thanks also from me @nomeata

PR #479, opened 2016-01-25
Orphans for v2.16

phadej:
Example: http://hackage.haskell.org/package/binary-orphans-0.1.3.0/docs/Data-Binary-Orphans.html

This patch is already in `master`


hvr:
I like :-)

Helkafen:
Just wondering: who is using 2.16 and why wouldn't they use 2.17?

bgamari:
Sebastian Méric de Bellefon <notifications@github.com> writes:

> Just wondering: who is using 2.16 and why wouldn't they use 2.17?
>
Users stuck on GHC 7.10 must also use Haddock 2.16.

PR #480, opened 2016-02-07
Move documentation to ReStructuredText

bgamari:
This drops the old DocBook toolchain in favor of ReStructuredText and Sphinx. This makes the documentation substantially easier to edit and build while sacrificing little of the expressiveness of DocBook.

This was largely motivated by the fact that GHC's DocBook infrastructure, which this code seems to have piggy-backed on, no longer exists.


bgamari:
I'm a bit unsure of whether we want to keep the autoconf infrastructure. I suspect not but I'd be curious to know what others think.

PR #481, opened 2016-02-08
Use -fprint-unicode-syntax when --use-unicode is enabled

bgamari:
This allows GHC to render `*` as its Unicode representation, among other
things.


bgamari:
Merged with 1ee7e9a058f035926ef9d37ef58af409e0c3c998.

PR #482, opened 2016-02-12
Collapse type/data family instances by default

### LINKED ISSUE #474: Type family instances for Rep/Rep1 printed out in full
Now that GHC generics has switched over to a [`-XDataKinds`-based encoding](https://phabricator.haskell.org/rGHC700c42b5e0ffd27884e6bdfa9a940e55449cff6f), the type instances for `Rep` and `Rep1` have become _much_ longer than they used to be, and in just about every Haddock page, they take up an insane amount of width. Look at the Haddocks for `base-4.9`'s [`Control.Applicative`](http://downloads.haskell.org/~ghc/master/libraries/html/base/Control-Applicative.html) for some examples.

This might be a regression, since `base-4.8.1.0`'s [`Control.Applicative`](http://hackage.haskell.org/package/base-4.8.1.0/docs/Control-Applicative.html) doesn't appear to print out any `Rep` or `Rep1` instance implementations.

RyanGlScott:
Fixes #474. Here is what this code:

```haskell
{-# LANGUAGE DeriveGeneric #-}
{-# LANGUAGE GADTs #-}
{-# LANGUAGE TypeFamilies #-}
module Example where

import GHC.Generics

data Letter = A | B | C deriving Generic

type family TFam a
type instance TFam a = Int

data family DFam a b
data instance DFam Int b
  = DFam1 { dfam1 :: Int }
  | DFam2 { dfam2 :: Int }
data instance DFam Char b where
    DFam3 :: { dfam3 :: Char } -> DFam Char Char
    DFam4 :: { dfam4 :: Char } -> DFam Char Char
```

now renders as, with some family instances expanded to show how they are rendered:

haddock-after

### DIFF
--- a/haddock-api/src/Haddock/Backends/Xhtml/Decl.hs
+++ b/haddock-api/src/Haddock/Backends/Xhtml/Decl.hs
@@ -594,20 +594,25 @@ ppInstHead links splice unicode qual mdoc origin orphan no ihd@(InstHead {..}) =
             , [subInstDetails iid ats sigs]
             )
           where
-            iid = instanceId origin no orphan ihd
             sigs = ppInstanceSigs links splice unicode qual clsiSigs
             ats = ppInstanceAssocTys links splice unicode qual clsiAssocTys
         TypeInst rhs ->
-            (ptype, mdoc, [])
+            ( subInstHead iid ptype
+            , mdoc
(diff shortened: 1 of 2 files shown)


bgamari:
Looks good to me. Thanks!

PR #486, opened 2016-02-28
[PRELIMINARY] Document record fields when DuplicateRecordFields is enabled

larskuhtz:
I make this PR to get some feed back on my approach to solve the issue.

I don't fully understand the logic how record fields are documented currently. So I am not sure if this is the right approach. This PR also doesn't solve the issue that explicitly exported record fields are documented with their internal name. Also I don't know a good way to provide test cases for these changes.

### DIFF
--- a/haddock-api/src/Haddock/Interface/Create.hs
+++ b/haddock-api/src/Haddock/Interface/Create.hs
@@ -505,10 +505,10 @@ mkExportItems
     Nothing -> fullModuleContents dflags warnings gre maps fixMap splices decls
     Just exports -> liftM concat $ mapM lookupExport exports
   where
-    lookupExport (IEVar (L _ x))         = declWith x
-    lookupExport (IEThingAbs (L _ t))    = declWith t
-    lookupExport (IEThingAll (L _ t))    = declWith t
-    lookupExport (IEThingWith (L _ t) _ _ _) = declWith t
+    lookupExport (IEVar (L _ x))         = declWith False x
+    lookupExport (IEThingAbs (L _ t))    = declWith False t
+    lookupExport (IEThingAll (L _ t))    = declWith True t
+    lookupExport (IEThingWith (L _ t) _ _ _) = declWith False t
     lookupExport (IEModuleContents (L _ m)) =
       moduleExports thisMod m dflags warnings gre exportedNames decls modMap instIfaceMap maps fixMap splices
     lookupExport (IEGroup lev docStr)  = return $
@@ -522,8 +522,8 @@ mkExportItems
         Nothing -> []
         Just doc -> return . ExportDoc $ processDocStringParas dflags gre doc
 
-    declWith :: Name -> ErrMsgGhc [ ExportItem Name ]
-    declWith t =
+    declWith :: Bool -> Name -> ErrMsgGhc [ ExportItem Name ]
+    declWith withAllSubs t =
       case findDecl t of
         ([L l (ValD _)], (doc, _)) -> do
           -- Top-level binding without type signature
@@ -559,15 +559,15 @@ mkExportItems
                     -- fromJust is safe since we already checked in guards
                     -- that 't' is a name declared in this declaration.
                     let newDecl = L loc . SigD . fromJust $ filterSigNames (== t) sig
-                    in return [ mkExportDecl t newDecl docs_ ]
+                    in return [ mkExportDecl withAllSubs t newDecl docs_ ]
 
                   L loc (TyClD cl@ClassDecl{}) -> do
                     mdef <- liftGhcToErrMsgGhc $ minimalDef t
                     let sig = maybeToList $ fmap (noLoc . MinimalSig mempty . noLoc . fmap noLoc) mdef
-                    return [ mkExportDecl t
+                    return [ mkExportDecl withAllSubs t
                       (L loc $ TyClD cl { tcdSigs = sig ++ tcdSigs cl }) docs_ ]
 
-                  _ -> return [ mkExportDecl t decl docs_ ]
+                  _ -> return [ mkExportDecl withAllSubs t decl docs_ ]
 
         -- Declaration from another package
         ([], _) -> do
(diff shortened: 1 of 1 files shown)


larskuhtz:
Here are examples how record fields are documented with the is PR (all with `DuplicateRecordFields` enabled):

### Example M0

```haskell
module M () where
data R0 = R0 { a0 :: Int, b0 :: Int }
```

not documentation is generated.

### Example M1

```haskell
module M1 (R1(..)) where
data R1 = R1 { a1 :: Int, b1 :: Int }
```

results in

```
data R1
Constructors
    R1	 
        a1 :: Int	 
        b1 :: Int	 
```

#### Example M2

```haskell
module M2 (R2, a2) where
data R2 = R2 { a2 :: Int, b2 :: Int }
```

results in

```
data R2
$sel:a2:R2 :: R2 -> Int
```

# Example M3

```haskell
module M3 (R3(..)) where
data R3
    = R3_1 { a3 :: Int, b3 :: Int }
    | R_2 { a3 :: Int, c3 :: Int }
```

results in

```
data R3
Constructors
    R3_1	 
        a3 :: Int	 
        b3 :: Int	 
    R_2	 
        a3 :: Int	 
        c3 :: Int	 
```

### Example M4

```haskell
module M4 (R4(R4_1)) where
data R4
    = R4_1 { a4 :: Int, b4 :: Int }
    | R4_2 { a4 :: Int, c4 :: Int }
```

results in

```
data R4
Constructors
    R4_1 Int Int
```

### Example M5

```haskell
module M5 (R5(R5_1), c5) where
data R5
    = R5_1 { a5 :: Int, b5 :: Int }
    | R5_2 { a5 :: Int, c5 :: Int }
```

results in

```
data R5
Constructors
    R5_1 Int Int
$sel:c5:R5_1 :: R5 -> Int
```

### Example M6

```haskell
module M6 (R6(R6_1), a6) where
data R6
    = R6_1 { a6 :: Int, b6 :: Int }
    | R6_2 { a6 :: Int, c6 :: Int }
```

results in

```
data R6
Constructors
    R6_1 Int Int
```

### Example M7

```haskell
module M7 (R7(R7_1, a7, b7, c7))
data R7
    = R7_1 { a7 :: Int, b7 :: Int }
    | R7_2 { a7 :: Int, c7 :: Int }
```

results in

```
data R7
Constructors
    R7_1 Int Int
```

### Example M8

```haskell
module M8 (R8(R8_1), a8, b8, c8)
data R8
    = R8_1 { a8 :: Int, b8 :: Int }
    | R8_2 { a8 :: Int, c8 :: Int }
```

results in

```
data R8
Constructors
    R8_1 Int Int	 
$sel:c8:R8_1 :: R8 -> Int
```

### Example M9

```haskell
module M9( R9(R9_1, R9_2), a9, b9, c9)
data R9
    = R9_1 { a9 :: Int, b9 :: Int }
    | R9_2 { a9 :: Int, c9 :: Int }
```

results in

```
data R9
Constructors
    R9_1 Int Int	 
    R9_2 Int Int
```

haddock-api/src/Haddock/Interface/Create.hs
```diff
@@ -582,23 +582,31 @@ mkExportItems
...
                 Just iface ->
-                   return [ mkExportDecl t decl (lookupDocs t warnings (instDocMap iface) (instArgMap iface) (instSubMap iface)) ]
+                   return [ mkExportDecl withAllSubs t decl (lookupDocs t warnings (instDocMap iface) (instArgMap iface) (instSubMap iface)) ]
 
         _ -> return []
 
 
-    mkExportDecl :: Name -> LHsDecl Name -> (DocForDecl Name, [(Name, DocForDecl Name)]) -> ExportItem Name
-    mkExportDecl name decl (doc, subs) = decl'
+    mkExportDecl :: Bool -> Name -> LHsDecl Name -> (DocForDecl Name, [(Name, DocForDecl Name)]) -> ExportItem Name
```
bgamari: For the sake of clarity I think this should be a dedicated type instead of a plain `Bool`.

bgamari:
Hi @larskuhtz,

Thanks for looking at this.

It would be very helpful to me if you could you start by describing how your proposed behavior differs from what Haddock currently offers. I'm not sure I understand this at the moment.

> Also I don't know a good way to provide test cases for these changes.

The usual approach here is to add a module to the html testsuite which uses the feature in question (e.g. `DuplicateRecordFields`). The testsuite driver will then compare the HTML output produced by Haddock against a reference; it's far from a perfect system but it's what we have.

bgamari:
Any progress on this? It would be nice to have this for 8.0.1 which is imminent.

tfausak:
GHC 8.0.1 has been out for about a month now. I would love for Haddock to be able to generate the correct documentation for duplicate record fields. 

For an example of what Haddock does currently, check out [this data type](https://hackage.haskell.org/package/octane-0.11.0/docs/Octane-Type-RawReplay.html#t:RawReplay). The documentation makes it look like there are no named fields, but [the source](https://hackage.haskell.org/package/octane-0.11.0/docs/src/Octane-Type-RawReplay.html#RawReplay) shows that they are not only named but also documented.

joneshf:
What do we do at this point?

alexbiehl:
@larskuhtz are you still interested in this PR? Maybe someone else could step up and resolve this issue. 

Also by looking at the examples from https://github.com/haskell/haddock/pull/486#issuecomment-189933387 I am not sure those excercise the `DuplicateRecordFields` extension. Since all the duplicate selectors belong to the same data type. 


```haskell
data X = A { a :: Int } | B { a :: Int }

data Y = C { a :: Int }
```
would be a test case where `DuplicateRecordFields` are needed. 

After all an explanation about differences to the current algorithm is needed as @bgamari already said.

alexbiehl:
This is in fact another occurrence of our brittle export logic. See https://github.com/haskell/haddock/issues/586.

alexbiehl:
This is subsumed by #649.

PR #493, opened 2016-03-28
Fix #492

### LINKED ISSUE #492: Collapsed sections are not stored properly in cookie
The javascript for collapsing and opening a section stores a cookie for remembering which sections was collapsed, so it can close those again the next time the page is loaded, which it does by toggling it's "collapsed" state. The problem is that some sections are collapsed by default, so when it tries to collapse it, it will instead be opened.

You can see this by opening the synopsis, closing it and then reloading the page. The synopsis will be open again.

anka-213:
Now the "collapsed" cookie stores which sections have changed state instead of which are collapsed.

### DIFF
--- a/haddock-api/resources/html/haddock-util.js
+++ b/haddock-api/resources/html/haddock-util.js
@@ -52,17 +52,17 @@ function toggleSection(id)
 {
   var b = toggleShow(document.getElementById("section." + id));
   toggleCollapser(document.getElementById("control." + id), b);
-  rememberCollapsed(id, b);
+  rememberCollapsed(id);
   return b;
 }
 
 var collapsed = {};
-function rememberCollapsed(id, b)
+function rememberCollapsed(id)
 {
-  if(b)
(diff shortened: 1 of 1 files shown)


Helkafen:
Thanks for the fix! What about adapting the naming?
For instance: 
- "collapsed" cookie name and dictionary -> "toggled"
- rememberCollapsed -> rememberToggled
- restoreCollapsed -> restoreToggled

Helkafen:
BTW I can't explain why the synopsis button is hidden with my local build. Is that a regression?
Steps:
```stack unpack pipes```
```cd pipes-4.2.0```
```cabal sandbox init ; cabal configure ; cabal install --only-dependencies; cabal configure```
```cabal haddock --with-haddock=/home/seb/.local/bin/haddock --haddock-options="--hyperlinked-source"```

alexbiehl:
I rebased and merged this in #608.

PR #494, opened 2016-04-19
Display bundled pattern synonyms as constructors.

mpickering:
Need to do a proper review but don't have time at the moment.

### DIFF
--- a/haddock-api/src/Haddock/Interface/Create.hs
+++ b/haddock-api/src/Haddock/Interface/Create.hs
@@ -505,10 +505,12 @@ mkExportItems
     Nothing -> fullModuleContents dflags warnings gre maps fixMap splices decls
     Just exports -> liftM concat $ mapM lookupExport exports
   where
-    lookupExport (IEVar (L _ x))         = declWith x
-    lookupExport (IEThingAbs (L _ t))    = declWith t
-    lookupExport (IEThingAll (L _ t))    = declWith t
-    lookupExport (IEThingWith (L _ t) _ _ _) = declWith t
+    lookupExport (IEVar (L _ x))         = declWith x Nothing
+    lookupExport (IEThingAbs (L _ t))    = declWith t Nothing
+    lookupExport (IEThingAll (L _ t))    = declWith t Nothing
+    lookupExport (IEThingWith (L _ t) _ cs _) = do
+      csDecls <- mapM (bundledPatToCon . unLoc) cs
+      declWith t (Just (concat csDecls))
     lookupExport (IEModuleContents (L _ m)) =
       moduleExports thisMod m dflags warnings gre exportedNames decls modMap instIfaceMap maps fixMap splices
     lookupExport (IEGroup lev docStr)  = return $
@@ -522,8 +524,12 @@ mkExportItems
         Nothing -> []
         Just doc -> return . ExportDoc $ processDocStringParas dflags gre doc
 
-    declWith :: Name -> ErrMsgGhc [ ExportItem Name ]
-    declWith t =
+    declWith :: Name
+             -> Maybe [(LConDecl Name,(Name, DocForDecl Name))]
+             -- ^ "Maybe" a list of bundled pattern synonyms, including their
+             -- documentation, encoded as GADT constructors.
+             -> ErrMsgGhc [ ExportItem Name ]
+    declWith t csM =
       case findDecl t of
         ([L l (ValD _)], (doc, _)) -> do
           -- Top-level binding without type signature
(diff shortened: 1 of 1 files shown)


haddock-api/src/Haddock/Interface/Create.hs
```diff
@@ -52,6 +52,8 @@ import BasicTypes ( StringLiteral(..) )
 import qualified Outputable as O
 import HsDecls ( gadtDeclDetails,getConDetails )
 
+import DynFlags (unsafeGlobalDynFlags)
```
bgamari: This isn't necessary, right?

mpickering: No.

haddock-api/src/Haddock/Interface/Create.hs
```diff
@@ -522,8 +526,8 @@ mkExportItems
         Nothing -> []
         Just doc -> return . ExportDoc $ processDocStringParas dflags gre doc
 
-    declWith :: Name -> ErrMsgGhc [ ExportItem Name ]
-    declWith t =
+    declWith :: Name -> Maybe [(LConDecl Name,(Name, DocForDecl Name))] -> ErrMsgGhc [ ExportItem Name ]
```
bgamari: A comment explaining `csM` would be nice.

haddock-api/src/Haddock/Interface/Create.hs
```diff
@@ -582,15 +586,15 @@ mkExportItems
...
                 Just iface ->
-                   return [ mkExportDecl t decl (lookupDocs t warnings (instDocMap iface) (instArgMap iface) (instSubMap iface)) ]
+                   return [ mkExportDecl t csM decl (lookupDocs t warnings (instDocMap iface) (instArgMap iface) (instSubMap iface)) ]
 
         _ -> return []
 
 
-    mkExportDecl :: Name -> LHsDecl Name -> (DocForDecl Name, [(Name, DocForDecl Name)]) -> ExportItem Name
-    mkExportDecl name decl (doc, subs) = decl'
+    mkExportDecl :: Name -> Maybe [(LConDecl Name,(Name, DocForDecl Name))] -> LHsDecl Name -> (DocForDecl Name, [(Name, DocForDecl Name)]) -> ExportItem Name
```
bgamari: Same here.

thomie:
See also https://ghc.haskell.org/trac/ghc/ticket/11954.

haddock-api/src/Haddock/Interface/Create.hs
```diff
@@ -602,6 +612,34 @@ mkExportItems
...
+        case export of
+          (ExportDecl (L l' (SigD (PatSynSig (L _ n) ty))) doc' _ _ _ _) ->
+            return [(L l' (ConDeclGADT [L l' n] ty Nothing),(n,doc'))]
+          _ -> return []
+      ([L _ (SigD (PatSynSig (L l' n) ty))], (doc, _)) ->
+        return [(L l' (ConDeclGADT [L l' n] ty Nothing),(n,doc))]
+      _ -> return []
+
+
+    -- Add bundled-pattens-converted-to-constructors to a data type declaration
```
ezyang: pattern

ezyang:
The approach seems OK to me; you've eyeballed the output? There probably should be a test?

christiaanb:
I want to note that I wrote this patch as a "better-something-than-nothing" patch for one of the 8.0.1 release canditates. With that out of the way, the haddock output produced by this patch can be ackward:

(repeating what I said in: https://github.com/haskell/haddock/issues/551#issuecomment-244099120)
So let's say you have something silly as:

```haskell 
data Either a b = Left a | Right b

pattern Both :: a -> b -> Either a b
```

Then the haddock would be printed as:

`data Either a b`

__Constructors__
__`Left`__ `a`
__`Right`__ `b`
__`Both`__ `:: forall a b . a -> b -> Either a b`


There are several things to notice:

* The explicit forall quantifier are also rendered in normal pattern synonym haddock. I think this is a bug, see also: https://ghc.haskell.org/trac/ghc/ticket/11955
* The reason I print pattern synonyms as GADT constructors is because they can have Contexts/Contraints (which normal const

TODO: my mouse cursor disappeared, so I'll post this comment in a half-done state and update it later.

harpocrates:
What are the next steps forward for this issue? It doesn't look like this will make it for 8.2, eh?

I'm currently writing a library where I intend to have a lot of data definitions that are pretty complex looking, but I'd like to offer the user the opportunity to only look at simpler looking patterns. For example, something like:

```Haskell
-- Expr/Internal.hs
module Expr.Internal (Expr_(Addition_, Literal_)) where

data Expr_ m a
  = Addition_ (m (Expr_ m a)) (m (Expr_ m a)) a
  | Literal_ (m Int) a

-- Expr.hs
module Expr (Expr(Addition, Literal)) where

type Expr a = Expr_ Identity a 

pattern Addition :: Expr a -> Expr a -> a -> Expr a
pattern Addition l r x = Addition_ (Identity l) (Identity r) x

pattern Literal :: Int -> a -> Expr a
pattern Literal i x = Literal_ (Identity i) x
```

I'm not really holding out for being able to export `Expr` as a type constructor instead of just as a type synonym (I'm pretty convinced that would in general not be good), but at least having `Addition` and `Literal` appear as constructors (even GADT-style) would be great. Is there anything I could do/contribute to help push this issue forward?

christiaanb:
@harpocrates I've written #627 which supersedes this patch. I guess depending on _when_ that is merged into haddock master I might be able to convice @bgamari to include it in the 8.2.1 release; seems unlikely though.

alexbiehl:
Closing this as we now have #627 merged.

PR #500, opened 2016-05-11
Issue 303

Helkafen:
Hide footer when printing.

### DIFF
--- a/haddock-api/resources/html/Classic.theme/xhaddock.css
+++ b/haddock-api/resources/html/Classic.theme/xhaddock.css
@@ -116,6 +116,11 @@ ul.links li {
   cursor: pointer;
 }
 
+@media print {
+  .noprint
+  { display: none; }
+}
+
 #package-header {
 	color: #ffffff;
 	padding: 5px 5px 5px 31px;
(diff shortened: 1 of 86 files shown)


bgamari:
Out of curiosity: Why not simply add a CSS rule for `#footer` with the appropriate media type?

bgamari:
e.g.,
```
@media print {
  #footer { display: none; }
}
```

This better matches with my intuitive understanding of how CSS is intended to be used:  separate document structure from presentation.

Helkafen:
Your version is cleaner. I was thinking of reusing the noprint class someday, but I can't find any practical use right now.

bgamari:
Looks good to me. However, could you be sure to squash the two commits into one with a proper commit message prior to merging?

Helkafen:
Tried to, but github doesn't let me force push the squashed commit.

hvr:
that's because you based thir pr off a branch in haskell/haddock which is protected against forcepushes... if you  create PRs from forked repos it would work....

bgamari:
Herbert Valerio Riedel <notifications@github.com> writes:

> that's because you based thir pr off a branch in haskell/haddock which
> is protected against forcepushes... if you create PRs from forked
> repos it would work....
>
Indeed, in general I work in my own forks whenever possible; if nothing
else it minimizes the chance for mistakes (which I am quite prone to it seems).

Helkafen:
Ok, I'll also work from my own fork then. Thanks for the explanation.

PR #502, opened 2016-05-14
Fix #303. Hide footer when printing

### LINKED ISSUE #303: The "produced by haddock" div and printing
The generated HTML has a footer, a `<div>`, containing the text

>Produced by Haddock version N

When printing a generated page this footer is positioned at the bottom of the first page, but the running text doesn't pay any attention to this.  The effect is that, the footer covers the last few lines of text on the first page.  Not so good.

The following picture hopefully illustrates the situation.  It's the generated documentation for Blaze.ByteString.Builder, loaded into Firefox (29.0.1) and then printed to PDF.  All done on Linux.

blaze-bytestring-builder-pdf

Helkafen:
The "Produced by Haddock" footer was overlapping the page's body when printing.
This patch hides the footer with a css media rule.

### DIFF
--- a/haddock-api/resources/html/Classic.theme/xhaddock.css
+++ b/haddock-api/resources/html/Classic.theme/xhaddock.css
@@ -116,6 +116,10 @@ ul.links li {
   cursor: pointer;
 }
 
+@media print {
+  #footer { display: none; }
+}
+
 #package-header {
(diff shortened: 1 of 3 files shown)


bgamari:
Indeed, looks good to me!

PR #504, opened 2016-05-15
Fix #280. Parsing of module header

### LINKED ISSUE #280: Wrong behaviour for multi-line fields not placed at the very beginning
**Original reporter**: *fuuzetsu@\<hidden\>*

Consider module header such as

```haskell 
{-|Copyright: Foo,
              Bar,
              Baz
-}
module F where
```

This does what we expect, we get the Copyright field with ‘Foo, Bar, Baz’.

If we then try to make this slightly less ugly and go with
```haskell 
{-|
Copyright: Foo,
           Bar,
           Baz
-}
module F where
```

the output is different: Copyright now has ‘Foo,’ and ‘Bar, Baz’ become the module description.

Helkafen:
The initial newlines were counted as indentation spaces, thus disturbing the parsing of next lines

### DIFF
--- a/haddock-api/src/Haddock/Interface/ParseModuleHeader.hs
+++ b/haddock-api/src/Haddock/Interface/ParseModuleHeader.hs
@@ -76,7 +76,7 @@ parseKey :: String -> String -> Maybe (String,String)
 parseKey key toParse0 =
    do
       let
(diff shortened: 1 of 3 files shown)


bgamari:
Looks reasonable to me.

Issue #510, opened 2016-05-23
Template haskell: incorrect highlighting with hyperlinked source (especially no anchors!)


phadej:
<img width="405" alt="screen shot 2016-05-23 at 10 31 29" src="https://cloud.githubusercontent.com/assets/51087/15463359/8ff37b18-20d1-11e6-8525-b0d61b46e514.png">

Issue #508, opened 2016-05-23
Internal error while processing `swagger2-2.0.2`


phadej:
and current master 863b414893da1db41da6a6e22dc4933684e29da9

```
    Warning: Data.Swagger.Lens: could not find link destinations for:
        ~
    Warning: Data.Swagger.Internal.Schema: could not find link destinations for:
        ~
    Warning: Data.Swagger.Schema: could not find link destinations for:
        ~
    haddock: internal error: spliceURL UnhelpfulSpan
```

*EDIT* happens when trying to find line to print `Source` for instances methods:

```diff
-      UnhelpfulSpan _ ->
-        error "spliceURL UnhelpfulSpan"
+      UnhelpfulSpan _ -> "very-weird-line"
```

<img width="899" alt="screen shot 2016-05-23 at 10 09 58" src="https://cloud.githubusercontent.com/assets/51087/15462919/907f33d6-20ce-11e6-92bf-03c34c982448.png">

IIRC those are `template-haskell` derived classes and instances: so the definition site is not-so-clear. The easy fix would be to drop `line-` anchor for now (i.e. make `line` variable of type `Maybe String`), I'm no idea how to make proper fix, or if it possible inside `haddock` at all.

*EDIT2* The above `consumes` link has the correct `line` anchor. Links from external modules have the format of `.../doc/swagger2-2.0.2/src/Data.Swagger.Lens.html#consumes` i.e. by name!

phadej:
I'm quite sure https://github.com/haskell/haddock/issues/505 is the same, as I noticed that the error is not always printed while trying to workaround the issue.

shajra:
I see the fix has merged into ghc-head.  I suppose I could install a bleeding-edge haddock+ghc to generate docs, but it would be nice if there was some known workaround to hobble along until GHC 8.0.2 is released.  Any ideas?

steshaw:
I'm getting bitten by this `UnhelpfulSpan`. I thought the fix might have found it's way into 2.17.3 but upgrading from 2.17.2 to 2.17.3 doesn't fix it.

Is there another way to work around the problem? I'm pulling in Haddock via Stack on lts-8.5.

alexbiehl:
A new release is on its way. But you can try this
https://github.com/haskell/haddock/tree/2.17.3.1-spanfix until we cut a
release.

Steven Shaw <notifications@github.com> schrieb am Mo., 3. Apr. 2017, 07:33:

steshaw:
[...]

PR #512, opened 2016-05-24
Fixes for #508 and #510

phadej:
First commit works around https://github.com/haskell/haddock/issues/508
Second commit works around https://github.com/haskell/haddock/issues/510
before/after screenshots:

haddock-before
haddock-after


mpickering:
Looks good to me, thanks Oleg.

bgamari:
Indeed, looks good to me as well.

phadej:
@mpickering I'd like to see `haddock` using GHC for this; I only struggle if it viable in the presence of CPP. You probably know better from `ghc-exactprint` experience?

PR #513, opened 2016-05-24
Make parser state a newtype

treeowl:
Previously, it was `data` wrapping a `Maybe`, which seems a bit
silly. Obviously, this can be changed back if anyone wants to add
more fields some day.


bgamari:
Sure, why not. Thanks!

Issue #274, opened 2014-05-08
Frames broken when using separate contents page


ghc-mirror:
**Original reporter**: *duncan*

The frames view for the docs on hackage has been broken since forever. In the past what we've done is just disable the frames view. The other alternative would be to fix it.

The symptom is that the `index-frames.html` file is missing, and of course that's one of the pages used in a sub-frame (top left).

I **think** that what is happening is that we call `cabal haddock --contents-location=URL` which calls haddock with `--use-contents=URL`, and then haddock does not create `index.html` (as that's not used) but then also omits the `index-frames.html` file.

So if that's right then the solution is to adjust the logic so that it only omits generation of `index.html` and does not omit `index-frames.html`.

That said, we're not sure we actually want the frames view on hackage, so an alternative feature request is for a `--no-frames` flag, that excludes that. Currently we have to hack the .js file to not add the frames link.

Issue #114, opened 2014-05-08
Make the frames version a separate output mode


ghc-mirror:
**Original reporter**: *duncan*

The frames version of the html haddock output is almost universally ignored. That is the files:
```haskell 
frames.html
index-frames.html
mini_$MODULE.html
```
which are generated to support the frames view.

Since nobody uses the frames view and it takes extra space and clutter, how about making it a separate mode.

```haskell 
--html         output in HTML
--html-frames  output in HTML using frames
```

Note: when extracting the frames code from the normal code, make sure to omit the on-load code from the html head:
```haskell 
<SCRIPT TYPE="text/javascript">
window.onload = function () {
  setSynopsis("mini_Distribution-Simple.html")
};
</SCRIPT>
```

Rufflewind:
[Frames are obsolete in HTML5](http://www.w3.org/TR/html5-diff/#obsolete-elements), so eliminating them is probably a good idea in the long run.  That being said, a (collapsible?) navigation panel can still be useful.

PR #514, opened 2016-05-25
remove framed view of the HTML documentation (see #114 and #274)

Helkafen:
Frames are a bit broken, ignored by Hackage, and considered obsolete in general.
This patch disables frames generation. The mini_*.html files are still used in the synopsis.


bgamari:
Sure, sounds fair enough to me. However, make sure you add a note in `CHANGES`.

Helkafen:
Sure.

PR #538, opened 2016-06-30
Add $ as a special character

massysett:
If this character is not escaped, documentation built with Haddock
2.17.2 will fail.  This was not an issue with 2.16 series, which
causes builds to fail and there is nothing in the docs or error
message giving a clue about why builds that used to succeed now
don't.

### DIFF
--- a/doc/markup.rst
+++ b/doc/markup.rst
@@ -501,7 +501,7 @@ Special characters
 ~~~~~~~~~~~~~~~~~~
 
 The following characters have special meanings in documentation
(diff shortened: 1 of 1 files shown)


Fuuzetsu:
$ is used for named comments: you can write

```
module Foo
  ( a
  , b
  -- $someName
  , c
  ) where
```

and at the bottom or wherever it pleases you

```
-- $someName This is some large comment
```

I'll merge this because you're right, it is a special character. I don't know why it would build before but fail now however. It is likely due to some change either in GHC parsing or Haddock itself which I'm not aware of.

PR #539, opened 2016-07-04
More general type for nameCacheFromGhc.

carlohamalainen:
More general type for  nameCacheFromGhc.

Example of where I'd use it: https://github.com/carlohamalainen/ghc-mod/blob/ghc-imported-from-merge/Language/Haskell/GhcMod/HaddockLookup.hs#L443-L455


alexbiehl:
A little late but I don't see why this shouldn't go in.

carlohamalainen:
Thanks :)

Issue #549, opened 2016-08-22
Rendered haddock of `Eq` and `Ord` classes are missing class methods


christiaanb:
The rendered haddock (included in the GHC 8.0.1 release) for the `Eq` and `Ord` classes are missing:
- Their class methods
- Their source links

See:

​http://hackage.haskell.org/package/base-4.9.0.0/docs/Data-Eq.html#t:Eq
​http://hackage.haskell.org/package/base-4.9.0.0/docs/Prelude.html#t:Eq
​http://hackage.haskell.org/package/base-4.9.0.0/docs/Data-Ord.html#t:Ord
​http://hackage.haskell.org/package/base-4.9.0.0/docs/Prelude.html#t:Ord

Other (deriviable) classes, for example Show, or Enum, do not have these problems.

I've also reported this as a GHC issue: https://ghc.haskell.org/trac/ghc/ticket/12519

k-bx:
In https://github.com/haskell/haddock/issues/552 duplicate I've provided a repo with easy reproduction steps.

niteria:
I have a fix here, but due to poor internet it's difficult for me to commit/send a pull request. https://ghc.haskell.org/trac/ghc/ticket/12519#comment:6

ysangkok:
I think this was fixed with https://github.com/haskell/haddock/commit/240bc38b94ed2d0af27333b23392d03eeb615e82 ?

PR #561, opened 2016-11-29
Fix haddock: internal error: links: UnhelpfulSpan

### LINKED ISSUE #554: haddock: internal error: links: UnhelpfulSpan
```
$ cat Protocol.hs
module Protocol
 (  Protocol(..) ) where

import Data.Proxy
type Type    = Int
class Protocol p where
  getByteType   :: Proxy p -> Type

$ cat Compact.hs
module Compact
  ( Compact ) where

import Protocol
import Data.Proxy

data Compact
instance Protocol Compact where

$ HADDOCK=/Users/bnitka/ghc/inplace/bin/haddock
$ $HADDOCK --html --source-entity '%{FILE}$%{LINE}' Protocol.hs Compact.hs
```

results in:

```
Haddock coverage:
   0% (  0 /  2) in 'Protocol'
  Missing documentation for:
    Module header
    Protocol (Protocol.hs:6)

Compact.hs:8:10: warning: [-Wmissing-methods]
    • No explicit implementation for
        ‘getByteType’
    • In the instance declaration for ‘Protocol Compact’
   0% (  0 /  2) in 'Compact'
  Missing documentation for:
    Module header
    Compact (Compact.hs:7)
Warning: Protocol: could not find link destinations for:
    Proxy Type *
Warning: Compact: could not find link destinations for:
    Proxy * Type
haddock: internal error: links: UnhelpfulSpan
  error, called at utils/haddock/haddock-api/src/Haddock/Backends/Xhtml/Layout.hs:299:30 in main:Haddock.Backends.Xhtml.Layout
```

It fails with GHC HEAD, I have a branch not much different than GHC 8.0.1 and it fails there as well. It worked with GHC 7.10.3.

niteria:
This fixes #554 for me. I believe this is another fall out
of `wildcard-refactor`, like #549.


niteria:
I'm not 100% sure this is the way we want to fix it, but I have this patch lying around and for a lack of a better alternative I'm publishing it here.

haddock-api/src/Haddock/Convert.hs
```diff
@@ -311,7 +311,7 @@ synifyDataCon use_gadt_syntax dc =
                          , con_doc =  Nothing }
 
 synifyName :: NamedThing n => n -> Located Name
-synifyName = noLoc . getName
+synifyName n = L (srcLocSpan (getSrcLoc n)) (getName n)
```
bgamari: This looks reasonable.

haddock-api/src/Haddock/Backends/Xhtml/Decl.hs
```diff
@@ -646,7 +646,7 @@ ppInstanceSigs links splice unicode qual sigs = do
     TypeSig lnames typ <- sigs
     let names = map unLoc lnames
         L loc rtyp = get_type typ
-    return $ ppSimpleSig links splice unicode qual loc names rtyp
+    return $ ppSimpleSig links splice unicode qual (getLoc $ head $ lnames) names rtyp
```
bgamari: I'm not so sure about this. Why not use `loc`, the span associated with the type? It likely makes little difference in the result, but it would be nice to understand why this change is necessary.

alexbiehl:
@blackgnezdo please take a look at haddock release 2.17.5 (http://hackage.haskell.org/package/haddock-2.17.5/changelog). You will find a fix to the problem there. I was thinking about reviewing this PR though as this seems to be a better solution to the problem.

niteria:
If you have suggestions on how to improve this patch and make it mergeable I'm happy to do the legwork.

blackgnezdo:
@alexbiehl my apologies, I failed to include the version information. I am already using v2_17_5. It exhibits the bug without this patch.

alexbiehl:
@blackgnezdo Interesting! Have you got a reproducer?

blackgnezdo:
I'll have to untangle it from the internal codebase. Time is an issue as usual...

haddock-api/src/Haddock/Convert.hs
alexbiehl: @niteria I understand this change. It is a great one! We were previously assuming `noLoc` where we now use the location of a name, great 100% more information!

haddock-api/src/Haddock/Backends/Xhtml/Decl.hs
alexbiehl: @niteria can you explain this change here? Would it make sense to extend the `sinify` business to give more meaningful src locs if possible to avoid not doing the obvious thing here?

bpfoley:
We can reproduce it with haddock 2.7.14 with
```haskell
module Test(Addable(..)) where

class Addable a where
  plus :: a -> a -> a

instance Addable Integer where
  plus = (+)
```

```
$ haddock --version
Haddock version 2.17.4, (c) Simon Marlow 2006
```

```
$ haddock -v0 --html --source-entity="http://x/%{FILE}" -o /tmp Test.hs

Warning: Test: could not find link destinations for:
    Integer
haddock: internal error: links: UnhelpfulSpan
  error, called at src/Haddock/Backends/Xhtml/Layout.hs:299:30 in haddock-api-2.17.4-2oc6jdou1s23DzKuE5Kc7x:Haddock.Backends.Xhtml.Layout
```

alexbiehl:
I have rebased and and added a comment. Thanks @niteria!

niteria:
Thanks! Sorry I missed your inline comment, I must have my github notifications set up wrong.

alexbiehl:
@niteria No problem @niteria.  I think I understand the issue now. Thanks for your patch!

PR #564, opened 2016-12-31
Fix test failures on Windows

igrep:
I'm trying to inspect another bug on Windows.
Before looking into it, I found some of the tests fail due to Windows-specific reasons:
- The difference in the line separator (CRLF and LF).
- The difference in the null file.
So I've fixed.

Note there's still a test failure in `html-test/Main.hs` only on Windows, but it seems to be avoided by `chcp 65001`.
It looks similar to https://github.com/commercialhaskell/stack/issues/793


alexbiehl:
@igrep As soon as this is green I will merge. Thank you and sorry for the long delay.

PR #568, opened 2017-01-12
Prevent GHC API from doing optimization passes.

awson:
Currently GHC HEAD's Haddock segfaults on Windows 64-bit when generating documentation for `language-c-quote-0.11.7.1` package. I believe, this is related to GHC bytecode generation/linking issues (I consistently see access violation on executable data at virtual addresses above 4GB mark). Btw, GHC segfaults too, but after repeated run, when it hasn't recompile all files, but only part of them, it succeeds.

I've examined what flags `cabal` supplies to Haddock and was surprised seeing `--optghc=-O` there. When I've rerun Haddock manually with no `--optghc=-O` flag, it succeeded. Moreover, it (obviously) was much faster.

Thus this patch makes Haddock working on Windows 64-bit for some complicated cases and also have a nice side-effect making Haddock much faster on other platforms.

### DIFF
--- a/haddock-api/src/Haddock.hs
+++ b/haddock-api/src/Haddock.hs
@@ -398,7 +398,7 @@ withGhc' libDir flags ghcActs = runGhc (Just libDir) $ do
     ghcMode   = CompManager,
     ghcLink   = NoLink
     }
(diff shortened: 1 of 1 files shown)


alexbiehl:
Good catch @awson! Sorry it took so long to merge

Issue #530, opened 2016-06-15
Add markup support for tables


hvr:
Providing support for simple tables would go a long way IMO,

Tables could be detected by looking for a paragraph whose first columns form a sequence of

 `'+'` (`'|'`{1..n}`'+'`){1..m}

This would allow writing tables which look like

``` hs
{- | Lorem ipsum dolor sit amet, consectetur adipiscing elit

sed do eiusmod tempor incididunt ut labore et dolore magna aliqua

+----------+----------+----------+
|  32bit   |  32bit   |   64bit  | 
+==========+==========+==========+
|  0x0000  |  0x0000  |  0x0000  |
+----------+----------+----------+

Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.

+------+--------------+------------------------------------------+
| code | message      | description                              |
+======+==============+==========================================+
| 200  |   @OK@       | operation successful                     |
+------+--------------+------------------------------------------+
| 204  | @No Content@ | operation successful, no body returned   |
+------+--------------+------------------------------------------+

-}
```

andrewthad:
This would be pretty sweet.

PR #577, opened 2017-02-17
Add markup support for tables

potomak:
Tables are composed by an optional header and body. The header is
composed by a single row. The body is composed by a non-empty list of
rows.

Example table with header:

    +------+--------------+------------------------------------------+
    | code | message      | description                              |
    +======+==============+==========================================+
    | 200  |   @OK@       | operation successful                     |
    +------+--------------+------------------------------------------+
    | 204  | @No Content@ | operation successful, no body returned   |
    +------+--------------+------------------------------------------+

Example table without header:

    +------+--------------+------------------------------------------+
    | 200  |   @OK@       | operation successful                     |
    +------+--------------+------------------------------------------+
    | 204  | @No Content@ | operation successful, no body returned   |
    +------+--------------+------------------------------------------+
    | 404  | @Not Found@  | resource not found                       |
    +------+--------------+------------------------------------------+

Example doc page (Classic theme):

Example doc page (Classic theme)

Example doc page (Ocean theme):

Example doc page (Ocean theme)

See https://github.com/haskell/haddock/issues/530.


haddock-library/test/Documentation/Haddock/ParserSpec.hs
```diff
@@ -649,6 +649,50 @@ spec = do
             , " bar"
             ] `shouldParseTo` DocExamples [Example "foo" [" bar"]]
 
+    context "when parsing tables" $ do
+      let simpleTable :: Maybe [DocH () String] -> [[DocH () String]] -> Doc String
+          simpleTable header rows = DocTable (Table header rows)
+          table :: Maybe [DocH () a] -> [[DocH () a]] -> Doc a
+          table header rows = DocTable (Table header rows)
+
+      it "parses a table with an header" $
```
noformnocontent: a header

potomak: Fixed(up).

ranjitjhala:
👍

TikhonJelvis:
I've definitely used lists where I wanted tables in my docs. This is a real improvement: looks good in the source and looks great when rendered.

Worth noting that Emacs, at least, [supports editing this style of table](https://www.gnu.org/software/emacs/manual/html_node/emacs/Text-Based-Tables.html).

ivan-m:
At the risk of bikeshedding, will other forms of table be supported?

This seems to copy what [Pandoc calls](http://pandoc.org/MANUAL.html#tables) "grid tables", but my general preference is for what it refers to as "pipe tables" (in part because I've tried editing this style of table in Emacs as @TikhonJelvis points out but found it very slow, but pipe-tables with `orgtbl-mode` worked very nicely).

mpickering:
I don't think this is properly specified yet and the tests don't look very comprehensive.

potomak:
@mpickering can you give me more info about what is it missing to be properly specified?

potomak:
@TikhonJelvis actually the parser, the way it's been implemented, doesn't support table elements on multiple lines.

potomak:
@ivan-m I followed the proposal in https://github.com/haskell/haddock/issues/530, I don't have a strong opinion on the table format to support.

potomak:
@bgamari you seem quite active on this repo, what do you think about this change?
Do you think I should just close this PR?
Do you have any suggestion how can I make progress on this?

alexbiehl:
@potomak Don't close this PR. Tables in tables are a useful feature and we want it! 

Maybe @mpickering can clarify his concerns so they can be addressed.

alexbiehl:
@potomak I see there is no documentation at all. I think it would help to add a few sentences to `doc/markup.rst`. I am sure this will get us half-way there.

potomak:
@alexbiehl great! Thanks for the suggestion. I will update the documentation. I'm also working on a second version of this PR, it uses a different approach at table parsing. I will the PR soon.

haddock-library/src/Documentation/Haddock/Parser.hs
```diff
@@ -422,6 +423,55 @@ takeIndent = do
...
+  content <- parseTableContent
+  return $ DocTable (Table mHeader content)
+
+parseTableHeader :: Parser [DocH mod Identifier]
+parseTableHeader = parseTableRow <* parseTableHeaderDivider
+
+parseTableContent :: Parser [[DocH mod Identifier]]
+parseTableContent = many1 (parseTableRow <* parseTableRowDivider)
+
+parseTableRow :: Parser [DocH mod Identifier]
```
alexbiehl: I think you should provide some way to escape `|` in table cells. Otherwise `|` results in odd parser errors.

alexbiehl:
@potomak what is the status on this one? Do you need any help/advice on moving this forward?

hvr:
@potomak you mentioned a 2nd version of this PR; what's the status on that? and why did you think a different version was needed?

alexbiehl:
This has been addressed in another pr.

PR #598, opened 2017-04-01
Update MathJax URL

Rufflewind:
MathJax is shutting down their CDN:
https://www.mathjax.org/cdn-shutting-down/

They recommend migrating to cdnjs.


alexbiehl:
Thanks @Rufflewind!

PR #600, opened 2017-04-18
Travis: Build master with ghc-8.2

alexbiehl:
We were using `ghc-head` for the master branch.


alexbiehl:
@hvr If this looks good for you I will land this and fix the testsuite failures in subsequent PR (namely https://github.com/haskell/haddock/pull/602).

haddock.cabal
```diff
@@ -12,9 +12,10 @@ bug-reports:          https://github.com/haskell/haddock/issues
 copyright:            (c) Simon Marlow, David Waern
 category:             Documentation
 build-type:           Simple
-cabal-version:        >= 1.10
+cabal-version:        >= 2.0
```
hvr: We'd need to check whether GHC HEAD can still be bootstrapped with this (GHC's build uses only the  `haddock.cabal` file). Would `buildtools: haddock` work as well?

hvr: I've tweaked `cabal-version` in GHC HEAD, and ran a `./validate` and it works just fine. So false alarm.

alexbiehl:
Thanks @hvr!

Issue #599, opened 2017-04-02
Synopsis section not present in the haddocks being generated


psibi:
Not sure if this is the right repository or the hackage-server repository would be better suited for this.

Recent version of Hackage, is not showing synopsis section at all in the generated docs.

Example:

Synopsis is present here: https://hackage.haskell.org/package/persistent-2.2.4/docs/Database-Persist.html

But in the latest one it's absent: https://hackage.haskell.org/package/persistent-2.6.1/docs/Database-Persist.html

In fact it has been absent for quite some months now.

Fuuzetsu:
Pretty sure it was switched off on Hackage for some reason (though not 100%)

psibi:
@Fuuzetsu Any related thread / link for the reasoning behind it ?

Fuuzetsu:
Not that I can recall off the top of my head. I might be completely wrong, just saying it's worth asking someone from Hackage.

hvr:
I investigated a bit, and it turns out that this appears to be an **unintended** side-effect of 474a0c182d6ef9a7e93e277a2755d04d943059a0

alexbiehl:
Thanks @hvr. I'm on it.

alexbiehl:
With https://github.com/haskell/haddock/commit/4346ad26ac1346aa5e59991315f6969c844bea60 there is now a proper fix in master.

psibi:
@alexbiehl Thanks for the fix! Feel free to close the issue.

alexbiehl:
@psibi Please be aware that a fix is already present in the latest released haddock-2.17.4. So you can already use that until we release haddock-2.18.

PR #601, opened 2017-04-18
Remove anything related to obsolete frames mode

alexbiehl:
This is a proper fix for #599. Synopsis would only show if it is
contained in a `.no-frame` block which is a remain prior #514.

Also removes some obsolete javascript for searching a haddock page(?).

### DIFF
--- a/haddock-api/resources/html/Ocean.std-theme/ocean.css
+++ b/haddock-api/resources/html/Ocean.std-theme/ocean.css
@@ -327,10 +327,6 @@ div#style-menu-holder {
 }
 
 #synopsis {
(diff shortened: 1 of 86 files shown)


alexbiehl:
@ndmitchell I saw you were adding the search stuff about 10 years ago in https://github.com/haskell/haddock/commit/12b1c76fefc78b2a0fa54138a550c73802b1847d. Do you know anyone still relying on this code? Is this safe to remove without breaking 3rd party tooling?

ndmitchell:
@alexbiehl if it's not referenced internally it's probably dead code.

PR #602, opened 2017-04-25
Fix testsuite

alexbiehl:
This will become a series patches making the testsuite great again. This should land *after* https://github.com/haskell/haddock/pull/600.


alexbiehl:
Thanks @hvr. Your help is much appreciated!

PR #604, opened 2017-04-26
Update changelog to reflect news in HEAD

alexbiehl:


hvr:
oh dear... this a deadlock :-)

PR #609, opened 2017-04-30
Lazy decoding of InstalledIntefaces

alexbiehl:
Before we were decoding InstalledInterfaces from InterfaceFiles up
front. This patch enables lazy per-module deserialization of InstalledInterfaces.

Test failures expected, the haddock interfaces installed with ghc-8.2 do not comply to this new format yet.


alexbiehl:
Ah I think this wouldn't work well. The render phase uses information from all loaded `InstalledInterface` which defeats the purpose of this patch. Maybe I can try a finer grained approach.

PR #610, opened 2017-04-30
Lazily decode docMap and argMap

alexbiehl:
These are only used in case of a doc reexport so most of the time
decoding these is wasted work.


alexbiehl:
I tried a few packages with lots of modules for example Agda: It loads ~500 `InstalledInterface`s but and accesses ~30 their `docMap`s and `argMap`s (c.f. https://gist.github.com/alexbiehl/8f8d6637a39154c36925984139de065f). So I think this is worthwhile.

PR #614, opened 2017-05-08
Improve documenation of Haddock markup

ntc2:
This patch includes extensive improvements to the Markup section of the Haddock ReST docs. See the commit msgs on the individual commits for a summary of the changes.

I can't actually find the Sphinx/ReST manual for Haddock online anywhere! All I can find is this several year old, pre Sphinx version, at https://www.haskell.org/haddock/. Is the new manual available somewhere? If not, is there something I could do to help make that happen? In any case, it might be nice to bundle the Sphinx manual with the Haddock package on Hackage, so that it's easy for users to look up the manual for the specific version of Haddock they're using.

### DIFF
--- a/doc/markup.rst
+++ b/doc/markup.rst
@@ -164,6 +164,8 @@ Individual arguments to a function may be documented like this: ::
        -> Float    -- ^ The 'Float' argument
        -> IO ()    -- ^ The return value
 
+.. _module-description:
+
 The module description
 ----------------------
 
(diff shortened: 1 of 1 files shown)


alexbiehl:
Thansk @ntc2. Looks very good! 

@bgamari where are the docs hosted? Let's make sure to also put up the newest haddock docs at ghc-8.2 release as well.

alexbiehl:
Thanks @ntc2!

PR #615, opened 2017-05-12
Haddock: Fix broken lazy IO in prologue reading

bgamari:
We previously used withFile in conjunction with hGetContents. The list returned
by the latter wasn't completely forced by the time we left the withFile block,
meaning that we would try to read from a closed handle.


alexbiehl:
Thanks @bgamari!

PR #616, opened 2017-05-12
Cherry-pick changes from ghc-head

alexbiehl:
We previously used withFile in conjunction with hGetContents. The list returned
by the latter wasn't completely forced by the time we left the withFile block,
meaning that we would try to read from a closed handle.


bgamari:
Thanks for thinking of cherry-picking, @alexbiehl!

PR #621, opened 2017-05-18
Improve Syb code

duog:
Specialize.hs and Ast.hs are modified to have their Syb code not recurse into
Name or Id in HsSyn types.

Specialize.hs is refactored to have fewer calls to gmapT and gmapQ

There is still a lot of performance on the floor of Ast.hs. The RenamedSource is traversed many times, and lookupBySpan is very inefficient. everywhereBut and lookupBySpan dominate the runtime whenever --hyperlinked-source is passed. 

GHC test results with this patch rebased onto ghc-head:

=====> haddock.base(normal) 1 of 3 [0, 0, 0]
bytes allocated value is too low:
(If this is because you have improved GHC, please
update the test so that GHC doesn't regress again)
    Expected    haddock.base(normal) bytes allocated: 25592972912 +/-5%
    Lower bound haddock.base(normal) bytes allocated: 24313324266 
    Upper bound haddock.base(normal) bytes allocated: 26872621558 
    Actual      haddock.base(normal) bytes allocated: 19433190448 
    Deviation   haddock.base(normal) bytes allocated:       -24.1 %
*** unexpected stat test failure for haddock.base(normal)
=====> haddock.Cabal(normal) 2 of 3 [0, 0, 0]
bytes allocated value is too low:
(If this is because you have improved GHC, please
update the test so that GHC doesn't regress again)
    Expected    haddock.Cabal(normal) bytes allocated: 18865432648 +/-5%
    Lower bound haddock.Cabal(normal) bytes allocated: 17922161015 
    Upper bound haddock.Cabal(normal) bytes allocated: 19808704281 
    Actual      haddock.Cabal(normal) bytes allocated: 15790936128 
    Deviation   haddock.Cabal(normal) bytes allocated:       -16.3 %
*** unexpected stat test failure for haddock.Cabal(normal)
=====> haddock.compiler(normal) 3 of 3 [0, 0, 0]

### DIFF
--- a/haddock-api/src/Haddock/Backends/Hyperlinker/Ast.hs
+++ b/haddock-api/src/Haddock/Backends/Hyperlinker/Ast.hs
@@ -2,12 +2,12 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE TypeFamilies #-}
-
+{-# LANGUAGE TypeApplications #-}
 
 module Haddock.Backends.Hyperlinker.Ast (enrich) where
 
 
-import Haddock.Syb
+import qualified Haddock.Syb as Syb
 import Haddock.Backends.Hyperlinker.Types
 
 import qualified GHC
@@ -16,6 +16,9 @@ import Control.Applicative
 import Data.Data
 import Data.Maybe
 
+everythingInRenamedSource :: (Alternative f, Data x)
+  => (forall a. Data a => a -> f r) -> x -> f r
+everythingInRenamedSource f = Syb.everythingButType @GHC.Name (<|>) f
 
 -- | Add more detailed information to token stream using GHC API.
 enrich :: GHC.RenamedSource -> [Token] -> [RichToken]
@@ -53,7 +56,7 @@ enrichToken _ _ = Nothing
 -- | Obtain details map for variables ("normally" used identifiers).
 variables :: GHC.RenamedSource -> DetailsMap
 variables =
-    everything (<|>) (var `combine` rec)
+    everythingInRenamedSource (var `Syb.combine` rec)
   where
     var term = case cast term of
         (Just (GHC.L sspan (GHC.HsVar name))) ->
@@ -68,8 +71,7 @@ variables =
 
 -- | Obtain details map for types.
 types :: GHC.RenamedSource -> DetailsMap
-types =
-    everything (<|>) ty
+types = everythingInRenamedSource ty
   where
     ty term = case cast term of
         (Just (GHC.L sspan (GHC.HsTyVar _ name))) ->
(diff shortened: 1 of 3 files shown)


alexbiehl:
Thanks again @duog! 

Travis found some errors! It might be because of your specializer changes but I am not sure what exactly causes the issues.

haddock-api/src/Haddock/Backends/Hyperlinker/Ast.hs
```diff
@@ -93,7 +96,7 @@ binds =
         (Just (GHC.L sspan (GHC.VarPat name))) ->
             pure (sspan, RtkBind (GHC.unLoc name))
         (Just (GHC.L _ (GHC.ConPatIn (GHC.L sspan name) recs))) ->
-            [(sspan, RtkVar name)] ++ everything (<|>) rec recs
+            [(sspan, RtkVar name)] ++ Syb.everything (<|>) rec recs
```
alexbiehl: Why do we want `everything` here? Do we want to look into `Name`s here?

duog: No, that's a mistake. Not sure how that slipped in.

duog:
I'm not sure the cause either. I will likely not have any time to look at this until Monday next week, but I'll fix these problems then.

alexbiehl:
Ha, this will be an unsetteling weekend for you. Thinking what went wrong all the time! I am curious too and will take alook later today.

duog:
I have had a little look today.
It looks like I misunderstood everythingBut, and is_not_t should be is_t. Boolean blindness strikes again.
I'm not sure how the ghc docs weren't affected. Either they don't exercise this code path or I erred again there somehow.
It might be back to the drawing board.

alexbiehl:
@duog Any new ideas yet?

duog:
Yes, I have resolved this.
I've been occupied, I'll try and ammend this PR today.

duog:
Rebased the branch and updated summary

alexbiehl:
Nice! Thanks @duog, I will take a look later. Tests seem fine at least which is half way already.

haddock-api/src/Haddock/Syb.hs
```diff
@@ -1,25 +1,53 @@
...
     , combine
     ) where
 
 
 import Data.Data
 import Control.Applicative
+import Data.Maybe
+import Data.Foldable
 
+-- | Returns true if a /= t.
```
alexbiehl: It should be a == t, right?

haddock-api/src/Haddock/Interface/Specialize.hs
```diff
@@ -1,8 +1,9 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE Rank2Types #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE RecordWildCards #-}
-
+{-# LANGUAGE TypeApplications #-}
```
alexbiehl: Duplicate pragma

duog:
Great! Fixed your nits.

PR #627, opened 2017-05-31
Haddock support for bundled pattern synonyms

christiaanb:
Given the following:

```haskell
{-# LANGUAGE DataKinds, GADTs, KindSignatures, PatternSynonyms, TypeOperators,
             ViewPatterns #-}
module Vector (Vec(Nil,(:>)), RTree (LR,BR) ) where

import GHC.TypeLits
import Prelude hiding (head, tail)
import Unsafe.Coerce

-- | Fixed size vectors.
--
-- * Lists with their length encoded in their type
-- * 'Vec'tor elements have an __ASCENDING__ subscript starting from 0 and
--   ending at @'length' - 1@.
data Vec :: Nat -> * -> * where
  Nil  :: Vec 0 a
  Cons :: a -> Vec n a -> Vec (n + 1) a

infixr 5 `Cons`

-- | Add an element to the head of a vector.
--
-- >>> 3:>4:>5:>Nil
-- <3,4,5>
-- >>> let x = 3:>4:>5:>Nil
-- >>> :t x
-- x :: Num a => Vec 3 a
--
-- Can be used as a pattern:
--
-- >>> let f (x :> y :> _) = x + y
-- >>> :t f
-- f :: Num a => Vec ((n + 1) + 1) a -> a
-- >>> f (3:>4:>5:>6:>7:>Nil)
-- 7
--
-- Also in conjunctions with (':<'):
--
-- >>> let g (a :> b :> (_ :< y :< x)) = a + b +  x + y
-- >>> :t g
-- g :: Num a => Vec ((((n + 1) + 1) + 1) + 1) a -> a
-- >>> g (1:>2:>3:>4:>5:>Nil)
-- 12
pattern (:>) :: a -> Vec n a -> Vec (n + 1) a
pattern (:>) x xs <- ((\ys -> (head ys,tail ys)) -> (x,xs))
  where
    (:>) x xs = Cons x xs

infixr 5 :>

head :: Vec (n + 1) a -> a
head (x `Cons` _) = x

tail :: Vec (n + 1) a -> Vec n a
tail (_ `Cons` xs) = unsafeCoerce xs

-- | Perfect depth binary tree.
--
-- * Only has elements at the leaf of the tree
-- * A tree of depth /d/ has /2^d/ elements.
data RTree :: Nat -> * -> * where
  LR_ :: a -> RTree 0 a
  BR_ :: RTree d a -> RTree d a -> RTree (d+1) a

textract :: RTree 0 a -> a
textract (LR_ x) = x
{-# NOINLINE textract #-}

tsplit :: RTree (d+1) a -> (RTree d a,RTree d a)
tsplit (BR_ l r) = (unsafeCoerce l, unsafeCoerce r)
{-# NOINLINE tsplit #-}

-- | Leaf of a perfect depth tree
--
-- >>> LR 1
-- 1
-- >>> let x = LR 1
-- >>> :t x
-- x :: Num a => RTree 0 a
--
-- Can be used as a pattern:
--
-- >>> let f (LR a) (LR b) = a + b
-- >>> :t f
-- f :: Num a => RTree 0 a -> RTree 0 a -> a
-- >>> f (LR 1) (LR 2)
-- 3
pattern LR :: a -> RTree 0 a
pattern LR x <- (textract -> x)
  where
    LR x = LR_ x

-- | Branch of a perfect depth tree
--
-- >>> BR (LR 1) (LR 2)
-- <1,2>
-- >>> let x = BR (LR 1) (LR 2)
-- >>> :t x
-- x :: Num a => RTree 1 a
--
-- Case be used a pattern:
--
-- >>> let f (BR (LR a) (LR b)) = LR (a + b)
-- >>> :t f
-- f :: Num a => RTree 1 a -> RTree 0 a
-- >>> f (BR (LR 1) (LR 2))
-- 3
pattern BR :: RTree d a -> RTree d a -> RTree (d+1) a
pattern BR l r <- ((\t -> (tsplit t)) -> (l,r))
  where
    BR l r = BR_ l r
```
The following is rendered:

screenshot

This patch supersedes #494

### DIFF
--- a/haddock-api/src/Haddock/Interface/Create.hs
+++ b/haddock-api/src/Haddock/Interface/Create.hs
@@ -1026,8 +1026,9 @@ mkVisibleNames (_, _, _, _, instMap) exports opts
   | otherwise = let ns = concatMap exportName exports
                 in seqList ns `seq` ns
   where
-    exportName e@ExportDecl {} = name ++ subs
-      where subs = map fst (expItemSubDocs e)
+    exportName e@ExportDecl {} = name ++ subs ++ patsyns
+      where subs    = map fst (expItemSubDocs e)
+            patsyns = concatMap (getMainDeclBinder . fst) (expItemPats e)
             name = case unLoc $ expItemDecl e of
               InstD d -> maybeToList $ M.lookup (getInstLoc d) instMap
               decl    -> getMainDeclBinder decl
--- a/haddock-api/src/Haddock/Interface/Create.hs
+++ b/haddock-api/src/Haddock/Interface/Create.hs
@@ -552,10 +552,12 @@ mkExportItems
     Nothing -> fullModuleContents dflags warnings gre maps fixMap splices decls
     Just exports -> liftM concat $ mapM lookupExport exports
   where
-    lookupExport (IEVar (L _ x))         = declWith $ ieWrappedName x
-    lookupExport (IEThingAbs (L _ t))    = declWith $ ieWrappedName t
-    lookupExport (IEThingAll (L _ t))    = declWith $ ieWrappedName t
-    lookupExport (IEThingWith (L _ t) _ _ _) = declWith $ ieWrappedName t
+    lookupExport (IEVar (L _ x))         = declWith [] $ ieWrappedName x
+    lookupExport (IEThingAbs (L _ t))    = declWith [] $ ieWrappedName t
+    lookupExport (IEThingAll (L _ t))    = declWith [] $ ieWrappedName t
+    lookupExport (IEThingWith (L _ t) _ cs _) = do
+      pats <- catMaybes <$> mapM (findBundledPatterns . ieWrappedName . unLoc) cs
+      declWith pats $ ieWrappedName t
     lookupExport (IEModuleContents (L _ m)) =
       -- TODO: We could get more accurate reporting here if IEModuleContents
       -- also recorded the actual names that are exported here.  We CAN
@@ -574,8 +576,8 @@ mkExportItems
         Nothing -> []
         Just doc -> return . ExportDoc $ processDocStringParas dflags gre doc
 
-    declWith :: Name -> ErrMsgGhc [ ExportItem Name ]
-    declWith t = do
+    declWith :: [(HsDecl Name, DocForDecl Name)] -> Name -> ErrMsgGhc [ ExportItem Name ]
+    declWith pats t = do
       r <- findDecl t
       case r of
         ([L l (ValD _)], (doc, _)) -> do
@@ -612,15 +614,15 @@ mkExportItems
                     -- fromJust is safe since we already checked in guards
                     -- that 't' is a name declared in this declaration.
                     let newDecl = L loc . SigD . fromJust $ filterSigNames (== t) sig
-                    in return [ mkExportDecl t newDecl docs_ ]
+                    in return [ mkExportDecl t newDecl pats docs_ ]
 
                   L loc (TyClD cl@ClassDecl{}) -> do
                     mdef <- liftGhcToErrMsgGhc $ minimalDef t
                     let sig = maybeToList $ fmap (noLoc . MinimalSig NoSourceText . noLoc . fmap noLoc) mdef
                     return [ mkExportDecl t
-                      (L loc $ TyClD cl { tcdSigs = sig ++ tcdSigs cl }) docs_ ]
+                      (L loc $ TyClD cl { tcdSigs = sig ++ tcdSigs cl }) pats docs_ ]
 
-                  _ -> return [ mkExportDecl t decl docs_ ]
+                  _ -> return [ mkExportDecl t decl pats docs_ ]
 
         -- Declaration from another package
         ([], _) -> do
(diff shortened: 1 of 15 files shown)


bgamari:
This looks quite good to me; much better than the previous approach.

bgamari:
Indeed, I'd be happy to take this for 8.2.

christiaanb:
@alexbiehl I'm on it

alexbiehl:
Also I think it would be good if you add your example as a test case to catch future regressions.

christiaanb:
screenshot2

Fixities are now included, and bundled pattern synonyms are included in the synopsis.

I would like to add a testcase... but I don't know how to add a reference. How do I create the proper HTML files? (running `haddock --html --pretty-html` gives slightly different output than the testbench it seems)

haddock-api/src/Haddock/Types.hs
```diff
@@ -217,6 +217,9 @@ data ExportItem name
         -- | A declaration.
         expItemDecl :: !(LHsDecl name)
 
+        -- | Bundled patterns for a data type declaration
+      , expItemPats :: ![(HsDecl name,DocForDecl name,[(name, Fixity)])]
```
alexbiehl: I think we should just reuse the already present `expItemFixities` field. Also I presume with your current design fixities won't be saved to interface files (`toInstalledInterface` etc.).

alexbiehl:
@christiaanb I think the way to add a test is this: 

- Add your test under `/html-test/src`
- Build the `html-test` test suite 
- Run `html-test` test suite with `-a` param (see https://github.com/haskell/haddock/blob/master/haddock-test/src/Test/Haddock/Config.hs#L135)

This should build the source file, it will complain that there is no `reference` to compare to, but with `-a` it will accept and copy over the generated html.

christiaanb:
@alexbiehl I've changed the way `Fixitities` are stored in the way you recommended, and I've added a test case. When you merge this PR could you do a `Squash&Merge` so that it becomes a single commit, making it easier to cherry-pick into other branches?

haddock-api/src/Haddock/Interface/Create.hs
```diff
@@ -638,21 +638,22 @@ mkExportItems
...
                 Just iface ->
-                   return [ mkExportDecl t decl pats (lookupDocs t warnings (instDocMap iface) (instArgMap iface) (instSubMap iface)) ]
+                   return [ mkExportDecl t decl pats patFixities (lookupDocs t warnings (instDocMap iface) (instArgMap iface) (instSubMap iface)) ]
 
         _ -> return []
 
 
-    mkExportDecl :: Name -> LHsDecl Name -> [(HsDecl Name, DocForDecl Name, [(Name,Fixity)])]
+    mkExportDecl :: Name -> LHsDecl Name -> [(HsDecl Name, DocForDecl Name)]
+                 -> [(Name,Fixity)]
```
alexbiehl: I think we don't need to thread the fixities; instead grab them from `fixMap`. You have the names in `pats`, just look them up in `mkExportDecl`.

alexbiehl:
I think we are almost there with the fixities. 

But I think there is more to do: Take a look at `mkMaps` it scans all declarations and creates some maps with certain information. Specifically there is the `subs` map which contains all "subordinates" of a declaration. I think we should make sure to make bundled pattern synonyms subordinates of their bundled data type. Pattern synonyms wouldn't show up reexported in data definitions without this.

haddock-api/src/Haddock/Interface/Create.hs
```diff
@@ -636,20 +638,22 @@ mkExportItems
...
                 Just iface ->
-                   return [ mkExportDecl t decl (lookupDocs t warnings (instDocMap iface) (instArgMap iface) (instSubMap iface)) ]
+                   return [ mkExportDecl t decl pats patFixities (lookupDocs t warnings (instDocMap iface) (instArgMap iface) (instSubMap iface)) ]
 
         _ -> return []
 
 
-    mkExportDecl :: Name -> LHsDecl Name -> (DocForDecl Name, [(Name, DocForDecl Name)]) -> ExportItem Name
-    mkExportDecl name decl (doc, subs) = decl'
+    mkExportDecl :: Name -> LHsDecl Name -> [(HsDecl Name, DocForDecl Name)]
```
alexbiehl: I think we don't need to thread the fixities; instead grab them from fixMap. You have the names in pats, just look them up in mkExportDecl.

christiaanb:
@alexbiehl I'm stuck with regards to `mkMaps` part. Bundled pattern synonyms are a syntactic sugar import/export thing, they are not part of the data declaration. So given an export item `IEThingWith` I can find the bundled pattern synonyms. But any other export items I'm stuck, as I there is no way to find out if a pattern in bundled with a data-constructor or not...

alexbiehl:
@christiaanb Thanks for your effort; fixities look good now!

For the other thing: 
It is a bit difficult as the maps are created before we even traverse the import/export things. But we have the `TypecheckedModule` available; we can grab the `Avails` via `minf_exports`.  When creating subordinates for a data type check its `AvailInfo` and see if it also exports pattern synonyms. Could this work? What do you think?




I have a little test case exhibiting the behaviour:
```
module Vec(Vec(..)) where

import Vector (Vec(..))
```
Running with `haddock -html Vec.hs Vector.hs`. As expected it doesn't list the bundled pattern synonym `(:>)` in `Vec`.

christiaanb:
I haven't been able to include fixity information for re-exported bundled pattern synonyms. My problem now is, is that I've already spend more time on this issue than initially planned. I have quite a busy schedule the next few weeks, so I won't be able to work on this for some time.

I see a couple of options now:
1. Someone else takes over the work in time for the GHC 8.2.1 release
2. This gets merged as-is, and we open an issue that Fixity information isn't included in re-exported bundled pattern synonyms.
3. We don't merge and we'll see when I have time again; most likely not before the GHC 8.2.1 release.

Unless option 1 happens, I would really like to have option 2 happen, given that IMHO, fixity information for re-exported bundled pattern synonyms is such a minor details; compared to have no haddock support for bundled pattern synonyms at all.

alexbiehl:
@christiaanb Thank you for your effort. I will finish this one for you. I want this in ghc-8.2! 

Thanks again for your initial work and time!

christiaanb:
@alexbiehl Thanks for finishing it! And my apologies for dropping this in your lap like this. The status of the patch that is there now was straightforward to get to (and or easy to copy/adapt from GHC code). I'm sure the fixity information for re-exported bundled pattern synonyms is out there somewhere; but given my complete unfamiliarity with the codebase, all my attempts to get it to the right place were basically trial&error with no indication of any form of progress.

alexbiehl:
Fixities for reexports get attached much later in the pipeline in `attachToExportItem`. It looks for them in the subordinates. So as I said in https://github.com/haskell/haddock/pull/627#issuecomment-305248747 we need to put the pattern synonyms into the subordinates in `mkMaps`.

alexbiehl:
@christiaanb @bgamari It is working now! Please have a look and suggest improvements.

alexbiehl:
I forgot about the possibility of bundled patsyns from InstalledInterfaces. This needed a change in the InterfaceFile format. Some tests will break and will stay broken until @bgamari bumps the haddock submodule in ghc-8.2 and @hvr releases a new build to his ppa.

christiaanb:
Thanks for finishing this up @alexbiehl, really happy that it's in time for the ghc821 release!

alexbiehl:
I'm glad too. There were quite some details to figure out but I think this should be reasonably solid now. Waiting now for @bgamari to bump haddock submodule in ghc-8.2. Then the tests will become green again.

Issue #623, opened 2017-05-23
Use DynFlags that suppress warnings and throw fewer errors.


duog:
I have two requests.
Firstly, I would like to disable pattern match warnings (-fwarn-incomplete-patterns et al) by default in haddock. Currently, -fno-code does not desugar code, so haddock will not produce pattern match warnings unless a package uses template haskell.
I have a ghc patch in development, https://phabricator.haskell.org/D3542, which enables desugaring for -fno-code, for the purpose of  generating more warnings and errors. However, this causes significant regressions in the haddock performance tests, due to some pathalogical pattern matches. Disabling pattern match warnings would resolve these.

Secondly, I would suggest that disabling all warnings would be a better default for haddock. I would also suggest that enabling some other flags, e.g -fdefer-type-errors would allow haddock to produce output for more packages, with no downside that I can see.

If you agree that enabling -fdefer-type-errors is a good idea, what other flags should be set? Perhaps -fdefer-typed-holes, -XPartialTypeSignatures.

I'll be happy to submit a PR if you're interested in some or all of these changes.

alexbiehl:
There are two points in this proposal: 

1. ghc-8.2 (?) will be released with `D3542` and I wouldn't want to take those regressions. Also I think `-fwarn-incomplete-patterns` doesn't do any harm for haddock generation I suppose. So lets do this.

2. As for all the other flags we should carefully sort them out, one by one looking at possible interaction with generated haddocks. For example I am not sure I want `-fdefer-type-errors` and let invalid type signatures sneak into my haddock database. But I am open for discussion.

duog:
ghc-8.2 will not have D3542, if it lands it will be in ghc-8.4.

As for 2., there are actually two sub-points. Disabling warnings, and enabling flags like -fdefer-type-errors to accept some invalid modules.

Fuuzetsu:
I think it's im best interest to stick close to flags you get during normal build. This way you don't have bad surprise when you find out that code you thought builds fine because haddock was able to get ghc to build and typecheck it is actually broken.

At the same time it seems like a abuse of functionality to me (compilation is an unwanted side effect) so I'd be OK with adding flags which speed things up as long as we document that. However that's as far as I'd go: if there is no performance benefit to a flag or the flag possibly results in failed documentation build where vanilla build works, I'm against it.

[...]

PR #628, opened 2017-06-01
Disable pattern match warnings

duog:
Merging to ghc-head because these warnings can't actually fire in ghc-8.2 (unless there's template haskell), because -fno-code doesn't do desugaring. 
As discussed in #623 
This disables the pattern match checker which can be very expensive in some
cases.

The disabled warnings include:
* Opt_WarnIncompletePatterns
* Opt_WarnIncompleteUniPatterns
* Opt_WarnIncompletePatternsRecUpd
* Opt_WarnOverlappingPatterns

### DIFF
--- a/haddock-api/src/Haddock.hs
+++ b/haddock-api/src/Haddock.hs
@@ -41,7 +41,7 @@ import Haddock.Utils
 
 import Control.Monad hiding (forM_)
 import Control.Applicative
-import Data.Foldable (forM_)
+import Data.Foldable (forM_, foldl')
 import Data.List (isPrefixOf)
 import Control.Exception
 import Data.Maybe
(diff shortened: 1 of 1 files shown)


haddock-api/src/Haddock.hs
```diff
@@ -160,8 +162,9 @@ haddockWithGhc ghc args = handleTopExceptions $ do
     forM_ (warnings args) $ \warning -> do
       hPutStrLn stderr warning
 
+  -- We disable pattern match warnings because than can be very expensive to check
   ghc flags' $ do
-
+    modifySession unsetPatternMatchWarnings
```
alexbiehl: We already have `withGhc'` at https://github.com/haskell/haddock/blob/master/haddock-api/src/Haddock.hs#L392 would it make sense to place this there? Having all the flag manipulation stuff in one place seems like a good idea.

duog: Yes that's much better.

alexbiehl:
Thanks @duog!

PR #636, opened 2017-06-11
Use new function getNameToInstancesIndex instead of tcRnGetInfo

duog:
Note that this patch is dependent on [D3636](https://phabricator.haskell.org/D3636).

There is some performance improvement.

GHC compiler:
| version | bytes allocated | cpu_seconds
---------------------------------
| before  | 56057108648     | 41.0
| after   | 51592019560     | 35.1

base:
| version | bytes allocated | cpu_seconds
---------------------------------
| before  | 25174011784     | 14.6
| after   | 23712637272     | 13.1

Cabal:

| version | bytes allocated | cpu_seconds
---------------------------------
| before  | 18754966920     | 12.6
| after   | 18198208864     | 11.6

### DIFF
--- a/haddock-api/src/Haddock/Interface/AttachInstances.hs
+++ b/haddock-api/src/Haddock/Interface/AttachInstances.hs
@@ -23,9 +23,10 @@ import Control.Arrow hiding ((<+>))
 import Data.List
 import Data.Ord (comparing)
 import Data.Function (on)
-import Data.Maybe ( maybeToList, mapMaybe )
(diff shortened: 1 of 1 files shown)


alexbiehl:
Thank you @duog. Before I merge, I need to get the ghc-head branch work on Travis. This seems like a potential case where tests might fail.

duog:
No problem, I'll ping you here once that diff is merged.

duog:
@alexbiehl The diff is merged.

PR #642, opened 2017-06-23
Lookup fixities for reexports without subordinates

### LINKED ISSUE #641: Haddock drops fixity declarations for re-exported class methods
Here's two modules:

```haskell
module Foo (C(..)) where

infixl 9 `a`

class C x where
  a :: x
```
```haskell
module Bar (a) where

import Foo
```

To reproduce this issue, run `haddock Bar.hs --html -o foobar`, and open up `foobar/Foo.html` and `foobar/Bar.html`. You'll notice that `Foo.html` looks fine:

screen shot 2017-06-22 at 4 18 29 pm

But `Bar.html` does not: it doesn't report the fixity of `a`!

screen shot 2017-06-22 at 4 17 48 pm

Moreover, this only happens when you re-export `a` by itself. If you bundle `a` with `C` when re-exporting it:

```haskell
module Bar (C(..)) where

import Foo
```

Then this issue does not occur:

screen shot 2017-06-22 at 4 20 08 pm

This is the version of `haddock` I'm using:

```
Haddock version 2.18.0, (c) Simon Marlow 2006
```

Although the issue also occurs with earlier `haddock`s as well.

alexbiehl:
This resolves #641: Reexported `a` in `Bar` doesn't have any subordinates (it is neither a class, type family nor a data declaration) so fixity lookup stops right there.

Solution: Return the empty list for declarations without subordinates now.

### DIFF
--- a/haddock-api/src/Haddock/Interface/AttachInstances.hs
+++ b/haddock-api/src/Haddock/Interface/AttachInstances.hs
@@ -18,6 +18,7 @@ import Haddock.Types
 import Haddock.Convert
 import Haddock.GhcUtils
 
+import Control.Applicative
 import Control.Arrow hiding ((<+>))
 import Data.List
 import Data.Ord (comparing)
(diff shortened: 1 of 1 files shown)


alexbiehl:
So we agree that reexported declarations which do not have subordinates (for example top-level functions) shouldn't have gotten fixities reexported according to the current logic. I wondered why for example `Prelude.($)` which is obviously reexported from `GHC.Base` has fixities attached (c.f. http://hackage.haskell.org/package/base-4.9.1.0/docs/Prelude.html#v:-36-). 

The reason is this: In `mkMaps` we lookup all the subordinates of top-level declarations, of course top-level functions don't have subordinates so for them the resulting list is empty. In https://github.com/haskell/haddock/pull/644 I established the invariant that there won't be any empty lists in the subordinate map. Without the patch from #642 top-level functions now started to fail reexporting their fixities.

To sum things up: With #642 it works correctly and with #644 we don't write unecessary stuff to interface files. This is a win!

PR #648, opened 2017-06-29
Make per-argument docs for class methods work again

### LINKED ISSUE #647: Regression: per-argument docs in type class methods are not rendered anymore
See: https://stackoverflow.com/questions/44820920/is-there-a-way-to-make-haddock-render-per-argument-docs-for-type-class-methods

2.16.1, works: https://hackage.haskell.org/package/megaparsec-5.0.0/docs/Text-Megaparsec-Prim.html (see the `failure` method of `MonadParsec`).

2.17.3, does not work: https://hackage.haskell.org/package/megaparsec-5.3.1/docs/Text-Megaparsec-Prim.html#v:failure

The docs have been there for all time.

That's quite unfortunate, people got confused without the docs.

alexbiehl:
This fixes #647. I am not sure why this stopped working somewhere between 2.16 and 2.17. git blame suggests the function didn't cover `ClassOpSig` for ages.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,8 +1,10 @@
 ## Changes in version 2.18.0
 
- * Support user defined signatures on pattern synonyms
-
  * Synopsis is working again (#599)
+
+ * Per-argument documentation on class methods working again (#647)
+
+ * Support user defined signatures on pattern synonyms
 
  * Support for bundled pattern synonyms (#494, #551, #626)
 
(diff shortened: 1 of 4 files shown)


Fuuzetsu:
Likely GHC version changed and started makinf use of ClassOpSig

alexbiehl:
@Fuuzetsu you are right: https://github.com/ghc/ghc/commit/1e041b7382b6aa329e4ad9625439f811e0f27232 introduced `ClassOpSig`.

PR #660, opened 2017-08-05
Fix renaming after instance signature specializing

### LINKED ISSUE #613: Haddock generated type signature of instance methods did not do α-conversion
Here's how to replicate (view) the bug:
1. Open the url [https://hackage.haskell.org/package/base-4.9.1.0/docs/Prelude.html#t:Applicative](https://hackage.haskell.org/package/base-4.9.1.0/docs/Prelude.html#t:Applicative)
2. Scroll down, until you find out the docs item of the instance Applicative ((->) a)
3. Extend the collapser

Here's the screenshot.
image
This is how the bug looks like.
You may notice that this docs item is not correct, and the production of the correct docs involves an α-conversion.

The correct docs item should be like this:
image
(The red mark stands for the prime character)

themoritz:
Fixes #613

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
  * to be released
 
+ * Fix renaming of type variables after specializing instance method signatures (#613)
+
  * Move markup related data types to haddock-library
 
 ## Changes in version 2.18.1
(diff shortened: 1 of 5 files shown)


alexbiehl:
Thank you!

PR #663, opened 2017-08-05
Fix: Generate pattern signatures for constructors exported as patterns

timjb:
This pull request also fixes a bug in the representation of pattern synonyms: Since pattern synonyms have two different contexts, we must render each context even if it is empty.


haddock-api/src/Haddock/Interface/Create.hs
```diff
@@ -1003,6 +1005,36 @@ extractDecl name decl
           _ -> error "internal: extractDecl (ClsInstD)"
       _ -> error "internal: extractDecl"
 
+extractPatternSyn :: Name -> Name -> [LHsType Name] -> [LConDecl Name] -> LSig Name
+extractPatternSyn nm t tvs cons =
+  case filter matches cons of
+    [] -> error "extractRecSel: constructor pattern not found"
```
alexbiehl: Should be extractPatternSyn.

timjb: Good catch!

PR #668, opened 2017-08-13
Add simple framework for running parser fixtures

phadej:
I'd like to resurrect table PR, and this will help writing tests


alexbiehl:
Nice! Thank you! 

Have you seen `haddock-test` by the way? It allows you write similar tests and offers some convenience. Would it be worth using that for the fixtures?

phadej:
@alexbiehl, I'd rather not *yet* make a dependency loop `haddock-library:lib -> haddock-test -> haddock-library:test:fixtures`, as `cabal new-build` cannot yet work with such configurations.

phadej:
@alexbiehl I plan to complete/release `structural-diff` (see http://oleg.fi/gists/posts/2016-07-24-sop-diff.html) library, I still need it for a while for Cabal parser hacking, and seems it will be useful for haddock, and others parsers to come. Would it be ok to include it, to make better-than-line based difference outputs? The dependency footprint won't escape Haskell Platform, except for generics-sop, ansi-terminal and (prettyprinter or pretty-compact).

alexbiehl:
@phadej That would be ok for me as long as GHC is still able to build haddock of course.

phadej:
I remade this to use https://github.com/phadej/tree-diff. it's yet unreleased as I'm polishing the rough edges.

For my own need I included quite a lot of instances there, I'll introduce a manual flag (for GHC) to disable most of them (i.e. the ones not needed for actual implementation).

phadej:
FWIW, it would be much more concise if http://hackage.haskell.org/package/tasty-golden variant existed for `hspec`.

phadej:
I asked @bgamari, according to him GHC build process doesn't run haddock test-suite because there's already a lot of external dependencies.

So I'd rather just leave `tree-diff` dependencies as they are (i.e. batteries included).

ping @alexbiehl, is it ok?

alexbiehl:
Yeah, that works for me!

alexbiehl:
Is this ready to be merged? 

I retriggerd the build now that tree-diff is released Travis is successful hopefully.

phadej:
I need to correct golden-files, as I made string output more faithful in the released version. It doesn't omit newline characters now.

Also I want to check if golden test support can be easily added to the hspec, so we won't have that much custom code in haddocks codebase

alexbiehl:
ping @phadej. How is your effort going on bringen golden tests to hspec? FWIW I think the code is good and would be willing to merge as long as you make the tests green.

alexbiehl:
ping @phadej

phadej:
I'll take a look why this fails on travis today. Let's not wait for hspec-golden to happen

alexbiehl:
Thanks!

alexbiehl:
@phadej I fixed the issues. Could you have a last look before I merge?

phadej:
@alexbiehl looks great, thanks! I completely forgot about this one.

PR #674, opened 2017-08-27
deleted the nowrap for interface method sigs

paolino:
with nowrap the interfaces method sigs would expand at libitum


alexbiehl:
Thanks @paolino!

PR #681, opened 2017-09-08
Overhaul Haddock's rendering of kind signatures

### LINKED ISSUE #189: Incorrect or at least misleading output with PolyKinds + TypeOperators
**Original reporter**: _illissius@<hidden>_

The code

``` haskell
{-# LANGUAGE PolyKinds, ConstraintKinds, TypeOperators, MultiParamTypeClasses, FlexibleInstances, UndecidableInstances #-}
class    (c a, d a) => (c :&: d) a
instance (c a, d a) => (c :&: d) a
```

produces the following documentation for the instance:

``` haskell
Instances
(c a, d a) => (k :&: c) d a
```

I don't know where it should be, but that k is clearly not in the right place.

### LINKED ISSUE #491: Haddock mixes Type variables and Kind Variables
In a variety of places, Haddock seems to be showing that kind variables as though they were type variables, resulting in some very misleading signatures.

Take a look at the signature for the `(~~)` class as seen in the doucmentation for `Data.Type.Equality` in [current ghc master](http://downloads.haskell.org/~ghc/master/libraries/html/base/Data-Type-Equality.html#t:-126--126-): 

```
class (~#) j k a b => (j ~~ k) a b 
```

That is severely messed up, and there is no way to view things where that is even approximately correct. The correct signature would be:

```
class (~#) a b => (a :: j) ~~ (b :: k)
```

Technically that is still not a valid signature because (~#) is technically not a constraint, but it really is what we want to display, because it captures the correct idea, per `Note [The equality types story]` of [TysPrim.hs](https://ghc.haskell.org/trac/ghc/browser/ghc/compiler/prelude/TysPrim.hs).

We also see kinds and kind variables posing as type variables in instance signatures. See for example all the instances for [Data.Functor.Sum](http://downloads.haskell.org/~ghc/master/libraries/html/base/Data-Functor-Sum.html). Every single one of them shows either a kind of `*` or a kind variable as though they are some type parameter. I'm not experienced enough working with PolyKinds to have any idea which type parameter(s) the kind or kind variable actually applies to  (without checking in GHCi). Somebody who has never read anything about PolyKinds is likely to be even more confused.

I'd suggest that kind parameters not be shown by default for a kind of `*`.  GHCi itself when using `:browse` or `:info` defaults to showing kind variables in type signatures (with an exception around RuntimeRep), but never shows the application of the Kind value \* to a kind variable in instance signatures (It's unclear to me how it handles application of other specific kind values). Where we do show kind variables it should be using the kind signature notation.  

I understand that with GHC 8.0 technically kind and types are the same thing, and thus a kind variable is a type variable, but there is still a distinction made. Type variables in a type constructor correspond to Binders in the Kind signature. The ones with a named binder with `Invisible` or `Qualified` are do not show up as parameters in the type constructor, and are therefore equivalent to kind variables in the above discussion. Those that have an Anonymous binder, or a Namerd Binder with `Visible` do appear as type parameters, and should be shown like any other type parameter. 

See the `Note [TyBinders and VisibilityFlags]` in [TyCoRep.hs](https://github.com/ghc/ghc/blob/master/compiler/types/) for more information about this.

### LINKED ISSUE #544: Haddock renders runtime rep arguments of unboxed tuples
An example can be seen [here](http://docs.futurice.com/haskell-mega-repo/primitive-0.6.1.0/Control-Monad-Primitive.html) if you expand the information for the `PrimMonad IO` instance:

runtime-reps

Notice the superfluous `VoidRep` and `PtrRepLifted` arguments.

We probably need to call [`dropRuntimeRepArgs`](http://git.haskell.org/ghc.git/blob/e98edbd1b8947dcb4a47d20fc854c0a9a6954bea:/compiler/types/TyCoRep.hs#l753) somewhere, but I'm not sure where...

RyanGlScott:
To put it nicely, the way Haddock renders kind signatures is less than ideal. This is best illustrated with examples, so here's some Haddocks taken from `GHC.Generics`:

haddock1

Ugh. There's multiple unsavory things happening here:

1. Those instances aren't using legal Haskell syntax. Haddock is mistakenly believing that kinds are visible arguments, leading to total garbage like `((:.:) * k f g)`.
2. Moreover, many of those kinds are quite redundant. There's no need to say, for instance, `Functor ((:.:) * * f g)`, as the kind of `Functor` forces the kinds to be `*` anyways.

Having recently tackled a similar problem in [Template Haskell](https://phabricator.haskell.org/rGHCad7b945257ea262e3f6f46daa4ff3e451aeeae0b), I realized that the solution would actually be a great fit for Haddock as well. Consequently, much of this code in this patch is cargo-culted from `TcSplice` in GHC, which handles turning GHC `Type`s into Template Haskell types (Haddock is doing something similar here, turning GHC `Type`s into `LHsType GhcRn`s.)

After my changes, here's what that part of `GHC.Generics` now looks like:

haddock2

Much better!

Off the top of my head, this resolves #189, resolves #491, and resolves #544.

### DIFF
--- a/haddock-api/src/Haddock/Backends/LaTeX.hs
+++ b/haddock-api/src/Haddock/Backends/LaTeX.hs
@@ -553,7 +553,7 @@ ppInstHead unicode (InstHead {..}) = case ihdInstType of
     TypeInst rhs -> keyword "type" <+> typ <+> tibody rhs
     DataInst _ -> error "data instances not supported by --latex yet"
   where
-    typ = ppAppNameTypes ihdClsName ihdKinds ihdTypes unicode
+    typ = ppAppNameTypes ihdClsName ihdTypes unicode
     tibody = maybe empty (\t -> equals <+> ppType unicode t)
 
 lookupAnySubdoc :: (Eq name1) =>
@@ -831,27 +831,27 @@ ppDataHeader _ _ = error "ppDataHeader: illegal argument"
 --------------------------------------------------------------------------------
 
 
--- | Print an application of a DocName and two lists of HsTypes (kinds, types)
-ppAppNameTypes :: DocName -> [HsType DocNameI] -> [HsType DocNameI] -> Bool -> LaTeX
-ppAppNameTypes n _ks ts unicode = ppTypeApp n [] ts ppDocName (ppParendType unicode)
+-- | Print an application of a DocName to its list of HsTypes
+ppAppNameTypes :: DocName -> [HsType DocNameI] -> Bool -> LaTeX
+ppAppNameTypes n ts unicode = ppTypeApp n ts ppDocName (ppParendType unicode)
 
 
 -- | Print an application of a DocName and a list of Names
 ppAppDocNameNames :: Bool -> DocName -> [Name] -> LaTeX
 ppAppDocNameNames _summ n ns =
-  ppTypeApp n [] ns (ppBinder . nameOccName . getName) ppSymName
+  ppTypeApp n ns (ppBinder . nameOccName . getName) ppSymName
 
 
 -- | General printing of type applications
-ppTypeApp :: DocName -> [a] -> [a] -> (DocName -> LaTeX) -> (a -> LaTeX) -> LaTeX
-ppTypeApp n [] (t1:t2:rest) ppDN ppT
+ppTypeApp :: DocName -> [a] -> (DocName -> LaTeX) -> (a -> LaTeX) -> LaTeX
+ppTypeApp n (t1:t2:rest) ppDN ppT
   | operator, not . null $ rest = parens opApp <+> hsep (map ppT rest)
   | operator                    = opApp
   where
     operator = isNameSym . getName $ n
     opApp = ppT t1 <+> ppDN n <+> ppT t2
 
-ppTypeApp n ks ts ppDN ppT = ppDN n <+> hsep (map ppT $ ks ++ ts)
+ppTypeApp n ts ppDN ppT = ppDN n <+> hsep (map ppT ts)
 
 
 -------------------------------------------------------------------------------
--- a/haddock-api/src/Haddock/Backends/LaTeX.hs
+++ b/haddock-api/src/Haddock/Backends/LaTeX.hs
@@ -833,7 +833,7 @@ ppDataHeader _ _ = error "ppDataHeader: illegal argument"
 
 -- | Print an application of a DocName and two lists of HsTypes (kinds, types)
 ppAppNameTypes :: DocName -> [HsType DocNameI] -> [HsType DocNameI] -> Bool -> LaTeX
-ppAppNameTypes n ks ts unicode = ppTypeApp n ks ts ppDocName (ppParendType unicode)
+ppAppNameTypes n _ks ts unicode = ppTypeApp n [] ts ppDocName (ppParendType unicode)
 
 
 -- | Print an application of a DocName and a list of Names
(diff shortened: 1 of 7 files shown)


haddock-api/src/Haddock/Convert.hs
```diff
@@ -377,37 +418,100 @@ synifyPatSynSigType ps = mkEmptyImplicitBndrs (synifyPatSynType ps)
...
+      = let (dropped_binders, remaining_binders)
+                  = splitAtList  tys binders
+            result_kind  = mkTyConKind remaining_binders res_kind
+            result_vars  = tyCoVarsOfType result_kind
+            dropped_vars = fvVarSet $
+                           mapUnionFV injectiveVarsOfBinder dropped_binders
+
+        in not (subVarSet result_vars dropped_vars)
+
+    injectiveVarsOfBinder :: TyConBinder -> FV
```
RyanGlScott: As mentioned before, `injectiveVarsOfBinder` (and `injectiveVarsOfType`) are taken straight from `TcSplice`. I would import them, except they're currently not exported from `TcSplice` at the moment.

haddock-api/src/Haddock/Backends/LaTeX.hs
```diff
@@ -832,7 +832,7 @@ ppDataHeader _ _ = error "ppDataHeader: illegal argument"
 
 -- | Print an application of a DocName and two lists of HsTypes (kinds, types)
 ppAppNameTypes :: DocName -> [HsType DocNameI] -> [HsType DocNameI] -> Bool -> LaTeX
-ppAppNameTypes n ks ts unicode = ppTypeApp n ks ts ppDocName (ppParendType unicode)
+ppAppNameTypes n _ks ts unicode = ppTypeApp n [] ts ppDocName (ppParendType unicode)
```
RyanGlScott: As part of making Haddock not render invisible kind arguments, I found myself needing to suppress pretty-printing of kinds here (as well as in `ppAppNamesTypes` in `Haddock.Backends.Xhtml.Decl`). Do you think it's worth refactoring `ppAppNameTypes` to remove these kind arguments altogether, or can you think of a reason we'd want to keep them around here?

alexbiehl:
Hello Ryan, I am on holidays right now and can't look into the code because
I am on mobile. I will take a look after the 20st of September.

In the meantime maybe @hvr or @bgamari can take of this?

To add tests you would

1. add your test file to html-test/src
2. Run html-test with -a (for accept)
3. Add the resulting html files in html-test/ref to the commit.

RyanGlScott:
One thing that crossed my mind—is it better to base this commit off of the `ghc-head` branch (what I'm currently doing), or `master`? Unfortunately, `ghc-head` doesn't seem to build on Travis.

Perhaps relatedly, I'm having enormous difficulty getting `html-test` to work under GHC HEAD, as I'm thwarted with the error `Couldn't find base test dependency: base`.

RyanGlScott:
Actually, ignore what I said about `master` vs. `ghc-head`—this definitely requires the `ghc-head` branch, since this uses features that aren't in GHC 8.2.

RyanGlScott:
@alexbiehl, ping?

alexbiehl:
Ah, sorry! Looks all good to me. Merging when I'm back on a computer.

Am Do., 28. Sep. 2017, 22:44 schrieb Ryan Scott <notifications@github.com>:

RyanGlScott:
I've updated this PR to import `injectiveVarsOfBinder` from GHC (instead of redefining it locally).

alexbiehl:
Thanks Ryan! And sorry for the delay!

Issue #683, opened 2017-09-21
QuickJump


alexbiehl:
- [x] (https://github.com/haskell/haddock/pull/684) Separate QuickJump css from Ocean.css. Currently there are css conflicts when using QuickJump on the package page (see http://hackage.haskell.org/package/Cabal)
- [x] (https://github.com/haskell/haddock/pull/684) Make Quickjump a real Javascript project which bundles all the dependencies in a single javascript file (npm, babel, etc.)
- [x] (https://github.com/haskell/haddock/pull/691) Give a reasonable error message in case the doc-index.json couldn't be downloaded (e.g. in case you are serving docs through `file://`)
- [ ] Finish the hackage doc upload validation (https://github.com/haskell/hackage-server/pull/604)

/cc @timjb @themoritz @eggzilla @hvr

alexbiehl:
It seems scrolling is broken on master: Check out https://hackage.haskell.org/package/lens-4.15.4
![bildschirmfoto 2018-01-06 um 11 24 54](https://user-images.githubusercontent.com/1876617/34638827-44f6b3c4-f2d4-11e7-8141-4ff4e74d9280.png)

/cc @timjb

alexbiehl:
Haddock 2.19.0 with QJ has been released. Closing this for now.

PR #684, opened 2017-09-22
Add compile step that bundles and compresses JS files

timjb:
Also, manage dependencies on third-party JS libraries using NPM. This addresses the second point in #683.

Todo:

* [x] improve JS code using ES2015 features and switch to TypeScript


.ghc.environment.x86_64-darwin-8.2.1
```diff
@@ -0,0 +1,60 @@
+-- This is a GHC environment file written by cabal. This means you can
```
alexbiehl: Please remove this file from the diff. We should probably put .ghc.environment.* to .gitignore.

haddock-api/src/Haddock/Utils.hs
```diff
@@ -326,16 +326,10 @@ makeAnchorId (f:r) = escape isAlpha f ++ concatMap (escape isLegal) r
 
 
 jsFile :: String
```
alexbiehl: Could you rename jsFile to something more meaningful? Now that we have several JavaScript includes this derserves a less generic name I think. How about haddockJsFile?

timjb: Fixed.

.ghc.environment.x86_64-darwin-8.2.1
timjb: Fixed.

alexbiehl:
This patch changes the QuickJump interface. We must not forget to change hackage accordingly: 

It needs to know how to load the new `quick-jump.js` on package pages. Knowing QuickJump would undergo some changes I implemented a versioning scheme in https://github.com/haskell/hackage-server/commit/51d86a372cc28817501cbe07d942e5f3f6ee7afb#diff-bed07c0e9d3155a8916521a5bd83a710R258. Haddock generates a `meta.json` which hackage inspects and chooses which version of QuickJump it has to deal with. 

Currently it decides on the `haddock_version` field  from `meta.json` but I think we should introduce a `quickjump_version` field to `meta.json`. And just like the way interface files are versioned (c.f. https://github.com/haskell/haddock/blob/master/haddock-api/src/Haddock/InterfaceFile.hs#L86) we increment the version everytime a breaking change to the QuickJump interface happens. This way we are decoupled from the haddock version.

timjb:
I don't see the need for a separate new version number. I believe that we can keep the QuickJump interface stable going forward. For instance, this pull request doesn't change the `quickNav.init` interface, it only changes the organization of the JS files (one minified, bundled file instead of three). If we ever need to pass an additional argument to `quickNav.init` we can just do so, without breaking backwards compatibility.

alexbiehl:
Looks good to me. Just one more thing: Please separate the QuickJump css from haddock css.

alexbiehl:
I will prepare the corresponding commit to hackage.

timjb:
> Just one more thing: Please separate the QuickJump css from haddock css.

Done. The layout bug should be gone when `quick-jump.css` is used together with the Hackage CSS.

alexbiehl:
Great! Looks good. Will merge as soon as travis is green.

alexbiehl:
Dang. I think you forgot to update the tests.

timjb:
> Dang. I think you forgot to update the tests.

Yep. I've sent an updated patch.

Issue #560, opened 2016-11-11
Collapsed sections should be visible by default and hidden with javascript


orlitzky:
Haddock docs should be usable without javascript, but they aren't if you view them in a web browser. The e.g. "Examples" sections all load collapsed, and you have no way to expand them. Instead, the default style should display everything, and then javascript (if it is enabled) can be used to hide the content immediately.

PR #690, opened 2017-10-09
Use <details> element for collapsibles

timjb:
The [&lt;details&gt; element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#Specifications) works even when JS is disabled, thereby solving #560. This element is [supported in all major browsers](http://caniuse.com/#feat=details) except Edge.

I've tested it in current versions of all major browsers:

* In Firefox the page works perfectly, no change in look or behaviour to before.
* In Chrome and Safari the CSS `list-style-image` rule replacing the standard triangle indicating whether &lt;details&gt; is open or closed with the Haddock plus/minus sign doesn't work. This will change in the future when these browsers implement the [specification section on the rendering of &lt;details&gt;](https://html.spec.whatwg.org/multipage/rendering.html#the-details-and-summary-elements).
* With Edge, collapsing doesn't work. Everything is visible. This will hopefully be fixed when Microsoft releases a new version supporting &lt;details&gt;. (They have [announced](https://wpdev.uservoice.com/forums/257854-microsoft-edge-developer/suggestions/6261266-details-summary-elements) to be working on an implementation in May.)


alexbiehl:
Nice! Didn't know about that one! Thank you @timjb!

Is this ready to be merged?

timjb:
Yes, this can be merged :-)

PR #692, opened 2017-10-12
Add better support for default signatures in class definitions

### LINKED ISSUE #439: Show default signatures for class members
Especially _minimal complete definition: Nothing_, when there are generics involved is misleading.

### LINKED ISSUE #567: Haddock regression in GHC 8.0.1: redundant documentation for default signatures
With this file:

```haskell
{-# LANGUAGE DefaultSignatures #-}
-- | Documentation for Bug
module Bug where

-- | Documentation for Foo
class Foo a where
  -- | Documentation for bar
  bar :: a -> String
  -- | Documentation for bar's default implementation
  default bar :: Show a => a -> String
  bar = show
```

If `haddock-2.16.1` (shipped with GHC 7.10.3) on it, then it renders like so:

haddock-old

But if I run `haddock-2.17.2` (shipped with GHC 8.0.1) on it, then it renders like so:

haddock-new

I certainly wouldn't have expected the current behavior. The ideal would be one of the proposals in https://github.com/haskell/haddock/issues/439#issuecomment-135090049 (or, failing that, reverting back to the old behavior in `haddock-2.17.2`).

mac-adder:
This should fix #439 and fix #567, though I'm not sure whether the solution is satisfying enough (at least it does limit the number of required changes)

### DIFF
--- a/haddock-api/src/Haddock/Backends/Xhtml/Decl.hs
+++ b/haddock-api/src/Haddock/Backends/Xhtml/Decl.hs
@@ -29,7 +29,6 @@ import Haddock.GhcUtils
 import Haddock.Types
 import Haddock.Doc (combineDocumentation)
 
-import           Control.Applicative   (liftA2)
 import           Data.List             ( intersperse, sort )
 import qualified Data.Map as Map
 import           Data.Maybe
--- a/haddock-api/src/Haddock/Backends/Xhtml/Decl.hs
+++ b/haddock-api/src/Haddock/Backends/Xhtml/Decl.hs
@@ -29,6 +29,7 @@ import Haddock.GhcUtils
 import Haddock.Types
 import Haddock.Doc (combineDocumentation)
 
+import           Control.Applicative   (liftA2)
 import           Data.List             ( intersperse, sort )
 import qualified Data.Map as Map
 import           Data.Maybe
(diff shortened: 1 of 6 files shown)


alexbiehl:
Thank you @mac-adder! I will have a look and think about it. I see why generating a new unique might be necessary but I have to think about the consequences.

alexbiehl:
So basically: 

- If you find a default method you give it a new unique name
- you store documentation for this new name
- when rendering and we encounter a default method, we `uniquify` its name again to see if there is some documentation. 

This seems to be the plan, right?

alexbiehl:
Two other points: 
- Is this also working for reexports?
- And could you add a test for default methods?

mac-adder:
Sorry for the delay ! (I needed some time to set up an environment where I could actually test the changes on the latest versions)

So, unique names are necessary to ensure that comments applied to default signatures will be preserved (unless we change how comments are stored), and using `setNameUnique` / `deriveUnique` gives us "transparent" unique names, which greatly limits the number of changes to be propagated compared to other potential solutions (but I agree that it may not feel completely reliable, so I'm looking forward to receiving feedback about it).

Regarding reexports, it was actually not fully working, and should now be fixed, thanks !

Also, I've just added the new test for default signatures, tell me if it looks fine

RyanGlScott:
@alexbiehl, I think it would be a good thing to have this in Haddock. Is there something more that needs to be done here?

RyanGlScott:
Did I miss something here? Why was this closed?

alexbiehl:
Oh Boy I have no idea

alexbiehl:
Aah. I just deleted the master branch and this pr targeted it. Maybe?

RyanGlScott:
Gotcha. I'm definitely interested in getting this into Haddock, so maybe I'll try to rebase this onto the `ghc-8.4` branch when I get a chance.

ntc2:

PR #694, opened 2017-10-27
Fix Haddock rendering of kind-indexed data family instances

RyanGlScott:
Haddock doesn't render the following code particularly well:

```haskell
{-# LANGUAGE GADTs #-}
{-# LANGUAGE NoImplicitPrelude #-}
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE TypeInType #-}
module Bug where

data Bool = False | True
data Unit = MkUnit

data family Sing (a :: k)
data instance Sing (z :: Bool) where
  SFalse :: Sing False
  STrue  :: Sing True
data instance Sing (z :: Unit) where
  SUnit :: Sing MkUnit
```

With the `ghc-head` branch of Haddock, you get this:

haddock-bad

Yikes! This gives the impression that each `Sing` instance is indexed by the same types, when in reality they are distinguished by their kinds. This is the same bug that surfaced in https://ghc.haskell.org/trac/ghc/ticket/9692, and while that bug was fixed in Template Haskell, the fix never made its way to Haddock. This PR accomplishes that by adapting the [necessary changes from `TcSplice`](http://git.haskell.org/ghc.git/blob/acd355a8fd40d0446c9f737e0f02d92fb5b7b935:/compiler/typecheck/TcSplice.hs#l1684).

After this patch, the above code renders like this, as expected:

haddock-good

### DIFF
--- a/haddock-api/src/Haddock/Convert.hs
+++ b/haddock-api/src/Haddock/Convert.hs
@@ -41,7 +41,8 @@ import TysWiredIn ( listTyConName, starKindTyConName, unitTy )
 import PrelNames ( hasKey, eqTyConKey, funTyConKey, ipClassKey
                  , tYPETyConKey, liftedRepDataConKey )
 import Unique ( getUnique )
-import Util ( compareLength, filterByList, filterOut, splitAtList )
+import Util ( chkAppend, compareLength, dropList, filterByList, filterOut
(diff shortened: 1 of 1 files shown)


alexbiehl:
Thank you Ryan! Looks good to me.

PR #697, opened 2017-10-30
Put Quickjump behind --quickjump flag

alexbiehl:


alexbiehl:
@hvr we should make cabal pass `--quickjump` to haddock.

Issue #699, opened 2017-11-10
Make collapsible sections more visible


Lysxia:
Some libraries, such as base, use these sections to show examples, but those are surprisingly discreet to the point that one might not even see it. Part of the fault lies in haddock users who arguably abuse this feature rather than the tool itself, but there are still ways for haddock to adapt.

- In my experience, these sections have always contained information that would be extremely helpful to the largest number of newcomers, so they should be expanded by default.

- A single switch to collapse/expand everything would be nice too.

- One cause of this visibiility problem may also be that only the box in "**⊞ Examples**", which occupies a very small area, gives an indication that the header is clickable at all. Perhaps adding a background color or some kind of border to these sections would be a stronger hint.

    Compare:  
    ![sc-2017-11-10t08 48 01-05 00](https://user-images.githubusercontent.com/2515201/32661567-66fb0044-c5f5-11e7-9ed5-0344a989b9bd.png)  
    ![sc-2017-11-10t08 57 08-05 00](https://user-images.githubusercontent.com/2515201/32661571-6a9400a2-c5f5-11e7-82c2-6d2af5037a21.png)

PR #700, opened 2017-11-10
Hide instances by default (#698)

Lysxia:
Fixes #698.

PR #701, opened 2017-11-10
Expand user-collapsible sections (#699)

Lysxia:
See #699 for discussion.


alexbiehl:
Nice, thank you. I think you have to update the tests to make Travis happy.

hvr:
Tbh, we made them collapsed by default *because* they would clutter everything. If they're expanded by default what's the point of having anything collapsible?

Lysxia:
@hvr That makes sense. I didn't think this through. I'll think of another way to make these sections more visible...

ntc2:
@Lysxia I'm excited about your #700 , but this PR seems to do the opposite for examples, and I agree with @hvr that it's a bad idea. However, I think adding a button that toggles visibility of all of the examples (but not the instances) could be useful, while leaving the examples collapsed by default.

Lysxia:
@ntc2 Thanks! That sounds like a better idea.

PR #702, opened 2017-11-13
Fix #548 by rendering datatype kinds more carefully

### LINKED ISSUE #548: Haddock always shows kind signatures for types re-exported from other packages
If I run haddock on the following modules

``` haskell
module Foo where

data Foo = MkFoo
```

``` haskell
module Bar (Bar(..), Foo(..)) where

import Foo

data Bar = MkBar
```

the types are formatted as expected, **without** kind-signatures.

However, if I move `Foo` into a separate package it suddenly has a kind-signature in the `Bar` haddocks.

It would be nice if haddock would only display kind signatures when one of the type parameters has a non-`*` kind, otherwise it's just noise IMO.

RyanGlScott:
Before, any data type that was loaded from an interface file (i.e., any data type that passes through `Haddock.Convert`) was displaying utterly wrong kind information, leading to #548. For example, one can see this effect on [the `mtl` Haddocks](http://hackage.haskell.org/package/mtl-2.2.1/docs/Control-Monad-State-Strict.html#t:StateT) on Hackage:

mtl-bad

Yuck! That return kind is utterly bogus, given that it also displays the type variables of `StateT`.

The simplest fix (adopted here) is to only display the _return_ kind. Moreover, since we're dealing with data types here, I opt not to display the return kind if it's `*` (since that is what it will be by default anyways).

After this change, here is what `StateT` looks like now:

mtl-good

(Haddock also now appears to render the kind signature of `m`. I wasn't responsible for making that happen, but it's fortunate that Haddock does this, as it's important to ensure that the same kind information is still presented to the user.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,5 +25,7 @@ TAGS
 .cabal-sandbox
 .ghc.environment.*
 cabal.sandbox.config
+cabal.project.local
+cabal.project.local~
 
 .stack-work/
--- a/haddock-api/src/Haddock/Convert.hs
+++ b/haddock-api/src/Haddock/Convert.hs
@@ -155,7 +155,7 @@ synifyTyCon _coax tc
                                                     -- algebraic data nor newtype:
                                       , dd_ctxt = noLoc []
                                       , dd_cType = Nothing
-                                      , dd_kindSig = Just (synifyKindSig (tyConKind tc))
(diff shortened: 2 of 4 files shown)


alexbiehl:
Lovely! Thanks @RyanGlScott!

PR #703, opened 2017-11-13
Actually render infix type operators as infix

RyanGlScott:
(This is based off of the `ghc-head` branch.)

Currently, infix type operators are always rendered prefix:

tildes-bad

Note how `(~)`, `(~~)`, `(:~:)`, and `(:~~:)` are rendered prefix. This stinks. Let's render them infix! Doing so is actually quite simple.

After this patch, here's what that code looks like now:

tildes-good


alexbiehl:
Thanks Ryan! Couldn't this go in master too?

RyanGlScott:
Perhaps, but the patch wouldn't apply cleanly, as the `ghc-head` branch does some filtering out of invisible types first, which is crucial for making `(~)` and `(~~)` render properly.

alexbiehl:
Alright then. Ping me when you are ready.

PR #706, opened 2017-11-20
Add autogen-modules

angerman:
> Packages using 'cabal-version: >= 1.25' and the autogenerated module Paths_* must include it also on the 'autogen-modules' field besides 'exposed-modules' and 'other-modules'. This specifies that the module does not come with the package and is generated on setup. Modules built with a custom Setup.hs script also go here to ensure that commands like sdist don't fail.

### DIFF
--- a/haddock.cabal
+++ b/haddock.cabal
@@ -146,6 +146,10 @@ executable haddock
       Haddock.Convert
       
       Paths_haddock
+
+    autogen-modules:
+      Paths_haddock
+
   else
     -- in order for haddock's advertised version number to have proper meaning,
     -- we pin down to a single haddock-api version.


alexbiehl:
Please rebase this one to 'ghc-head´ branch too.

angerman:
This should end up in master as well, right? I can try to rebase onto `ghc-head` if that's easier. Will ghc-head be merged into master eventually?

alexbiehl:
Yes, `ghc-head` will be merged into `master` when ghc-8.4.1 is released. Currently we have to make sure that haddock builds with ghc-8.2 branch and as long as your build system/cabal changes are not in there we should land these patches to `ghc-head`.

angerman:
Better, I believe.

angerman:
@alexbiehl looks like the change of base, has killed the CI. :-/ Also for the other PR.

alexbiehl:
Yes, that is expected. Unfortunately CI is a bit broken for ghc-head currently.

PR #708, opened 2017-11-24
WIP Annot backend

thumphries:
This is an attempt to revive the Annot backend. This backend serialises type checking information into a textual format that looks as follows:

```
unHtml
Hydrant.Data
23
5
1
unHtml :: Html -> Builder

unAttribute
Hydrant.Data
45
1
1
unAttribute :: Attribute -> (Text, Text)
```

This file format is then consumed by HsColour to add type annotations to highlighted source code. These type annotations appear when moused over. An example can be seen [here](http://goto.ucsd.edu/~rjhala/Annot/Cabal/src/Distribution-Simple-Haddock.html#hscolour).

This work was [largely done by Ranjit Jhala in 2013](https://github.com/ranjitjhala/haddock-annot). It was fully merged into HsColour, but stalled on the way into Haddock because of an unwanted `syb` dependency. Nowadays we have copied large swaths of SYB into Haddock, so integrating this work is easy enough.

While I would rather see this work reimplemented as part of the fancy new hyperlinked backend, I feel the fact that HsColour already supports this format makes completing this backend worthwhile.

### DIFF
--- a/haddock-api/src/Haddock/Backends/Annot.hs
+++ b/haddock-api/src/Haddock/Backends/Annot.hs
@@ -27,7 +27,7 @@ import IfaceType        (ShowForAllFlag (..))
 import NameSet          (NameSet)
 import Outputable
 import PprTyThing
-import Var              (Var, isId)
+import Var              (Var (..), isId)
 
 import Data.Data
 import Data.Maybe       (catMaybes)
@@ -171,7 +171,15 @@ idPredicate :: Var -> Bool
 idPredicate v =
   isId v && and [
       not (isDictId v)
+    , not (isGeneratedId v)
     ]
+
+-- | Filter out compiler-generated names like @$trModule@ or @$cmappend@.
+isGeneratedId :: Var -> Bool
+isGeneratedId v =
+  case unsafePpr v of
+    ('$':_:_) -> True -- $a but not $
+    _ -> False
 
 findLEs :: Data a => a -> [LHsExpr Id]
 findLEs a = listifyBut (isGoodSrcSpan . getLoc) skipGuards a
--- a/haddock-api/src/Haddock/Backends/Annot.hs
+++ b/haddock-api/src/Haddock/Backends/Annot.hs
@@ -1,0 +1,176 @@
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE RankNTypes #-}
+
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Haddock.Backends.Annot
+-- Copyright   :  (c) Ranjit Jhala
+-- License     :  BSD-like
+--
+-- Maintainer  :  haddock@projects.haskell.org
+-- Stability   :  experimental
+-- Portability :  portable
+--
+-- Write out HsColour compatible type annotations
+-----------------------------------------------------------------------------
+
+module Haddock.Backends.Annot (
+    ppAnnot
+  ) where
+
+import DynFlags         (unsafeGlobalDynFlags)
+import FastString
+import GHC
+import GHC.Exts         (groupWith)
+import Id               (idName, isDictId)
+import IfaceSyn         (ShowSub (..), ShowHowMuch (..))
+import IfaceType        (ShowForAllFlag (..))
+import NameSet          (NameSet)
+import Outputable
+import PprTyThing
+import Var              (isId)
+
+import Data.Data
+import Data.Maybe       (isJust)
+import Haddock.Syb
+import Haddock.Types
+import System.FilePath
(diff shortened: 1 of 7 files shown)


haddock-api/src/Haddock/Backends/Annot.hs
```diff
@@ -0,0 +1,185 @@
...
+-- | Get a list of all entities that meet a predicate
+listifyBut :: Typeable r => (r -> Bool) -> [TypeRep] -> GenericQ [r]
+listifyBut p q
+  = everythingButQ (++) q ([] `mkQ` (\x -> if p x then [x] else []))
+
+-- | Types to avoid inspecting. We should not enter any PostTcKind
+-- nor NameSet because these are blank after type checking.
+skipGuards :: [TypeRep]
+skipGuards = [ typeRep (Proxy :: Proxy NameSet)
+             , typeRep (Proxy :: Proxy (PostTc Id Kind))]
```
thumphries: I'm not totally sure if this is still necessary - I saw no ill effect from turning these off. The rationale behind them is in Tamar's 2011 blog post linked above.

thumphries:
Most of this code is straight out of the 2011 version. I've cleaned up a few bits that bothered me.

It currently descends into deriving clauses, which seems wrong, I'll try to fix it. Leads to typeclass methods appearing underneath the derivation:

screen shot 2017-11-24 at 9 19 29 am

haddock-api/src/Haddock/Backends/Annot.hs
```diff
@@ -0,0 +1,185 @@
...
+findIds a = listifyBut idPredicate skipGuards a
+
+idPredicate :: Var -> Bool
+idPredicate v =
+  isId v && and [
+      not (isDictId v)
+    , not (isGeneratedId v)
+    ]
+
+-- | Filter out compiler-generated names like @$trModule@ or @$cmappend@.
```
thumphries: If there's a better way to do this, I'd love to know about it! The `IdDetails` are completely blank for all IDs by the time it gets here.

haddock-api/src/Haddock/Backends/Annot.hs
```diff
@@ -0,0 +1,185 @@
...
+  where stripParens ('(':s)  = stripParens s
+        stripParens s        = stripRParens (reverse s)
+        stripRParens (')':s) = stripRParens s
+        stripRParens s       = reverse s
+
+getNames ::  (Data a) => FastString -> a -> [(String, (Id, Maybe Loc))]
+getNames src z = [(unsafePpr x, (x, idLoc src x)) | x <- findIds z]
+
+renderId :: Id -> String
+renderId = showSDocForUser unsafeGlobalDynFlags neverQualify . pprTyThing showSub . AnId
```
thumphries: I'm using `unsafeGlobalDynFlags` throughout. Since we run after typechecking this seems safe?

alexbiehl:
Thank you for your effort! This sounds definitely interesting! I will have a look later.

haddock-api/src/Haddock/Backends/Annot.hs
alexbiehl: Please thread `DynFlags` through. Relying on global state breaks my reasoning.

haddock-api/src/Haddock/Backends/Annot.hs
alexbiehl: This is ok I guess.

haddock-api/src/Haddock/Backends/Annot.hs
alexbiehl: Yes, this is ugly but I am not sure there are other ways to check if an `Id` is compiler generated.

haddock-api/src/Haddock/Types.hs
```diff
@@ -151,6 +151,8 @@ data Interface = Interface
     -- | Tokenized source code of module (avaliable if Haddock is invoked with
     -- source generation flag).
   , ifaceTokenizedSrc :: !(Maybe [RichToken])
+
+  , ifaceTcSource :: TypecheckedSource
```
alexbiehl: Let's signal here that `ifaceTcSource` is evaluated.

harpocrates:
FWIW, if the generation of the `AnnMap` moves out of its own module, incorporating this into the hyperlinked backed should be quite easy. In fact, once this gets merged, I don't mind walking through the plumbing of all that myself.

thumphries:
Awesome! Sorry I let this languish, I was busy moving job and country. Will revive the PR this weekend.

ntc2:

PR #712, opened 2017-12-04
Make <code> text more readable in Ocean theme.

skrytebane:
Hi.

Depending on browser/fonts it can be hard to distinguish text in `<code></code>` blocks from regular text, and this makes the Haddock generated documentation harder to read, in my opinion. How about doing something like this instead? (Inspired by bootstrap.css...)

(I'm not a designer, so don't really know for sure if this is a good idea or not.)

### DIFF
--- a/haddock-api/resources/html/Ocean.std-theme/ocean.css
+++ b/haddock-api/resources/html/Ocean.std-theme/ocean.css
@@ -83,9 +83,12 @@ table {
 }
 
 pre, code, kbd, samp, tt, .src {
-	font-family:monospace;
-	*font-size:108%;
-	line-height: 124%;
+    font-family: monospace;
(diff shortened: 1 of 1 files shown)


alexbiehl:
Hey, thank you for your suggestion. Could make a screenshot or demo for me to see how it looks?

hvr:
I tried injecting the .css change into my browser, but it interacts weirdly with doctest markup:

untitled

skrytebane:
Hi. I think I'll have to take a closer look at this. (Also having trouble with the build. Need to learn Cabal. :) )

alexbiehl:
@skrytebane I can only recommend `cabal new-build` for this. Also you will need `ghc-8.2.1` to build the master branch.

Issue #488, opened 2016-03-04
Quasiquoters rendering broken in hyperlinked sources


geraldus:
Awesome hyperlinked sources have incorrect quasiquoters rendering, example http://hackage.haskell.org/package/yesod-form-richtext-0.1.0.0/docs/src/Yesod.Form.Summernote.html#snHtmlFieldCustomized

mpickering:
The problem is that @mrhania implemented a custom lexer rather than reusing GHC's lexer. I tried to port it to use GHC's lexer a few months ago but there were some bad performance problems I didn't work out.

harpocrates:
@mpickering Do you happen to have that branch lying around somewhere? I'd love to pick up the torch on this one, but it would be nice to hear what sorts of issues you had, advice you'd like to give.

Also, how were you measuring performance? I don't see any benchmarks in the `.cabal` files.

alexbiehl:
@harpocrates IIRC this is the branch https://github.com/mpickering/haddock/commits/perf

harpocrates:
I think I've managed to get @mpickering 's branch back into shape. I fixed a handful of things and it seems to be mostly working now (I haven't tried running it on anything large yet, but the performance issues I felt initially went away after rebasing).  I'll open a WIP PR when I get home from work. There are a couple fundamental issues that warrant discussion in any case though:

  * CPP is _not_ handled in the GHC lexer, yet Haddock will have to handle it. Currently, the solution is to recover from _all_ lexical errors by turning problematic lines into comments (and then restarting the lexer from after that line). That seems like an OK failure mode to me, but I think it should be discussed...
  * Language pragmas are not lexed as pragmas. This (probably) requires a GHC change. In the meantime, language pragmas would look like regular comments.
  
    GHC's handling of top-level pragmas is a (IMHO) huge lexer-level hack to try to avoid loading the whole file and we are feeling that here. (If `bytestring` ever becomes a stage 0 package, `Data.ByteString.Lazy` is what that _really_ needs - not a hand coded list-based parser.)

In any case, the actual output looks great!

alexbiehl:
This is fixed in #714. Thank you @harpocrates!

mpickering:
The performance problems I had were obvious trying to use the modified version of haddock at all. It would just take ages on any file and use a lot of memory. Thanks for finishing this for me. 

I don't understand the comment about language pragmas which seem to get lexed to `ITlanguage_prag`?

harpocrates:
[...]

PR #714, opened 2017-12-06
Use the GHC lexer for the Hyperlinker backend

harpocrates:
Lead up discussion is [here](https://github.com/haskell/haddock/issues/488). Moving my concerns here:

  * CPP is not handled in the GHC lexer, yet Haddock will have to handle it. Currently, the solution is to recover from _all_ lexical errors by turning problematic lines into comments (and then restarting the lexer from after that line). That seems like an OK failure mode to me, but I think it should be discussed...

 *  Language pragmas (actually _all_ top-level pragmas) are not lexed as pragmas. This (probably) requires a GHC change. In the meantime, language pragmas would look like regular comments.

    GHC's handling of top-level pragmas is a (IMHO) huge lexer-level hack to try to avoid loading the whole file and we are feeling that here. (If `bytestring` ever becomes a stage 0 package, `Data.ByteString.Lazy` is what that really needs - not a hand coded list-based parser.)

### DIFF
--- a/haddock-api/src/Haddock/Backends/Hyperlinker/Parser.hs
+++ b/haddock-api/src/Haddock/Backends/Hyperlinker/Parser.hs
@@ -6,9 +6,11 @@ import Data.List
 import Data.Maybe
 import qualified Lexer as L
 import Lexer (Token(..))
-import qualified GHC as GHC
+import qualified GHC
 import SrcLoc
 import FastString
+import Data.Either
+import StringBuffer
 
 import Haddock.Backends.Hyperlinker.Types
 import Haddock.Backends.Hyperlinker.Types as T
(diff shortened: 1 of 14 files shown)


harpocrates:
Alright. I think I've actually found an acceptable workaround to my two points from the previous comment. In a nutshell, we will do no better than we did before for parsing CPP and pragmas than we did before.

  - we still identify CPP based on the suboptimal criteria that the line starts with `#`
  - we still assumes anything of the form `{-# ... #-}` is a pragma

These both sound like acceptable workarounds. We could probably do better for the second point, but it would require some non-trivial GHC lexer hacking - which is probably not worth it.

TODO before this is mergeable:

  - [ ] test performance
  - [ ] update tests

alexbiehl:
Nice! Finally a chance to throw the custom lexer over board! This sounds
reasonable! Have you ran the hyperlinker tests?

I think a lot of performance improvement in this area came with
https://github.com/haskell/haddock/commit/740458ac4d2acf197f2ef8dc94a66f9b160b9c3c
which eliminates an accidentally quietadric look-up.

Since we are working with GHCs SrcLoc and SrcSpan we can also eliminate the
custom Position and Span from haddock, right?

Alec Theriault <notifications@github.com> schrieb am Mi., 6. Dez. 2017,
06:31:

harpocrates:
I think this is ready to merge. Sample of changes (before and after):

screen shot 2017-12-07 at 2 09 22 pm screen shot 2017-12-07 at 2 09 34 pm

Here are the only caveats I can think of:

  * Less flexibility around failure. If the lexer fails on some input, the whole chunk of Haskell source (bounded by the nearest CPP lines) will be marked as "unknown". We could conceivably just fall back onto the lexer we had before, but that isn't currently implemented. I'd like to witness _some_ failing chunk of valid Haskell before bothering to do this.
  * CPP is a hack. In a nutshell, the file gets split up into CPP chunks and non-CPP chunks. The claim is that the non-CPP chunks can be lexed just fine on their own. I'm not sure this claim _always_ holds. Besides, our heuristics for separating CPP from non-CPP could be buggy.

**Please do try this out before merging** - I'd like more confidence about performance in particular. 😄 

Thanks!

alexbiehl:
Exciting stuff! Will try this at home!

Boy, the next haddock release is packed with new features and fixes. Awesome!

haddock-api/src/Haddock/Backends/Hyperlinker/Parser.hs
```diff
@@ -1,214 +1,368 @@
...
+    go :: RealSrcLoc -> [Either String String] -> [Located L.Token]
+    go _   [] = []
+    go pos ls =
+      let (hLinesRight,  ls')  = span isRight ls
+          (cppLinesLeft, rest) = span isLeft ls'
+          
+          hSrc   = concat [ hLine   | Right hLine  <- hLinesRight  ]
+          cppSrc = concat [ cppLine | Left cppLine <- cppLinesLeft ]
+      
+      in case L.lexTokenStream (stringToStringBuffer hSrc) pos dflags of
```
alexbiehl: IIRC for each chunk (that is code between some CPP markers) we call the lexer. For each chunk we call `stringToStringBuffer`. If the lexer fails we make the offending part into a comment and try again with the rest, create a new `StringBuffer` and call the lexer again.  Calling `stringToStringBuffer` is pretty expensive and doing it over and over again seems unecessary.

harpocrates: I don't think that is happening. If the lexer fails, that chunk gets marked as one big "unknown" token and we move on (see the `PFailed` case - we just shove `hSrc` into one big token).

alexbiehl:
Right! The result looks really good!

harpocrates:
Thanks for reviewing - I really appreciate how responsive you've been!

Issue #579, opened 2017-03-01
Clickable anchors for headings & more descriptive ids


Profpatsch:
It would be nice to have `<a>` tags to headings that can be copied easily.

Additionally, the generated `id`s for headings are quite non-descriptive at the moment:

```html
<h1 id="g:3">Imports</h1>
```

kautif:
This is my first time trying to contribute to an open source project, so I'm going to ask you to bear with me a little bit.

I'm trying to find the exact location in which this issue is present. Please give a specific folder and file name so that I can locate where to attempt to address this issue. Thanks.

Profpatsch:
> I'm trying to find the exact location in which this issue is present. Please give a specific folder and file name so that I can locate where to attempt to address this issue. Thanks.

You might want to look into `Haddock/Backends/Xhtml.hs`, especially at the `ExportGroup` constructor defined in `Haddock/Types.hs`.

PR #716, opened 2017-12-14
Clickable anchors for headings

harpocrates:
See #579. This just adds an <a> tag around the heading, pointing to the heading itself. I didn't implement the second half of that feature request (more descriptive anchor names) because I actually prefer the short names - they are consistent with the other generated anchor names.

This is a tiny change, but it is something I've also wanted in the past.


alexbiehl:
Yes, this sounds reasonable. Small but useful thing. Will merge when travis is green.

PR #717, opened 2017-12-17
Improve treatment of escaped ']' in definition lists

### LINKED ISSUE #546: Escaped square bracket in definition term produces confused HTML (regression since 2.13.2)
Example:

``` haskell
module Test where

-- |Test:
--
-- [@[code with square brackets\]@]  lorem ipsum
x = 1
```

With Haddock 2.17.2 (and GHC 8.0.1) the relevant snippet from the HTML output looks like:

``` html
<p>Test:</p><dl><dt>@[code with square brackets\</dt><dd>@]  lorem ipsum</dd></dl>
```

Clearly, escaping the `]` didn't work: the definition term ends at the escaped `]` and as a result the `@@` markup isn't handled correctly either. Trying old GHC versions, I was able to reproduce this with Haddock 2.14.3 (GHC 7.8.4), but with Haddock 2.13.2 (GHC 7.6.3) I  got the expected behaviour (minus the extraneous line break after "Test:" but that's irrelevant):

``` html
<p>Test:
</p><dl><dt><code>[code with square brackets]</code></dt><dd>  lorem ipsum
</dd></dl>
```

So apparently this is a regression somewhere between haddock-2.13.2 and haddock-2.14.3-release. [Github compare view of the commits in that range.](https://github.com/haskell/haddock/compare/haddock-2.13.2...haddock-2.14.3-release)

[The docs](https://www.haskell.org/haddock/doc/html/ch03s08.html#idm140354810748256) explicitly allow this:

> Other markup operators may be used freely within the definition term. You can escape ] with a backslash as usual.

alexbiehl:
This fixes #546.

### DIFF
--- a/haddock-library/src/Documentation/Haddock/Parser.hs
+++ b/haddock-library/src/Documentation/Haddock/Parser.hs
@@ -338,13 +338,26 @@ definitionList :: BS.ByteString -> Parser (DocH mod Identifier)
 definitionList indent = DocDefList <$> p
   where
     p = do
-      label <- "[" *> (parseStringBS <$> takeWhile1 (notInClass "]\n")) <* ("]" <* optional ":")
+      label <- "[" *> (parseStringBS <$> scan False accept) <* ("]" <* optional ":")
       c <- takeLine
       (cs, items) <- more indent p
       let contents = parseString . dropNLs . unlines $ c : cs
       return $ case items of
         Left x -> [(label, contents `docAppend` x)]
         Right i -> (label, contents) : i
+
+    -- handle '\]' escapes
+    accept True ']'  = Just False
+
+    -- stop on ']' or '\n'
+    accept _    ']'  = Nothing
+    accept _    '\n' = Nothing
+
+    -- starting an escape sequence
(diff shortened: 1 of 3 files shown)


alexbiehl:
@harpocrates I think we have an open issue regarding the hyperlinker:

``` 
Haddock.Backends.Hyperlinker.Parser

  parse

    is total

    retains file layout FAILED [1]

    when parsing single-line comments

      should ignore content until the end of line

      should allow endline escaping

    when parsing multi-line comments

      should support nested comments

      should distinguish compiler pragma

    should recognize preprocessor directives

    should distinguish basic language constructs

    should parse do-notation syntax


Failures:


  test/Haddock/Backends/Hyperlinker/ParserSpec.hs:48: 

  1) Haddock.Backends.Hyperlinker.Parser.parse retains file layout

       Falsifiable (after 35 tests and 2 shrinks): 

       NoTabs " \n"


Randomized with seed 563507825
```

harpocrates:
@alexbiehl I'll take a look. The issue here is amost certainly related to GHC replacing certain character sequences with others.

PR #718, opened 2017-12-21
Grid Tables

phadej:
Resolves https://github.com/haskell/haddock/issues/530

Based on RST Grid Tables http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#grid-tables,
which are more expressive then Emacs tables https://www.gnu.org/software/emacs/manual/html_node/emacs/Text-Based-Tables.html (AFAICS, strict superset)

Uses & supersedes #577: at least styles, and simple table test.

Trying to improve over #577 
- there are example-based tests using examples from RST spec, and few others.
- spec: the same as RST Grid Table
- *I think you should provide some way to escape | in table cells. Otherwise typing | results in odd parser errors.*, you don't as you can add empty line to the cells or nudge contents a little (see RST spec and/or test examples)
- I'll be happy to add documentation, but would prefer to do it as a separate PR after this one is merged. The spec is clear, so there won't be surprises in syntax description.

ATM the parser is quite lax, it would be handy to add warnings to haddock, so one can report them, but still don't fail (I guess it's possible by adding them to `ParserState`). However, that better done in a separate PR.

```
+------------------------+------------+----------+----------+
| Header row, column 1   | Header 2   | Header 3 | Header 4 |
| (header rows optional) |            |          |          |
+========================+============+==========+==========+
| body row 1, column 1   | column 2   | column 3 | column 4 |
+------------------------+------------+----------+----------+
| body row 2             | Cells may span columns.          |
+------------------------+------------+---------------------+
| body row 3             | Cells may  | \[                  |
+------------------------+ span rows. | f(n) = \sum_{i=1}   |
| body row 4             |            | \]                  |
+------------------------+------------+---------------------+
```


hvr:
:ping_pong:

phadej:
I got something working towards supporint colspan&rowspans based on ideas in http://docutils.sourceforge.net/docutils/parsers/rst/tableparser.py

There /will/ be a problem with UTF-8 as for tables we need to count characters. I won't do anything for that at this point.

phadej:
screenshot from 2017-12-25 02-41-27

phadej:
@hvr @alexbiehl this is ready for review.

alexbiehl:
Oleg, this is looking good. I added a changelog entry and basic documentation. Let's make sure this comes with ghc-8.4.

phadej:
@alexbiehl thanks for writing the docs!

PR #720, opened 2017-12-22
Patch flaky parser test

harpocrates:
This test was a great idea, but it doesn't port over too well to using
the GHC lexer. GHC rewrites its input a bit - nothing surprising, but
we need to guard against those cases for the test.

[Here](https://github.com/haskell/haddock/pull/717#issuecomment-352244382) was the observed failure.


alexbiehl:
Thank you Alec, I fixed one last issue. Merging travis is green.

harpocrates:
There was one other stray `NoTabs`. 😄

alexbiehl:
Alec, thanks for your effort!

Issue #710, opened 2017-11-26
`<<loop>>` while generating doc of svg-tree with GHC 8.2.2


Twinside:
Using head of [svg-tree](https://github.com/Twinside/svg-tree) (equivalent to the one on [hackage](https://hackage.haskell.org/package/svg-tree) ) we get the following result:

```
[11:28:10] ~/Documents/Coding/svg-tree> stack haddock . --resolver=nightly-2017-11-25
... (lots of generation)
Running Haddock on library for svg-tree-0.6.2..
links will be generated to these packages: fail-4.9.0.0
Haddock coverage:
 100% ( 16 / 16) in 'Graphics.Svg.CssTypes'
   0% (  0 /  2) in 'Graphics.Svg.NamedColors'
  Missing documentation for:
    Module header
    svgNamedColors (src\Graphics\Svg\NamedColors.hs:8)
  97% (131 /135) in 'Graphics.Svg.Types'
  Missing documentation for:
    MeshGradientType (src\Graphics\Svg\Types.hs:228)
    HasMeshGradient (src\Graphics\Svg\Types.hs:844)
    HasMeshGradientRow (src\Graphics\Svg\Types.hs:820)
    HasMeshGradientPatch (src\Graphics\Svg\Types.hs:808)
   0% (  0 / 12) in 'Graphics.Svg.PathParser'
  Missing documentation for:
    Module header
    transformParser (src\Graphics\Svg\PathParser.hs:162)
    command (src\Graphics\Svg\PathParser.hs:69)
    pathParser (src\Graphics\Svg\PathParser.hs:66)
    viewBoxParser (src\Graphics\Svg\PathParser.hs:48)
    pointData (src\Graphics\Svg\PathParser.hs:63)
    gradientCommand (src\Graphics\Svg\PathParser.hs:234)
    serializePoints (src\Graphics\Svg\PathParser.hs:110)
    serializeCommand (src\Graphics\Svg\PathParser.hs:132)
    serializeGradientCommand (src\Graphics\Svg\PathParser.hs:248)
    serializeCommands (src\Graphics\Svg\PathParser.hs:129)
    serializeViewBox (src\Graphics\Svg\PathParser.hs:54)
   0% (  0 /  6) in 'Graphics.Svg.ColorParser'
  Missing documentation for:
    Module header
    colorParser (src\Graphics\Svg\ColorParser.hs:56)
    colorSerializer (src\Graphics\Svg\ColorParser.hs:53)
[...]

Issue #679, opened 2017-09-04
Haddock-2.18 hangs forever (ghc-8.2) when using makeClassy


PierreR:
`haddoc-2.18.1`  (ghc-8.2.1) will hang forever with `stack haddock` or `cabal haddock` for the `language-puppet`  package (v-1.3.10)

With the same source, both command works flawlessly in `ghc-8.0.2` (`haddock-2.17.3`)

I have been trying to fix this with `{-# OPTIONS_HADDOCK hide, prune  #-}` without success. `haddock` would hang on another module and spicing modules with the pragma leads nowhere.

There are no error message. The `cpu` level just goes off and the process hangs forever. 

Is there any sane way to debug this ? The fact that it just hangs is so abnormal.

See https://github.com/fpco/stackage/issues/2822

alexbiehl:
I will investigate. But it sounds like a GHC thing to me. Hoping to find
out why.

Pi3r <notifications@github.com> schrieb am Mo., 4. Sep. 2017 um 13:49 Uhr:

PierreR:
@alexbiehl Thanks. Here is the [travis build](https://travis-ci.org/bartavelle/language-puppet) 

I am also able to reproduce this locally on my machine (nixos)

PierreR:
In case it helps this is the end output with `strace`:
```
    puppetDBFacts (src/Facter.hs:174)
  93% ( 27 / 29) in 'Puppet.Interpreter.Resolve'
) = ? ERESTARTSYS (To be restarted if SA_RESTART is set)
--- SIGVTALRM {si_signo=SIGVTALRM, si_code=SI_TIMER, si_timerid=0, si_overrun=0, si_value={int=0, ptr=NULL}} ---
rt_sigreturn({mask=[]})                 = 202
futex(0x21b8254, FUTEX_WAIT_PRIVATE, 0, NULL) = ? ERESTARTSYS (To be restarted if SA_RESTART is set)
--- SIGVTALRM {si_signo=SIGVTALRM, si_code=SI_TIMER, si_timerid=0, si_overrun=0, si_value={int=0, ptr=NULL}} ---
rt_sigreturn({mask=[]})                 = 202
futex(0x21b8254, FUTEX_WAIT_PRIVATE, 0, NULL) = ? ERESTARTSYS (To be restarted if SA_RESTART is set)
[...]

PR #723, opened 2017-12-31
Fix infinite loop when specializing instance heads

harpocrates:
The bug can only be triggered from TH, hence why it probably went un-noticed for so long. I'd like to include the following test-case in `html-test` but I'm not sure how to generate the reference HTML...

```haskell
{-# LANGUAGE TemplateHaskell #-}
{-# OPTIONS_GHC -ddump-splices #-}

module Foo where

import Language.Haskell.TH

data Bar a = Bar

$(do
   a <- newName "a"
   
   let classN = mkName "Foo"
   let methodN = mkName "foo"

   methodTy <- [t| $(varT a) -> $(varT a) |]
   let cla = ClassD [] classN [PlainTV a] [] [SigD methodN methodTy]

   -- Note we are reusing the 'a' variable from before
   instanceHead <- [t| $(conT classN) (Bar $(varT a)) |]
   idCall <- [e| id |]
   let ins = InstanceD Nothing [] instanceHead [FunD methodN [Clause [] (NormalB idCall) []]]
    
   pure [cla,ins])
```

I think this is causing #710 and #679. It so happens that `makeClassy` reuses type variables from class to instance, which runs specializing those instance heads into an infinite loop.

I'm not sure this is correct, but it seems to make the above _not_ loop, so it should at least be a step in the right direction...

### DIFF
--- a/haddock-api/src/Haddock/Interface/Specialize.hs
+++ b/haddock-api/src/Haddock/Interface/Specialize.hs
@@ -31,17 +31,19 @@ import qualified Data.Set as Set
 specialize :: forall name a. (Ord name, DataId name, NamedThing name)
             => Data a
             => [(name, HsType name)] -> a -> a
-specialize specs = go
+specialize specs = go spec_map0
   where
-    go :: forall x. Data x => x -> x
-    go = everywhereButType @name $ mkT $ sugar . specialize_ty_var
-    specialize_ty_var (HsTyVar _ (L _ name'))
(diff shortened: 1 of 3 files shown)


alexbiehl:
Alec, great work! I will take a look in the next days. 

To make your example a valid test case you can do: 

1. Copy your example to `html-test/src/Bug710.hs`
2. `$ html-test -a` will create a ref file for you in `html-test/ref/`

alexbiehl:
Thanks Alec!

Issue #469, opened 2016-01-01
Array documentation is missing its Read instance


rwbarton:
See http://hackage.haskell.org/package/array-0.5.1.0/docs/Data-Array.html, or the corresponding file from the GHC 7.10.3 binary download.

```
...
(Ix i, Eq e) => Eq (Array i e)   
(Ix i, Ord e) => Ord (Array i e)     
(Ix a, Show a, Show b) => Show (Array a b)   
```

But there is also an instance `instance (Ix a, Read a, Read b) => Read (Array a b)`. (It's even documented in the Haskell 2010 Report!)

I'm guessing this is due to the fact that `Array` and its `Eq`, `Ord` and `Show` instances are defined in module `GHC.Arr` in base, while the `Read` instance is defined in module `GHC.Read`, which imports `GHC.Arr`. Still the instance really should be documented.

Blaisorblade:
Based on https://github.com/haskell/haddock/issues/297#issuecomment-239144362, this is probably a dup of #372, but with a diagnosis.
This diagnosis appears to match the case of `Text` in `text` I mentioned [here](https://github.com/haskell/haddock/issues/297#issuecomment-239144362).

RyanGlScott:
This was fixed in https://github.com/haskell/haddock/pull/724, and will work in the Haddock bundled with GHC 8.4.1. Proof:

![haddock](https://user-images.githubusercontent.com/2364661/35779982-68390dce-09a3-11e8-8192-e802664673ed.png)

harpocrates:
@alexbiehl Can we close this?

RyanGlScott:
Indeed we can :)

PR #724, opened 2018-01-06
Pass to GHC visible modules for instance filtering

harpocrates:
The GHC-side `getNameToInstancesIndex` filters out incorrectly some
instances because it is not aware of what modules are visible. On the
Haddock side, we need to pass in the modules we are processing.

On the GHC side, we need to check against _those_ modules when checking
if an instance is visible.

Here is the GHC patch: https://phabricator.haskell.org/D4290

### DIFF
--- a/haddock-api/src/Haddock/Interface.hs
+++ b/haddock-api/src/Haddock/Interface.hs
@@ -58,6 +58,8 @@ import GHC hiding (verbosity)
 import HscTypes
 import FastString (unpackFS)
 import MonadUtils (liftIO)
+import TcRnTypes (tcg_rdr_env)
+import RdrName (plusGlobalRdrEnv)
 
 #if defined(mingw32_HOST_OS)
 import System.IO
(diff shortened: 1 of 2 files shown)


alexbiehl:
Very good catch Alec! Looking forward to this!

harpocrates:
Me too! This sort of bug is the worst from a user perspective - you don't know that there are instances you aren't seeing, and once you do, you cease to trust Haddock.

Please do let me know if there is anything I should/could do to help this along - I understand the scheduling is a bit tight. I will endeavor to be responsive. 😃

harpocrates:
Alright. I circled back and figured out why https://github.com/haskell/haddock/issues/469 (the `Read` instance of `Array`) was missing. The problem turned out to be only tangentially related to `getNameToInstancesIndex`.

We simply weren't updating the `GlobalRdrEnv` when processing modules, which meant that `loadUnqualIfaces` in `getNameToInstancesIndex` would load nothing (in fact, that was my clue: I noticed that it _was_ loading stuff when I would use `:info` in GHCi, but it wasn't loading anything when called from Haddock).

The fix is just to update the environment every time we process a module. There are no extra GHC-side changes.

AFAICT, there are no more missing instances anywhere.

alexbiehl:
I will merge this as is, could you open another ghc diff updating the haddock submodule?

PR #725, opened 2018-01-06
Filter RTS arguments from 'ghc-options' arguments

### LINKED ISSUE #666: haddock-api: haddock does not support +RTS flags in --optghc options
Example failure:
```
$ haddock Main.hs --optghc="+RTS" --optghc=-A512M --optghc=-RTS
haddock: Couldn't parse GHC options: +RTS -A512M -RTS
```

The situation arises when non-cabal build systems use haddock.
For example building ghc itself does not work if we add the following to `mk/build.mk`:
```
SRC_HC_OPTS += +RTS -A512M -RTS
```

At documentation generation phase ghc build will fail. I suggest filtering out all options
from `+RTS` to `-RTS` in `parseGhcFlags` area.

harpocrates:
This fixes #666.

### DIFF
--- a/haddock-api/src/Haddock.hs
+++ b/haddock-api/src/Haddock.hs
@@ -437,13 +437,26 @@ withGhc' libDir flags ghcActs = runGhc (Just libDir) $ do
   _ <- setSessionDynFlags dynflags''
   ghcActs dynflags''
   where
+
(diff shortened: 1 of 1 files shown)


alexbiehl:
Travis has some hiccups I think. I will retrigger the build once more.

Issue #727, opened 2018-01-09
LaTeX backend: declNames error


ugeorge:
When trying to generate LaTeX documentation of a more complex project (i.e. makes use of type classes and type families), Haddock throws the following error and no output is generated:

    haddock: internal error: declaration not supported by declNames
    CallStack (from HasCallStack):
      error, called at utils/haddock/haddock-api/src/Haddock/Backends/LaTeX.hs:257:8 in main:Haddock.Backends.LaTeX

Possible cause: at the [pointed line](https://github.com/haskell/haddock/blob/e812e1a371187b2d0c2a789f3b5afd430163440a/haddock-api/src/Haddock/Backends/LaTeX.hs#L258) `declNames` throws an error instead of handling the remaining type constructors `InstD` and  `DerivD`) safely.

Haddock version: 2.16.1
Cabal version: 2.0.0.1
GHC version: 8.2.1

harpocrates:
Fixed in ae23b4f25a972620686617b5aab5375d5046b1c9.

PR #709, opened 2017-11-25
Constructor and pattern synonym argument docs

harpocrates:
This adds support for

  * docs on arguments of non-record plain data constructors
  * docs on arguments of non-record GADTs data constructors
  * docs on arguments of bundled patterns
  * docs on arguments of non-bundled patterns

Here is the [diff to Phabricator](https://phabricator.haskell.org/D4094) (currently blocked until this PR is done) which introduces changes in the GHC parser to make this possible. Consider the following module:

```haskell
{-# LANGUAGE GADTs, PatternSynonyms #-}

module Test (Foo(..), Boo(Foo, Fo), pattern Bo) where

data Foo
   = Boa             -- ^ doc on the `Boa` constrictor
       Int           -- ^ doc on the `Int` field of `Boa`
       String        -- ^ doc on the `String` field of `Boa`

   | Int             -- ^ doc on the `Int` field of the `:*` constructor
       :*            -- ^ doc on the `:*` constructor
     String          -- ^ doc on the `String` field of the `:*` constructor

infixr 1 `Foo`
infixr 2 `Boa`
infixr 3 :*

data Boo where
  -- | Info about a 'Foo'
  Foo :: Int    -- ^ `Int` field of `Foo`
      -> String -- ^ `String` field of `Foo`
      -> Boo    -- ^ Make a `Boo`

infixr 4 `Boo`

-- | Info about bundled 'Fo'
pattern Fo :: Int    -- ^ an 'Int'
           -> String -- ^ a 'String'
           -> Boo    -- ^ a 'Boo'
pattern Fo x y = Foo x y

infixr 5 `Fo`

-- | Info about not-bundled 'Bo'
pattern Bo :: Int    -- ^ an 'Int'
           -> String -- ^ a 'String'
           -> Boo    -- ^ a 'Boo'
pattern Bo x y = Foo x y

infixr 6 `Bo`
```

Without this PR, this is what the generated docs used to look like:

screen shot 2017-11-24 at 8 40 03 pm

Now, they look like:

screen shot 2017-11-24 at 8 42 12 pm

Note that the constructor arguments only get expanded out if there are docs on at least one of them (same for pattern synonyms). Although this is still WIP, I'd appreciate any thoughts/feedback.

## TODO

 - [x] check if the LaTeX (or some other) backend needs to be also updated
 - [ ] add tests for the new functionality

PR #728, opened 2018-01-09
addresses #727. ppTyFam and declNames are now safe.

ugeorge:
While not *the* ideal solution and rather a quick patch, at least it dumps LaTeX files without errors. These files can further be fed to a stream editor for further fixes and tweaks. 

N.B.: I have successfully used this build (+sed/perl scripts) to generate one chapter of a user manual. PM if interested in the result. I am willing to look further into the matter and contribute in order to get rid of the scripts.

### DIFF
--- a/haddock-api/src/Haddock/Backends/LaTeX.hs
+++ b/haddock-api/src/Haddock/Backends/LaTeX.hs
@@ -115,6 +115,7 @@ ppLaTeXTop doctitle packageStr odir prologue maybe_style ifaces = do
   let tex = vcat [
         text "\\documentclass{book}",
         text "\\usepackage" <> braces (maybe (text "haddock") text maybe_style),
+        text "",
         text "\\begin{document}",
         text "\\begin{titlepage}",
         text "\\begin{haddocktitle}",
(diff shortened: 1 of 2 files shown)


ugeorge:
I now see that @harpocrates is also working on the LaTeX backend in  #709, and is further down in plumbing than I will ever be. I am eagerly awaiting updates. 

> Though nobody complained so far so it is probably not really used by anyone.

 @alexbiehl I disagree! The LaTeX backend *is* used or at least *would be* used by several, even among my acquaintances. But since it has seldom produced satisfactory documents (or any documents at all, see #727), people usually shrug and carry on writing documents from scratch, not wanting to mess with the "black magic" behind it, which is a shame really!

alexbiehl:
George, thank you for you work! I will take a more detailed look at your work in the next few days.

harpocrates:
@ugeorge I fear you will be disappointed. My changes were only around fixing data declarations / pattern synonyms. That said, I wouldn't mind whipping up support for all declarations if useful. Would it be?

ugeorge:
@harpocrates Sure, that would be wonderful! I could also lend a hand during weekends as far as I can manage. I can use #727 as a discussion thread to identify the particular declarations I am interested in for prioritizing, but I need a bit of time to look into the matter. In any case, thanks for your work!

ntc2:

Issue #300, opened 2014-05-09
really confusing / noisy haddocks for typeclass + data family heavy code 


cartazio:
when i generate the haddocks for 
https://github.com/wellposed/numerical-core/blob/master/src/Numerical/Array/Layout/Base.lhs

It may very well be that the way its rendering is sensible,
but one concern I have is that the comments are shifted FAR to the right
for a given data family instance, and thats a bit unsatisfactory

(edit: i've now made a copy of the actual haddocks online) http://www.wellposed.com/examples-cached/Numerical-Array-Layout-Base.html should link to the relevant html

using ghc 7.8.2 haddock, heres some screen caps 

![screen shot 2014-05-09 at 6 44 43 pm](https://cloud.githubusercontent.com/assets/4290/2934193/cc2d7634-d7cb-11e3-86b3-49ea02ad91bc.png)
![screen shot 2014-05-09 at 6 44 33 pm](https://cloud.githubusercontent.com/assets/4290/2934194/cc3968fe-d7cb-11e3-8f68-44b75e881abd.png)

tomsmalley:
Is there any movement on this? It's particularly noticeable in the servant-server docs: http://hackage.haskell.org/package/servant-server-0.10/docs/Servant-Server.html.

I think it would be good to wrap constraints in some `inline-block` `span`s and allow wrapping in the `td`, thereby breaking very long lines on the commas and `=>`.

Fuuzetsu:
No movement.

cartazio:
Bump: this issue is probably still biting code I write

cartazio:
One direction might be to render type fam vs type clsss vs data fam instances in distinct groups?

There’s also the funny issue that if I define a single new Data fam constructor and export it, the docs in that module show alll visible constructors/ instances. So another collapsible axis is perhaps locally defined vs everything else that reaches that module?

cartazio:
@harpocrates any thoughts on this one?  It was a bad enough leakage in some of my code that I redid how I wrote an api or two to avoid the extra noise in the docs! (Like an entire copy of vector instances being listed for everything local plus the ones coming from vector ...)

harpocrates:
I agree this is gross. The main blocker here is just figuring out a nice alternate design. (I've contemplated this issue a fair bit and I will get around to it eventually - but don't hold your breath. 😄)

[...]

PR #734, opened 2018-01-16
Latex type families

harpocrates:
This PR adds some support for type/data families in the LaTeX backend, and cleans up some of the existing code around type/data families in the XHTML backend.

The XHTML backend has not changed in behaviour, save to use `newtype` instead of `data` for newtype data instances. The following use to not be accepted by the LaTeX backend.

```haskell
{-# LANGUAGE TypeFamilies #-}

module Test where

-- | A closed type family
type family Foo a where
  Foo () = Int
  Foo _ = ()

-- | An open family
type family Bar a

type instance Bar Int = ()
type instance Bar () = Int

-- | A data family 
data family Baz a

data instance Baz () = Baz1
data instance Baz Int = Baz2 Bool
newtype instance Baz Double = Baz3 Float
```

Now, it produces the following:

screen shot 2018-01-16 at 1 27 50 pm

Obviously, it would be nice to also print out the data definitions behind these data instances. I haven’t done that yet mainly because I’m not sure what would look nicest. I’m not satisfied with what the XHTML backend does here either. In particular, I’d like to have a solution that uses the documentation that may be attached to these instances - but I’m not sure yet how this would be best achieved. When I do come up with something, I’ll propose it in a separate PR.

@ugeorge

### DIFF
--- a/haddock-api/src/Haddock/Backends/LaTeX.hs
+++ b/haddock-api/src/Haddock/Backends/LaTeX.hs
@@ -12,9 +12,8 @@
 -- Portability :  portable
 -----------------------------------------------------------------------------
 module Haddock.Backends.LaTeX (
-  ppLaTeX
+  ppLaTeX,
 ) where
-
 
 import Documentation.Haddock.Markup
 import Haddock.Types
@@ -285,7 +284,7 @@ ppDecl :: LHsDecl DocNameI                         -- ^ decl to print
        -> LaTeX
 
 ppDecl decl pats (doc, fnArgsDoc) instances subdocs _fxts = case unLoc decl of
-  TyClD _ d@FamDecl {}         -> ppTyFam doc instances d unicode
+  TyClD _ d@FamDecl {}         -> ppFamDecl doc instances d unicode
   TyClD _ d@DataDecl {}        -> ppDataDecl pats instances subdocs (Just doc) d unicode
   TyClD _ d@SynDecl {}         -> ppTySyn (doc, fnArgsDoc) d unicode
 -- Family instances happen via FamInst now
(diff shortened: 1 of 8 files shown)


ugeorge:
I have trouble installing your branch. It it pretty much the same reason Travis complains about. 

 * when using ghc-8.2.1, it complains that cabal rejects `base-4.10.0.0`, as your new `cabal-api` depends on

       build-depends: base            ^>= 4.11.0
                    , Cabal           ^>= 2.0.0
                    , ghc             ^>= 8.3
                    , ghc-paths       ^>= 0.1.0.9
                    , haddock-library ^>= 1.4.6
                    , xhtml           ^>= 3000.2.2
    
 * when using ghc-8.4.1, it complains that `async` is rejected (dependency of `hspec-core-2.4.7`). Apparently the "last one" has a dependency constraint `base>=4.3 && <4.11`. 

What setup are you using, so that I might be able to reproduce?

ugeorge:
I would like to try out your branch, but I have trouble setting it up. Any specific reason why your `haskell-api-2.18.2` is restricted to `ghc-8.3` and `base-4.11`? It does not compile with other (newer) releases, and needless to say "cheating" the Cabal file got me nowhere. I tried with `ghc-8.2.2` and `ghh-8.4.1`. It is strange that the only reference to [`ghc-8.3-start`](https://github.com/ghc/ghc/releases/tag/ghc-8.3-start) dates all the way back to March last year, and no other sign was seen since then. As there is no "official" release/installer, am I to clone/install that one?

alexbiehl:
George, unfortunately the situation in ghc-head/ghc-8.4 branches is less than optimal. To build those I always pass '--allow-newer' to cabal and hope for the best. Especially the test suite does not compile without much effort as the used libraries do not yet support ghc >= 8.4.

alexbiehl:
Alec, is it necessary to live in ghc-head? Should this be merged to ghc-8.4?

harpocrates:
I have no preference regarding where this is merged. Unless you want to resolve the merge conflicts though, I recommend you put this in a branch with https://github.com/haskell/haddock/pull/709 (wherever it ends up).

I suggested `ghc-head` because, at least right now, those changes are still only in `ghc-head`. If  GHC HQ ends up moving to 8.4 the GHC-side changes https://github.com/haskell/haddock/pull/709 needed (see [Support constructor Haddocks in more places](https://phabricator.haskell.org/D4292)), you can always move this to `ghc-8.4` when you move https://github.com/haskell/haddock/pull/709.

Or you can even hold off on merging this at all until the fate of https://github.com/haskell/haddock/pull/709 has been decided... 😄

harpocrates:
I've rebased and added tests.

harpocrates:
@alexbiehl Bump? I have some improvements for data family rendering ([preview here](https://github.com/haskell/haddock/issues/300#issuecomment-402349593)) but they are queued up behind these changes.

Issue #526, opened 2016-06-11
Show instances for associated types


vagarenko:
See this page for example: http://hackage.haskell.org/package/reflex-0.4.0/docs/Reflex-Class.html
It has `Reflex` class with associated type `Behavior`.
If you dig into the source code (http://hackage.haskell.org/package/reflex-0.4.0/docs/src/Reflex-Class.html) you'll find that `Behavior` has instances for `Functor`, `Applicative`, `Monad`, `Semigroup`, and `Monoid` which are not present on the haddock page.

Can you show instances of associated types of a class along with the instances for that class?

PR #736, opened 2018-01-19
 extractDecl: Correctly extract associated types 

### LINKED ISSUE #735: GHC 8.4.1-alpha panic (extractDecl)
I originally observed this in [this Travis build](https://travis-ci.org/ekmett/ad/jobs/328314307#L1535). Here is a minimal way to reproduce the issue:

```haskell
{-# LANGUAGE FlexibleContexts #-}
{-# LANGUAGE TypeFamilies #-}
module Bug (Scalar) where

class (Num (Scalar t)) => Mode t where
  type Scalar t
```

In `haddock-2.18.1` (the Haddock bundled with GHC 8.2.2), this works fine:

```
$ /opt/ghc/8.2.2/bin/haddock Bug.hs --html
Haddock coverage:
   0% (  0 /  1) in 'Bug'
  Missing documentation for:
    Module header
```

However, in `haddock-2.18.2` (the Haddock bundled with GHC 8.4.1-alpha), this panics:

```
$ /opt/ghc/8.4.1/bin/haddock Bug.hs --html
Haddock coverage:
haddock: panic! (the 'impossible' happened)
  (GHC version 8.4.0.20171222 for x86_64-unknown-linux):
        extractDecl

Ambiguous decl for Scalar in class:
    class (Num (Scalar t)) => Mode t where
      type Scalar t
      {-# MINIMAL #-}
Matches:
    []
Call stack:
      callStackDoc, called at compiler/utils/Outputable.hs:1150:37 in ghc:Outputable
      pprPanic, called at utils/haddock/haddock-api/src/Haddock/Interface/Create.hs:1013:16 in main:Haddock.Interface.Create
```

alexbiehl:
(fixes #735)

Generating documentation for
```haskell
{-# LANGUAGE FlexibleContexts #-}
{-# LANGUAGE TypeFamilies #-}
module Bug (Scalar) where

class (Num (Scalar t)) => Mode t where
  type Scalar t
```
Results in 

image

- [x] Do we expect a "class context" here to signify that this is an associated type? What would it look like?
- [ ] Add a test
- [ ] Merge to `ghc-head` branch

### DIFF
--- a/haddock-api/src/Haddock/Interface/Create.hs
+++ b/haddock-api/src/Haddock/Interface/Create.hs
@@ -999,21 +999,31 @@ extractDecl name decl
   | otherwise  =
     case unLoc decl of
       TyClD d@ClassDecl {} ->
-        let matches = [ lsig
-                      | lsig <- tcdSigs d
-                      , ClassOpSig False _ _ <- pure $ unLoc lsig
-                        -- Note: exclude `default` declarations (see #505)
-                      , name `elem` sigName lsig
-                      ]
+        let
+          matchesMethod =
+            [ lsig
+            | lsig <- tcdSigs d
+            , ClassOpSig False _ _ <- pure $ unLoc lsig
+              -- Note: exclude `default` declarations (see #505)
+            , name `elem` sigName lsig
+            ]
+
+          matchesAssociatedType =
+            [ lfam_decl
+            | lfam_decl <- tcdATs d
+            , name == unLoc (fdLName (unLoc lfam_decl))
+            ]
+
             -- TODO: document fixity
-        in case matches of
-          [s0] -> let (n, tyvar_names) = (tcdName d, tyClDeclTyVars d)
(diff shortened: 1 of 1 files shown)


RyanGlScott:
This looks great!

To answer the question in the first bullet point: no, I don't have a strong opinion about whether we should print anything about the surrounding class. Inside of GHC, associated type families are desugared into top-level type families, so it's not technically inaccurate to present one independently of its surrounding class.

Moreover, since this is a scenario where we're specifically exporting the associated type family but _not_ the surrounding class, perhaps it's better to simply not mention the surround class at all in the Haddocks.

alexbiehl:
Alright thanks for the feedback!

gbaz:
Does this get things any closer on https://github.com/haskell/haddock/issues/526 and https://github.com/haskell/haddock/issues/664 ?

alexbiehl:
Yes.

gbaz <notifications@github.com> schrieb am Mo., 5. Feb. 2018, 08:54:

PR #738, opened 2018-01-22
Properly color pragma contents in hyperlinker

harpocrates:
The hyperlinker backend now classifies the content of pragmas as `TkPragma`. That means that in something like `{-# INLINE foo #-}`, `foo` still gets classified as a pragma token.

Given the following source:

```haskell
module Kaa where

kaa :: Int -> Int
kaa = (+1)
{-# INLINE [1] kaa #-}

koo :: Int -> ()
koo _ = ()
```

Output prior to this PR:

screen shot 2018-01-22 at 11 52 03 am

New output:

screen shot 2018-01-22 at 11 52 11 am

---

Apart from this, the only thing that remains to be done for the hyperlinker backend is to add a case to the `classify`/`inPragma` functions for the new `ITcolumn_prag` that https://phabricator.haskell.org/D4336 will introduce.

### DIFF
--- a/haddock-api/src/Haddock/Backends/Hyperlinker/Parser.hs
+++ b/haddock-api/src/Haddock/Backends/Hyperlinker/Parser.hs
@@ -136,22 +136,35 @@ spanToNewline n (c:str) =
 -- | Turn a list of GHC's 'L.Token' (and their source 'String') into a list of
 -- Haddock's 'T.Token'.
 ghcToks :: [(Located L.Token, String)] -> [T.Token]
-ghcToks = reverse . snd . foldl' go (start, [])
+ghcToks = reverse . (\(_,ts,_) -> ts) . foldl' go (start, [], False)
   where
     start = mkRealSrcLoc (mkFastString "lexing") 1 1
 
-    go :: (RealSrcLoc, [T.Token])    -- ^ current position, tokens accumulated
-       -> (Located L.Token, String)  -- ^ next token, its content
-       -> (RealSrcLoc, [T.Token])    -- ^ new position, new tokens accumulated
-    go (pos, toks) (L l tok, raw) = ( next_pos
-                                    , classifiedTok ++ maybeToList white ++ toks
-                                    )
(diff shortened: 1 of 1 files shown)


harpocrates:
Since https://phabricator.haskell.org/D4336 has been merged, I've added a commit for supporting `ITcolumn_prag`. Unfortunately, it won't build except on GHC HEAD. However, the lexer changes are only in `master` (and not `ghc-head`) because I didn't initially think they would require GHC changes.

I'm not sure exactly what cutting a release involves, but I figure you'll be bumping the compiler version and cherry-picking changes from `ghc-head` to `master`. Once the lexer changes make it into `master` this should be mergeable.

This is the last change I'll bug you to try to squeeze into 8.4. 😄 Thanks!

mpickering:
I’m not sure this is a desirable change. Does ‘foo’ still get linked to it’s definition? What was wrong before with highlighting the variable differently?

harpocrates:
@mpickering AFAIK `foo` never got linked to its definition. The comment-like syntax of pragmas is the main reason I want to preserve the existing behaviour of giving them only one colour.

alexbiehl:
Alec, looks good. Do we need this in ghc-8.4? I convinced Ben to merge https://phabricator.haskell.org/D4336. Is this PR needed to make use of D4336?

harpocrates:
> Do we need this in ghc-8.4? I convinced Ben to merge https://phabricator.haskell.org/D4336. Is this PR needed to make use of D4336?

Yes, this is needed to support `ITcolumn_prag` (added in D4336). Thanks!

alexbiehl:
Oh boy, alright then!

alexbiehl:
This is a PR against `master` which does not pass travis right now. I applied your patches to ghc-8.4. Closing this PR.

harpocrates:
@alexbiehl Thanks, and sorry for any confusion. I opened this against master since the lexer changes where at the time only there.

Now that I see what the release process is like, I'll definitely try to make my patches smaller.

PR #744, opened 2018-02-05
Hyperlink pattern synonyms and 'module' imports

### LINKED ISSUE #731: Hyperlinked source links to pattern synonyms don't work
In the hyperlinked sources, clicking on a pattern synonym does _not_ bring you to its definition. Ditto for modules in export lists.

harpocrates:
@alexbiehl  Your comments in #731 made me see how simple it would be to fix this, so I did.

----

Links to pattern synonyms are now generated, as well as links from `module`s in import lists.

Fixes #731.


alexbiehl:
Nice, thanks Alec. Looking at the code I think one thing might be missing still: considering 'pattern P x y = ...  x ... y... ' will x and y be highlighted and clickable just like function arguments? (sorry, Cant try your code right now l, on the phone)

harpocrates:
Nope, that case is already covered. :)

alexbiehl:
Great let's merge. I will put this in ghc-8.4 branch. Let's see where Ben cuts off.

PR #745, opened 2018-02-05
Don't barf on 'HsSpliceTy'

### LINKED ISSUE #574: haddock fails with secret-sharing
as said on #233 

The latest version of haddock (version 2.17.3) fails when generating docs for secret-sharing-1.0.0.3 with:

haddock: internal error: renameType: HsSpliceTy
  error, called at utils/haddock/haddock-api/src/Haddock/Interface/Rename.hs:266:30 in main:Haddock.Interface.Rename

Haddock version 2.17.2 works fine.

G.

harpocrates:
This handles `HsSpliceTy`s by replacing them with what they expand to. IIUC everything that is happening, `renameHsSpliceTy` should not be able to fail for the inputs we feed it from GHC.

This fixes #574.

---

Troublesome snippet from `secret-sharing` that was causing the error (FYI, this is [`primeField`](https://hackage.haskell.org/package/finite-field-0.9.0/docs/Data-FiniteField-PrimeField.html#v:primeField)):

```haskell
-- | A finite prime field. All computations are performed in this field.
newtype FField = FField { number :: $(primeField $ fromIntegral 1021) }
  deriving(Show,Read,Ord,Eq,Num,Fractional,Generic,Typeable,FiniteField)
```

Snippet of Haddock produced after this PR

screen shot 2018-02-05 at 1 51 24 pm

### DIFF
--- a/haddock-api/src/Haddock/Interface/Rename.hs
+++ b/haddock-api/src/Haddock/Interface/Rename.hs
@@ -264,9 +264,20 @@ renameType t = case t of
   HsCoreTy a              -> pure (HsCoreTy a)
   HsExplicitListTy i a b  -> HsExplicitListTy i a <$> mapM renameLType b
   HsExplicitTupleTy a b   -> HsExplicitTupleTy a <$> mapM renameLType b
-  HsSpliceTy _ _          -> error "renameType: HsSpliceTy"
+  HsSpliceTy s _          -> renameHsSpliceTy s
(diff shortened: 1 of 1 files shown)


alexbiehl:
Nice! Thanks Alec (btw. could you open your next PRs to the GHC-8.4 branch? It will become the default branch once ghc-8.4.1 is released)

PR #746, opened 2018-02-05
Don't warn about missing '~'

### LINKED ISSUE #532: Add exception for ~
Haddock (at least version 2.17.2 with GHC 8.0.1) reports 

```
could not find link destinations for:
    ~
```

in modules that use type equality. Since `~` is [magical](https://ghc.haskell.org/trac/ghc/ticket/10056) and doesn't actually have a definition anywhere, it shouldn't be reported as a warning.

harpocrates:
This manually filters out `~` from the list of things to warn about. It truly makes no sense to warn on this since `~` has nothing it could link to - it is magical.

This fixes #532.


alexbiehl:
This is what we want! Thanks Alec!

PR #747, opened 2018-02-06
Support unicode operators, proper modules

### LINKED ISSUE #458: Unicode operators
At the moment, Haddock doesn't link unicode identifiers. The sequence `@('⊆')@` just renders as `('⊆')`.

However, when I looked in the source to hopefully fix it, I found a comment in the `moduleName` parser saying "But as we **can't** match on unicode characters, this is currently not possible." I couldn't find any mention anywhere of _why_ unicode identifiers can't be matched, though. So I was hoping someone more familiar with the codebase could clarify.

harpocrates:
Unicode operators are a pretty big thing in Haskell, so supporting linking them seems like it outweighs the cost of the extra machinery to force Attoparsec to look for Unicode.

Fixes #458.

### DIFF
--- a/haddock-library/src/Documentation/Haddock/Parser/Monad.hs
+++ b/haddock-library/src/Documentation/Haddock/Parser/Monad.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE GeneralizedNewtypeDeriving, TypeFamilies, BangPatterns #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving, TypeFamilies #-}
 module Documentation.Haddock.Parser.Monad (
   module Documentation.Haddock.Parser.Monad
 , Attoparsec.isDigit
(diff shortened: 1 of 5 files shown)


harpocrates:
Oops! Just realized `peekUnicode` was stricter than necessary.

haddock-library/src/Documentation/Haddock/Parser/Monad.hs
```diff
@@ -73,6 +75,25 @@ char = lift . Attoparsec.char
...
 char8 = lift . Attoparsec.char8
 
+-- | Peek a unicode character and return the number of bytes that it took up
+peekUnicode :: Parser (Char, Int)
+peekUnicode = lift $ Attoparsec.lookAhead $ do
+  
+  -- attoparsec's take fails on shorter inputs rather than truncate
+  bs <- Attoparsec.choice (map Attoparsec.take [4,3,2,1])
+  
+  let c = head . decodeUtf8 $ bs
```
alexbiehl: This is all so horrible. But I agree with you that it is the simplest thing possible. Once GHC ships `text` as boot library we could think about rewriting the parser to avoid these hacks.

harpocrates: I agree this is madness. Using `Data.Attoparsec.Text` would be so much cleaner.

alexbiehl: On a related note: Changing the types from `String` to `Text` will have an impact on interface size for haddock. Currently it always stores `Char` as 4 byte. Using a more compact encoding like utf-8 will reduce interface size immensely.

alexbiehl:
Alec, I manually rebased and pushed your commits to `ghc-8.4` branch. Closing this. Thank you very much!

Issue #296, opened 2014-05-08
Show where instances are coming from


ghc-mirror:
**Original reporter**: _mail+haddock-trac@<hidden>_

Haddock puts orphan instances, in whatever module they may be declared, into the instances list under a data type declaration's documentation.

That can be very confusing, because haddock suggests an instance is available when after importing the module when it is actually only available after importing another module.

Example: http://hackage.haskell.org/package/vect-0.4.7/docs/Data-Vect-Double-Base.html#t:Vec3

Here Haddock suggests `instance Eq Vec3`, but when loading `Data.Vect.Double.Base` with GHCI it will complain such instance does not exist. The actual instance is in `Data.Vect.Double.Instances` (http://hackage.haskell.org/package/vect-0.4.7/docs/Data-Vect-Double-Instances.html), but in contrast to GHCI Haddock looks at instances globally, across the whole package.

It would be great if Haddock could display a link next to orphan instances that make clear from what module the instance comes.

Maybe even better, Haddock could do this for _all_ instances, so that we can directly jump to the source like we can for functions.

PR #721, opened 2017-12-26
 Improve appearance and readability

NunoAlexandre:
## Motive

I work with haddock and hackage almost daily, both at work and at home, and the current theme is unfriendly. I find it very hard to read content with such a wide `#content`, and spacing is not great either. 

Furthermore, I'd say many Haskell developers would appreciate the presense of the latest Haskell's logo (purple) colors in these docs. 

## Changes

- use latest Haskell's logo colors
- decrease #content width to improve readability
- use, custom nicer font: [Merriweather Sans](https://fonts.google.com/specimen/Merriweather+Sans)
- improve sizes and distances

## Preview

**EDIT 04-02-2018**: See https://github.com/haskell/haddock/pull/721#issuecomment-362934997

<br>
Using: https://hackage.haskell.org/package/lens-tutorial-1.0.3/docs/Control-Lens-Tutorial.html

**Before:**

before

**After:**
after

haddock-after-tablet-standing

haddock-after-tablet-laying

haddock-after-iphone7

PR #748, opened 2018-02-09
Show where instances are defined

harpocrates:
This is analogous to what GHCi's `:info` command does - it displays in the instance details where the instance was defined.

```
ghci> :info Bool
data Bool = False | True        -- Defined in ‘GHC.Types’
instance Bounded Bool -- Defined in ‘GHC.Enum’
instance Enum Bool -- Defined in ‘GHC.Enum’
instance Eq Bool -- Defined in ‘GHC.Classes’
instance Ord Bool -- Defined in ‘GHC.Classes’
instance Read Bool -- Defined in ‘GHC.Read’
instance Show Bool -- Defined in ‘GHC.Show’
instance Ix Bool -- Defined in ‘GHC.Arr’
```

Here is a snippet of the instances for `Bool` (Note the "(Defined in ...)" links):

screen shot 2018-02-08 at 6 23 56 pm

This fixes https://github.com/haskell/haddock/issues/296 as much as it can ever really hope to be fixed.

### DIFF
--- a/haddock-api/src/Haddock/Backends/Xhtml/Decl.hs
+++ b/haddock-api/src/Haddock/Backends/Xhtml/Decl.hs
@@ -566,8 +566,8 @@ ppInstances links origin instances splice unicode qual
   where
     instName = getOccString origin
     instDecl :: Int -> DocInstance DocNameI -> (SubDecl,Located DocName)
-    instDecl no (inst, mdoc, loc) =
-        ((ppInstHead links splice unicode qual mdoc origin False no inst), loc)
+    instDecl no (inst, mdoc, loc, mdl) =
+        ((ppInstHead links splice unicode qual mdoc origin False no inst mdl), loc)
 
 
 ppOrphanInstances :: LinksInfo
@@ -581,8 +581,8 @@ ppOrphanInstances links instances splice unicode qual
     instOrigin inst = OriginClass (ihdClsName inst)
 
     instDecl :: Int -> DocInstance DocNameI -> (SubDecl,Located DocName)
-    instDecl no (inst, mdoc, loc) =
-        ((ppInstHead links splice unicode qual mdoc (instOrigin inst) True no inst), loc)
+    instDecl no (inst, mdoc, loc, mdl) =
+        ((ppInstHead links splice unicode qual mdoc (instOrigin inst) True no inst mdl), loc)
 
 
 ppInstHead :: LinksInfo -> Splice -> Unicode -> Qualification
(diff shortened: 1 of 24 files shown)


harpocrates:
Woops! I had forgotten about data and type families (I've never been bitten by this, but I can imagine  it would be annoying to have a type family be stuck just because you forgot to import the right module...). Here is what that looks like:

screen shot 2018-02-08 at 7 00 16 pm

haddock-api/src/Haddock/Types.hs
```diff
@@ -388,7 +388,7 @@ mkPseudoFamilyDecl (FamilyDecl { .. }) = PseudoFamilyDecl
 
 
 -- | An instance head that may have documentation and a source location.
-type DocInstance name = (InstHead name, Maybe (MDoc (IdP name)), Located (IdP name))
+type DocInstance name = (InstHead name, Maybe (MDoc (IdP name)), Located (IdP name), Maybe Module)
```
alexbiehl: Unfortunately this is a breaking change :( This has to go into the next major version.

harpocrates: I suspected as much... Should I reopen this PR on `ghc-head` then?

haddock-api/src/Haddock/Backends/Xhtml/Decl.hs
```diff
@@ -614,11 +615,12 @@ ppInstHead links splice unicode qual mdoc origin orphan no ihd@(InstHead {..}) =
         DataInst dd ->
             ( subInstHead iid pdata
             , mdoc
-            , [subFamInstDetails iid pdecl])
+            , [subFamInstDetails iid pdecl mname])
           where
             pdata = keyword "data" <+> typ
             pdecl = pdata <+> ppShortDataDecl False True dd [] unicode qual
   where
+    mname = maybe noHtml (\m -> hsep [toHtml "(Defined in", ppModule m, toHtml ")"]) mdl
```
Fuuzetsu: Please leave out the parenthesis, they look bad and are completely unnecessary.

harpocrates: Done.

harpocrates:
@alexbiehl Gentle ping on this. If you are waiting to merge this after #721, that's fine - I just want to make sure this isn't forgotten. Once this gets merged, I'll also fix the tests on  #749.

alexbiehl:
Alec, sorry. I didn't notice you pushed an update to this one. Merging right away.

Issue #266, opened 2014-05-08
add a --show=MODULE and/or --ignore-hide to complement --hide=MODULE


ghc-mirror:
**Original reporter**: _carter.schonwald@<hidden>_

add a --show=MODULE and/or --ignore-hide to complement --hide=MODULE

having some sort of --ignore-hide flag would be great for making

cabal haddock --internal  behave as expected, because currently, cabal haddock --internal won't generate the docs for an exposed module which has a module level pragma telling haddock to hide that.

cartazio:
Yes I guess I filed this one too.   :)
Both the show module foo and  more global ignore hide flavors would be useful. :)

harpocrates:
@cartazio https://github.com/haskell/haddock/pull/752 Implements `--show=MODULE` and `--show-all`. Does that cover your use cases?

cartazio:
Should be!  Unless there’s other things they mean than what I imagine.

It does make me curious how cabal emulates the show internal stuff in
global cabal confit

harpocrates:
Good! Once you get to try this out, feel free to close the issue. 

> It does make me curious how cabal emulates the show internal stuff in global cabal config

Wonder no more and discover for yourself! If you pass in `-v` to cabal, you can see enough debug output of commands that are run to piece together what it is doing. Normally, Cabal passes `--hide=<mod>` for every internal module. When you pass it the show internal stuff flag, it just skips over hiding the internal modules.

cartazio:
And now show / show all ignores module hide pragmas?

harpocrates:
> And now show / show all ignores module hide pragmas?

Yes. More accurately, it overrides hide (either passed in the command line earlier, or in the file's `OPTIONS_HADDOCK`). So, short of patching the `--show-all` flag into Cabal, you could see what command to Haddock Cabal is issuing and manually re run that, but with `--show-all` appended to the very end of the command.

cartazio:
Ok.

I assume I can’t have my cabal config set to add this globally to cabal
haddock  invocations until the ghc and associated  haddock pair lands? Or
when? :)


Thx!!!

cartazio:
Seriously thx.  This has been like the most pleasant bug/feature request
experience I’ve had with haddock to date


(Grumbles about how Data fam constructor docs are pinched to the right in
crazy ways)

harpocrates:
Closed in 5141e4b76af8462e49abdf48e44bb9cddb183383.

PR #752, opened 2018-02-09
Add 'show' option to complement 'hide'

### LINKED ISSUE #751: expose haddock flag to suppress/ignore haddock hide module attribute
for more sophisticated/power user developers who understand they're possibly getting into gritty internals, having the internal modules actually have fully rendered haddocks is super handy.

this is currently somehow possible via cabal config file at ~/.cabal/config, but not through haddock CLI as far as I can tell...

harpocrates:
The behaviour is for flags passed in the command line to override flags in file headers. In the command line, later flags override earlier ones.

Fixes #751 and #266.


cartazio:
1) how is cabal currently emulating the show all flag at its layer? (I guess that doesn’t matter for you )

2) for super Giant prjects , would it make sense to allow globs in the module names ?  Doesn’t matter for me, but thinking out loud ;)

PR #761, opened 2018-02-21
Run builds concurrently

JonCoens:
Haddock builds run sequentially, unlike GhcMake which runs builds of independent modules concurrently. This changes Haddock to process modules concurrently using a similar technique that GhcMake uses itself.
I'm not convinced this is the best way to do things and worry I'm missing nuances in places.

I've also added a test that exercises the difference in performance.
I don't think this is a great test to keep around. Putting a `threadDelay` inside `processModule` demonstrated a large performance difference.

I've tested this on a centos machine as well as OSX

I'm open to any suggestions.


haddock-api/src/Haddock/Interface.hs
```diff
@@ -162,17 +170,52 @@ createIfaces0 verbosity modules flags instIfaceMap =
...
+                   Nothing -> liftIO getNumProcessors
+                   Just n -> return n
+
+  par_sem <- liftIO $ newQSem n_jobs
+
+  let
+    withSem sem = bracket_ (waitQSem sem) (signalQSem sem)
+
+    spawnWorkers s = forM graph $ \(_mn, (m, mvar)) ->
+      forkIOWithUnmask $ \_unmask -> withSem par_sem $ do
```
simonmar: I think you want the `withSem` lower down, right before the `processModule`. Otherwise you can get a deadlock, where some modules have grabbed resources from the semaphore and are blocked on dependencies, but the compilation of the dependencies can't proceed because they need the semaphore.

alexbiehl:
Ah, maybe rebase was the wrong word, I meant it more like reopen against ghc-8.4 :)

JonCoens:
Yeah whoops, this is all messed up now. Will reopen a new request against 8.4

JonCoens:
I created https://github.com/haskell/haddock/pull/762

PR #762, opened 2018-02-22
Run builds concurrently

JonCoens:
Haddock builds run sequentially, unlike GhcMake which runs builds of independent modules concurrently. This changes Haddock to process modules concurrently using a similar technique that GhcMake uses itself.
I'm not convinced this is the best way to do things and worry I'm missing nuances in places.

I've also added a test that exercises the difference in performance.
I don't think this is a great test to keep around. Putting a `threadDelay` inside processModule demonstrated a large performance difference.

I'm open to any suggestions.

I was unable to thoroughly test this on the ghc-8.4 branch as cabal doesn't feel like resolving constraints.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 /hypsrc-test/out/
 /latex-test/out/
 /hoogle-test/out/
+/perf-test/out/
 
 /doc/haddock
 /doc/haddock.ps
--- a/haddock-api/src/Haddock/Interface.hs
+++ b/haddock-api/src/Haddock/Interface.hs
@@ -42,19 +42,27 @@ import Haddock.Options hiding (verbosity)
 import Haddock.Types
 import Haddock.Utils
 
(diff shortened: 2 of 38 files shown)


alexbiehl:
Alright, thank you. Looks good so far. No worries about the test suite. Once GHC-8.4 is out of the door I will take care fixing It properly.

JonCoens:
Using this patch to build all GHC documentation tickled a bug where boot file modules aren't handled appropriately. I'm looking at fixing that, but it currently looks to be a much larger change than this diff looked to be.

alexbiehl:
Jon, how are things going? 

I have finally managed to build the testsuite with ghc-8.4.1 and made it work again. Please rebase your branch so we can have travis chew on it when you are ready.

JonCoens:
I underestimated how much knowledge is required to get module loops building appropriately and concurrent. It's not as simple as forking threads and running them behind a semaphore. More work (and probably a different concurrent approach) will be needed here.

Issue #763, opened 2018-02-25
FIle handle leak


bgamari:
The current commit pointed to by GHC's `haddock` submodule crashes while building documentation for the `ghc` library on OS X,

```
compiler/deSugar/PmExpr.hs:33:1: error:
    Bad interface file: /Users/bgamari/bin-dist-8.4.0.20180224-Darwin/ghc/libraries/transformers/dist-install/build/Control/Monad/Trans/State/Lazy.dyn_hi
        /Users/bgamari/bin-dist-8.4.0.20180224-Darwin/ghc/libraries/transformers/dist-install/build/Control/Monad/Trans/State/Lazy.dyn_hi: openBinaryFile: resource exhausted (Too many open files)
```
I believe `--hyperlinked-sources` is the culprit as the issue vanishes if it is disabled.

alexbiehl:
@harpocrates, might you have a look? I think you touched the lines around 'mkTokenizedSrc' recently. I think we had an 'evaluate' there before.

hvr:
@harpocrates while at it, @bgamari also ran into the issue that `readFile` in

https://github.com/haskell/haddock/blob/3a34ce56935acdd86ea856b6f393be18516fe452/haddock-api/src/Haddock/Interface/Create.hs#L1115-L1122

is locale-dependent, i.e. if you `LANG=C haddock --hyperlinked-source SomeModuleWithUtf8.hs` you'll run into a decoding error.

alexbiehl:
Aw, we had this already. I must have missed that in the hyperlinker PRs.

Herbert Valerio Riedel <notifications@github.com> schrieb am So., 25. Feb.
2018, 09:12:

harpocrates:
I've opened a PR for this that should fix the problem. Sorry for the regression!

PR #764, opened 2018-02-25
Fix file handle leak (#763)

harpocrates:
Brought back some mistakenly deleted code for handling encoding and eager
reading of files from e0ada1743cb722d2f82498a95b201f3ffb303137.


alexbiehl:
Alec, thanks! Looks good to me.

hvr:
@alexbiehl what's up with the test-suite failures?

alexbiehl:
I will take care of fixing the test suite Next week. With one or two exceptions it is a matter if just accepting the new output.

Issue #781, opened 2018-03-20
Quickjump menu should be reset/emptied upon close.


merijn:
If I type 's' to open the quickjump menu, type some text, then hit ESC to close the text box, then the text I typed is still there when I hit 's' to open it again. This should probably reset after closing and/or have a shortcut for doing so.

alexbiehl:
I agree this hits me too. Care to make a PR?

merijn:
I'll put it on my yak-shaving queue, but I have a ton of higher priority yaks to get to first >.>

alexbiehl:
I am pinging @tfausak here, maybe he can put this issue on his "Call for participation" section. I can see more calls for QuickJump fixes coming.

tfausak:
Sorry, I didn't see this in time for this week's issue! I'll be sure to include it next week: https://github.com/haskellweekly/haskellweekly.github.io/issues/173

Issue #783, opened 2018-03-21
Don't show 'QuickJump' button in header if not present


alexbiehl:
We always show the QJ button even if the package wasn't generated with `--quickjump` flag. This is confusing!

![bildschirmfoto 2018-03-21 um 11 06 25](https://user-images.githubusercontent.com/1876617/37704033-f4695b24-2cf7-11e8-932b-93178fe7cdfd.png)

PR #789, opened 2018-03-30
Clear search string on quickjump hide for #781

jhenahan:
Not sure if it's too aggressive to reset the search string every time the quickjump is hidden, so I'm happy to change it if it's too much.


alexbiehl:
Thank you @jhenahan, looks good to me! @merijn you opened the issue, do you have an opinion about this?

alexbiehl:
Jack, maybe https://github.com/haskell/haddock/issues/783 is also interesting for you? It will involve a small haskell patch this time.

merijn:
Looks sensible enough to me.

jhenahan:
@alexbiehl I was looking into that one, as well. I have a couple of ideas about it, so I’ll try to get a patch together if no one beats me to it. :)

alexbiehl:
Nice, looking forward to it! Merging this one for now.

PR #813, opened 2018-05-05
Filter out CRLFs in hyperlinker backend

### LINKED ISSUE #812: Windows line endings (CRLF) affect HTML source code rendering?
On Hackage, Windows line endings (CRLF) affect the rendering of source code. See, for example, https://hackage.haskell.org/package/ansi-terminal-0.8.0.3/docs/src/System.Console.ANSI.html, which renders with a blank line between each line of source code. The hackage-server GitHub repository has identified this as a Haddock-related issue to be raised here (https://github.com/haskell/hackage-server/issues/750).

This sort of issue may have arisen in the past: see #57.

harpocrates:
This prevents spurious lines from appearing in the final output.

This appears to fix #812.


alexbiehl:
Looks good to me. Thanks Alec!

PR #818, opened 2018-05-08
Remove 'TokenGroup' from Hyperlinker

harpocrates:
Since the hyperlinker backend now relies on the GHC tokenizer, something like `Bar.Baz.foo` already gets bunched together into one token (as opposed to being spread across `Bar`, `.`, `Baz`, `.`, and `foo`).


harpocrates:
This should not change the hyperlinker output in any way. That was the case when I tried this on the`lens` package.

PR #821, opened 2018-05-09
Use the response file utilities defined in `base`

ckoparkar:
Summary: The response file related modules were recently copied from
`haddock` into `base`. This patch removes them from `haddock`.

GHC Trac Issues: [#13896](https://ghc.haskell.org/trac/ghc/ticket/13896)

/cc @bgamari

[Phab:4612](https://phabricator.haskell.org/D4612)


RyanGlScott:
Thanks, @ckoparkar!

PR #823, opened 2018-05-10
Hide instances by default

### LINKED ISSUE #698: Hide instances by default
Even browsing through the Prelude is a hassle, among many other examples, because of the hundreds of instances that come with each type class.

Lysxia:
Fixes #698.

### DIFF
--- a/haddock-api/src/Haddock/Backends/Xhtml/Layout.hs
+++ b/haddock-api/src/Haddock/Backends/Xhtml/Layout.hs
@@ -203,7 +203,7 @@ subInstances :: Maybe Package -> Qualification
              -> [(SubDecl, Maybe Module, Located DocName)] -> Html
 subInstances pkg qual nm lnks splice = maybe noHtml wrap . instTable
   where
-    wrap contents = subSection (hdr +++ collapseDetails id_ DetailsOpen (summary +++ contents))
+    wrap contents = subSection (hdr +++ collapseDetails id_ DetailsClosed (summary +++ contents))
     instTable = subTableSrc pkg qual lnks splice
     subSection = thediv ! [theclass "subs instances"]
     hdr = h4 ! collapseControl id_ "instances" << "Instances"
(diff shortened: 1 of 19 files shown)


sjakobi:
I would like to have this.

@alexbiehl: Can this get merged?

Lysxia:
Ping @harpocrates. This is ready for review.

harpocrates:
@Lysxia Thanks for rebasing! I haven't merged this sooner because I'm a bit wary of pushing through stylistic changes which aren't obvious improvements (especially after the new theme). In other words: it isn't clear to me that this is a change that most people will want. If you could encourage folks to give feedback either in a mailing list, on IRC, on Reddit, or simply as 👍/👎 on the PR, I'll happily merge.

Personally, I'm neither a 👍 nor a 👎.

Again, I'm really sorry for letting this live in a limbo state for so long. 😞

Lysxia:
I see, that makes sense. I'll start a thread on haskell-café to get feedback.

ntc2:
**I'm an enthusiastic :+1: on collapsing instances by default.**

**For example, the `Prelude` Haddocks are 98 screens tall with instances expanded [1], vs 40 screens with instances collapsed [2].** I.e. the `Prelude` documentation is 60% instances! When looking at documentation I usually don't care about the instances until I've zeroed in on a specific data type. Similar with examples, where I'm happy they are collapsed by default.

[1] prelude-expanded-98-screens
[2] prelude-collapsed-40-screens

david-christiansen:
It seems to me that this might make it more difficult to find information by searching in the page, given that Ctrl-F will typically skip hidden content. In the past, I've used Ctrl-F to get an overview of instances and datatypes when trying to figure out a library.

What about adding an "expand all" button, to continue supporting this use case?

Lysxia:
Thanks everyone for the feedback! After more thinking and discussion on Haskell-café, it seems too radical to switch the default globally. Or at the very least, more experimentation is necessary. So I will close this PR for now and continue discussion on #698.

Issue #834, opened 2018-05-22
Some parentheses cause MINIMAL pragma to be excluded from Haddock output


taktoa:
With Haddock 2.18.1, the following code (excerpted from [here](https://github.com/taktoa/eqsat/blob/547fdc9b8b7cc45e0ab407c02f5ad4a4d8716e2f/library/EqSat/TermIndex/Class.hs#L483-L553)):

```haskell
class (TermIndex index) => Mergeable index where
  {-# MINIMAL (merge | mergeMany) #-}

  merge
    :: (Monad m, Key node var)
    => index node var value
    -> index node var value
    -> (value -> value -> m value)
    -> m (index node var value)
  merge left right comb = do
    mergeMany (BV.fromList [left, right]) comb

  mergeMany
    :: ( Monad m, Key node var
       , Vector vec (index node var value)
       , Vector vec (Mut index node var value (PrimState m))
       )
    => vec (index node var value)
    -> (value -> value -> m value)
    -> m (index node var value)
  mergeMany indices comb = do
    Vector.foldr (\x my -> my >>= \y -> merge x y comb) new indices

  mergeMut
    :: (PrimMonad m, Key node var)
    => Mut index node var value (PrimState m)
    -> Mut index node var value (PrimState m)
    -> (value -> value -> m value)
    -> m (Mut index node var value (PrimState m))
  mergeMut leftMut rightMut comb = do
    left  <- freeze leftMut
    right <- freeze rightMut
    merge left right comb >>= unsafeThaw

  mergeManyMut
    :: ( PrimMonad m, Key node var
       , Vector vec (index node var value)
       , Vector vec (Mut index node var value (PrimState m))
       )
    => vec (Mut index node var value (PrimState m))
    -> (value -> value -> m value)
    -> m (Mut index node var value (PrimState m))
  mergeManyMut indicesMut comb = do
    indices <- Vector.mapM freeze indicesMut
    mergeMany indices comb >>= unsafeThaw
```

results in this output:

![selection_205](https://user-images.githubusercontent.com/553443/40360565-30f2f8a6-5d8c-11e8-8bba-e43f1f91f16f.png)

If I remove the parentheses from the `MINIMAL` pragma (i.e.: `{-# MINIMAL merge | mergeMany #-}`), I get this output:

![selection_206](https://user-images.githubusercontent.com/553443/40360573-35084c0c-5d8c-11e8-9408-4a620b834b4c.png)

[...]

PR #835, opened 2018-05-22
Use `ClassOpSig` instead of `TypeSig` for class methods

harpocrates:
From #834:

```haskell
module Minimal (C(..)) where 

-- | A  class
class C index where
  {-# MINIMAL (merge | mergeMany) #-}

  -- | One 
  merge :: index -> index -> index
  merge x1 x2 = mergeMany [x1,x2]

  -- | Two
  mergeMany :: [index] -> index
  mergeMany = foldr1 merge 
```

Looked like

screen shot 2018-05-22 at 9 43 42 am

And now looks like

screen shot 2018-05-22 at 9 43 47 am

I was going to add a test case, but it looks like the test suite is already full of test cases that were broken: minimal pragmas that weren't supposed to show up but did, minimal pragmas that were supposed to show up but didn't, class methods that weren't should up in the synopsis, you name it!


alexbiehl:
Looks good! Thanks Alec!

PR #839, opened 2018-05-24
Extract docs from strict/unpacked constructor args

### LINKED ISSUE #836: haddocks incorrectly parses UNPACK pragma
This might be a problem with cabal new-haddock, but given that it fails in a parsing stage, I think it's haddock causing the problem. Let me know if you think I should open up the issue on cabal as well/instead.

```bash
#GHC version
$ ghc --version
The Glorious Glasgow Haskell Compilation System, version 8.4.2

#cabal version
$ cabal --version
cabal-install version 2.2.0.0
compiled using version 2.2.0.1 of the Cabal library

#haddock version
$ haddock --version
Haddock version 2.20.0, (c) Simon Marlow 2006

#Operating System
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.4 LTS
Release:	16.04
Codename:	xenial

#Replicate
$ git clone git@github.com:chessai/freq && cd freq
freq$ nix-shell
nix-shell$ cabal new-build #this will succeed
nix-shell$ rm -rf dist*
nix-shell$ cabal new-haddock #this will fail with the following:
Build profile: -w ghc-8.4.2 -O1
In order, the following will be built (use -v for more details):
 - freq-0.1.0.0 (lib) (first run)
Preprocessing library for freq-0.1.0.0..
Running Haddock on library for freq-0.1.0.0..
Haddock coverage:

src/Freq/Internal.hs:208:3: error:
    parse error on input ‘{-# UNPACK’
    |
208 |   {-# UNPACK #-} !ByteArray -- ^ Square two-dimensional array of Double, maps first char and second char to probability
    |   ^^^^^^^^^^
cabal: Failed to build documentation for freq-0.1.0.0.

```

Most of the above was inside of a nix-shell.
I tried to create a minimal example of just a module with a similarly laid-out (textually) data type, like so:

```haskell
{-# LANGUAGE BangPatterns #-}

module Example
  ( -- * The Foo type
    Foo(..)
  ) where

data Foo = Foo
  {-# UNPACK #-} !Int
  {-# UNPACK #-} ![Int]

```

but haddock parses this correctly, under the same setup, inside of the same nix shell.

harpocrates:
This fixes #836 by matching the GHC changes in https://phabricator.haskell.org/D4727.

The GHC diff fixes the parse error around strictness annotations by moving the `HsDocTy` into the `HsBangTy` (if there is one). This patch correspondingly teaches Haddock to look into `HsBangTy` for docstrings.

### DIFF
--- a/haddock-api/src/Haddock/Interface/Create.hs
+++ b/haddock-api/src/Haddock/Interface/Create.hs
@@ -472,6 +472,7 @@ conArgDocs con = case getConArgs con of
                    RecCon _ -> go 1 ret
   where
     go n (HsDocTy _ _ (L _ ds) : tys) = M.insert n ds $ go (n+1) tys
+    go n (HsBangTy _ _ (L _ (HsDocTy _ _ (L _ ds))) : tys) = M.insert n ds $ go (n+1) tys
     go n (_ : tys) = go (n+1) tys
     go _ [] = M.empty
 


alexbiehl:
Sorry for my bad repo management lately, I have a new job and an constantly under fire.

PR #846, opened 2018-06-01
Improve Hyperlinker's 'spanToNewline'

### LINKED ISSUE #837: haddock segfaults
Summary: Haddock segfaults when I do 'stack haddock' on a new nightly-2018-05-22 project that has haskell-src-exts as a dependency.

Originally, I made an issue here: https://github.com/commercialhaskell/stack/issues/4036

Unfortunately, being unfamiliar with the intricacies of stack and cabal, I don't have a smaller test case than the following:

1. Start with a fresh installation (official stack 1.7.1 binary installed, no ~/.stack/ directory)
2. `stack new --resolver=nightly-2018-05-22 test`
3. `cd test`
4. Edit package.yaml, adding haskell-src-exts to dependencies.
5. `stack haddock` and witness the crash.

It's clear that haddock segfaulted, not merely exited with a regular failure code.

harpocrates:
`spanToNewline` is used to help break apart the source into lines which
can then be partioned into CPP and non-CPP chunks. It is important that
`spanToNewline` not break apart tokens, so it needs to properly handle
things like

  * block comments, possibly nested
  * string literals, possibly multi-line
  * CPP macros, possibly multi-line

Comments in string literals in particular were not being properly handled.
The fix is to to fall back on `Text.Read.lex` to help lex things that are not
comments.

Fixes #837.

### DIFF
--- a/haddock-api/src/Haddock/Backends/Hyperlinker/Parser.hs
+++ b/haddock-api/src/Haddock/Backends/Hyperlinker/Parser.hs
@@ -3,6 +3,8 @@ module Haddock.Backends.Hyperlinker.Parser (parse) where
 import Data.Either         ( isRight, isLeft )
 import Data.List           ( foldl', isPrefixOf, isSuffixOf )
 import Data.Maybe          ( maybeToList )
+import Data.Char           ( isSpace )
+import qualified Text.Read as R
 
 import GHC                 ( DynFlags, addSourceToTokens )
 import SrcLoc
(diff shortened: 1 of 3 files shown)


alexbiehl:
Wow, tricky. And what consequences! Thanks for the patch!

harpocrates:
This brittleness is a bit frustrating - our lexer should not segfault, it should produce ugly output (and hopefully get back on its feet eventually). The segfaulting behaviour should have been fixed by https://phabricator.haskell.org/D4336, but it looks like there is one other place where GHC's lexer's internal position is modified: old-style line pragmas:

```
# 17 "/nix/store/p85kjy91dfvs4in358zyfxlksvibw0zn-glibc-2.26-131-dev/include/stdc-predef.h" 3 4
```

Do you think it is worth shoehorning these into `ITline_prag` for GHC's `lexTokenStream` (the same way `{-# LINE ... #-}` also get put into `ITline_prag`)?

alexbiehl:
If that makes things more stable then I am all for it!

alexbiehl:
Merging this for now. We can iterate once you adjusted GHC.

Profpatsch:
@alexbiehl Can this fix be backported to `8.2`? Haddock segfaults often on current nixpkgs.

PR #867, opened 2018-07-01
Preserve docs on type family instances

### LINKED ISSUE #726: Support haddock for type family instances
Currently there's no way to add documentation to type families instances. Maybe I just haven't find a way to do this. But it would be good to have clear way to specify docs for type family instances.

harpocrates:
The only problem was that the instance location was slightly off
for type family instances. This fixes #726.

Example:

```
{-# LANGUAGE TypeFamilies #-}
module Lib where

data Bar = Bar

-- | A 'Foo'
type family Foo a :: *

-- | 'Foo' of 'Bar'
type instance Foo Bar = Bar

-- | 'Eq' of 'Bar'
instance Eq Bar where _ == _ = True
```

now produces

screen shot 2018-07-01 at 9 31 35 am

### DIFF
--- a/haddock-api/src/Haddock/Interface/Create.hs
+++ b/haddock-api/src/Haddock/Interface/Create.hs
@@ -419,9 +419,12 @@ mkMaps dflags pkgName gre instances decls = do
     instanceMap = M.fromList [ (getSrcSpan n, n) | n <- instances ]
 
     names :: SrcSpan -> HsDecl GhcRn -> [Name]
-    names l (InstD d) = maybeToList (M.lookup loc instanceMap) -- See note [2].
+    names _ (InstD d) = maybeToList (M.lookup loc instanceMap) -- See note [2].
       where loc = case d of
-              TyFamInstD _ -> l -- The CoAx's loc is the whole line, but only for TFs
+              -- The CoAx's loc is the whole line, but only for TFs. The
+              -- workaround is to dig into the family instance declaration and
+              -- get the identifier with the right location.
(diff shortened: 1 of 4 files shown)


harpocrates:
No need to add test cases - there are already some there (whose output I've accepted).

rampion:
This doesn't seem to work for closed type families.

I tried to use `-- ^` to document individual instances

```haskell
-- | Shape of the preimage of a function; how
--   many elements in the domain can map to a single
--   element in the codomain
type Preimage :: Injectivity -> Surjectivity -> Type -> Type
type family Preimage i s a where
  Preimage 'Noninjective 'Nonsurjective a = [a]       -- ^ general case: no limit on the size of the preimage
  Preimage 'Noninjective 'Surjective a = NonEmpty a   -- ^ surjective : ∀ b, ∃ at least one a s.t f a = b
  Preimage 'Injective 'Nonsurjective a = Maybe a      -- ^ injective : ∀ b, ∃ at most one a s.t. f a = b
  Preimage 'Injective 'Surjective a = a               -- ^ bijective : ∀ b, ∃ exactly one a s.t. f a = b
```

But got complaints when building the documentation. 

```
$ cabal haddock
...
src/Pear/Function.hs:65:55: error:
    parse error on input ‘-- ^ general case: no limit on the size of the preimage’
   |
65 |   Preimage 'Noninjective 'Nonsurjective a = [a]       -- ^ general case: no limit on the size of the preimage
   |                                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
cabal: Failed to build documentation for pear-1.0.0.0.
```

Issue #806, opened 2018-04-24
Hoogle backend oddities with type families


RyanGlScott:
Take this file:

```haskell
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE UndecidableInstances #-}
module Bug where

import Data.Proxy

-- | 'F' docs
type family F
-- | 'D' docs
data family D

v :: Int
v = 42

-- | 'C' docs
class C a where
  -- | 'AT' docs
  type AT a
  type AT a = Proxy (Proxy (Proxy (Proxy (Proxy (Proxy (Proxy (Proxy (Proxy (Proxy)))))))))
```

And run `haddock --hoogle --package-name=bug --package-version=0.0.0 Bug.hs`, you'll get the following `bug.txt` file:

```
-- Hoogle documentation, generated by Haddock
-- See Hoogle, http://www.haskell.org/hoogle/

@package bug
@version 0.0.0

module Bug

-- | <a>F</a> docs

-- | <a>D</a> docs
v :: Int

-- | <a>C</a> docs
class C a where {
    type family AT a;
    type AT a =
        Proxy (Proxy (Proxy (Proxy (Proxy (Proxy (Proxy (Proxy (Proxy (Proxy)))))))));
}
```

There are many things wrong here:

1. Neither `F` nor `D` show up at all! This causes `v` to have a completely incorrect docstring associated with it.
2. `AT` _does_ show up, but it's missing its Haddocks.
3. The default definition for `AT` spans multiple lines. According to @ndmitchell (in https://github.com/ndmitchell/hoogle/issues/249#issuecomment-383954166), this should never happen, as Hoogle is line-orientated.

RyanGlScott:
For the life of me, I can't figure out how to fix part (3). But parts (1) and (2) are straightforward enough—see https://github.com/haskell/haddock/pull/808.

alexbiehl:
3) Seems to be related to pretty printing, right? Shouldn't we be able to influence line breaking there?

RyanGlScott:
(3) was fixed in #868.

PR #868, opened 2018-07-01
Avoid line breaks due to line length in Hoogle

harpocrates:
Hoogle operates in a line-oriented fashion, so we should avoid ever
breaking due to long lines.

One way of doing this non-intrusively is to modify the `DynFlags` that
are threaded through the `Hoogle` module (note this is anyways only
passed through for use in the various `showSDoc` functions).

This fixes the (3) mentioned in #806.


RyanGlScott:
Very good catch, @harpocrates. LGTM!

Issue #873, opened 2018-07-05
Hoogle backend removes parentheses erroneously for functions reexported from other packages


RyanGlScott:
Take this module:

```haskell
module Bug (($)) where
```

And run `haddock --hoogle --package-name=bug --package-version=0.0.0 Bug.hs`. You'll get some very questionable results in `bug.txt`:

```
-- Hoogle documentation, generated by Haddock
-- See Hoogle, http://www.haskell.org/hoogle/

@package bug
@version 0.0.0

module Bug
($) :: () => a -> b -> a -> b
```

The type of `($)` is quite wrong, since it's missing some critical parentheses.

As far as I can tell, this issue only occurs with functions that were defined in a separate package. If I try to define `($)` locally, or within a different module in the same package, then the issue does not occur.

harpocrates:
The HTML and LaTeX backends each have a section to deal with properly printing types (see [this](https://github.com/haskell/haddock/blob/ghc-head/haddock-api/src/Haddock/Backends/Xhtml/Decl.hs#L1101) and  [this](https://github.com/haskell/haddock/blob/ghc-head/haddock-api/src/Haddock/Backends/LaTeX.hs#L904)). The Hoogle backend just uses the GHC `Outputable` instance.

I'll have a fix soon.

PR #874, opened 2018-07-06
Refactor handling of parens in types

harpocrates:
Right now, there is duplicated logic in the XHTML and LaTeX backends for adding parenthesis where necessary. As #873 shows, we need that logic in the Hoogle backend too!

This PR adds to `GhcUtils` a `reparenType :: HsType a -> HsType a` function that add parens where necessary. The backends no longer need to worry about precedence - they print parens only when they see `HsParTy`. The only downside I see to this approach is that it does mean we traverse types twice (once to "reparen" them and again to print them) instead of once.


harpocrates:
This could all go on the `ghc-8.4` branch, but it touches a lot of `HsType` constructors. Rebasing that on top of the TTG stuff would not be fun.

hoogle-test/ref/type-sigs/test.txt
```diff
@@ -6,11 +6,11 @@
 
 module ReaderT
 newtype ReaderT r m a
-ReaderT :: r -> m a -> ReaderT r m a
+ReaderT :: (r -> m a) -> ReaderT r m a
```
RyanGlScott: Yikes, I can't believe I didn't notice this before—we even checked in this junk!

hoogle-test/ref/Bug873/test.txt
```diff
@@ -0,0 +1,26 @@
...
+--   <pre>
+--   f $ g $ h x  =  f (g (h x))
+--   </pre>
+--   
+--   It is also useful in higher-order situations, such as <tt><a>map</a>
+--   (<a>$</a> 0) xs</tt>, or <tt><a>zipWith</a> (<a>$</a>) fs xs</tt>.
+--   
+--   Note that <tt>($)</tt> is levity-polymorphic in its result type, so
+--   that foo $ True where foo :: Bool -&gt; Int# is well-typed
+($) :: () => (a -> b) -> a -> b
```
RyanGlScott: Darn, I was faintly hoping that fixing this issue would also fix this strange empty context that appears here. Oh well, that can wait for another day.

haddock-api/src/Haddock/GhcUtils.hs
```diff
@@ -226,6 +228,95 @@ getGADTConTypeG (ConDeclH98 {}) = panic "getGADTConTypeG"
   -- Should only be called on ConDeclGADT
 getGADTConTypeG (XConDecl {}) = panic "getGADTConTypeG"
 
+
+-------------------------------------------------------------------------------
+-- * Parenthesization 
+-------------------------------------------------------------------------------
+
+-- | Precedence level (inside the 'HsType' AST).
+data Precedence
```
RyanGlScott: Perhaps we could reuse the [`PprPrec` data type](http://git.haskell.org/ghc.git/blob/45f0026818402aa08398131507bc587ea4a2b387:/compiler/basicTypes/BasicTypes.hs#l674) from GHC itself?

haddock-api/src/Haddock/GhcUtils.hs
```diff
@@ -226,6 +228,95 @@ getGADTConTypeG (ConDeclH98 {}) = panic "getGADTConTypeG"
...
+  go p (HsQualTy x ctxt ty)
+    = paren p PREC_FUN $ HsQualTy x (fmap (map reparenLType) ctxt) (reparenLType ty)
+  go p (HsFunTy x ty1 ty2)
+    = paren p PREC_FUN $ HsFunTy x (goL PREC_FUN ty1) (goL PREC_TOP ty2)
+  go p (HsAppTy x fun_ty arg_ty)
+    = paren p PREC_CON $ HsAppTy x (goL PREC_FUN fun_ty) (goL PREC_CON arg_ty)
+  go p (HsOpTy x ty1 op ty2)
+    = paren p PREC_FUN $ HsOpTy x (goL PREC_OP ty1) op (goL PREC_OP ty2)
+  go p (HsParTy _ t) = unLoc $ goL p t -- pretend the paren doesn't exist - it will be added back if needed
+  go _ t = t
```
RyanGlScott: I'm generally not a fan of using catch-call cases like this, since it will make it hard to remember to change this definition later when more constructors are to `HsType`. Let's list out each constructor explicitly.

haddock-api/src/Haddock/GhcUtils.hs
```diff
@@ -226,6 +228,95 @@ getGADTConTypeG (ConDeclH98 {}) = panic "getGADTConTypeG"
...
+
+-- | Add parenthesis around the types in a 'LHsType' (see 'reparenTypePrec')
+reparenLType :: (XParTy a ~ NoExt) => LHsType a -> LHsType a
+reparenLType = fmap reparenType
+
+-- | Add parenthesis around the types in a 'HsTyVarBndr' (see 'reparenTypePrec')
+reparenTyVar :: (XParTy a ~ NoExt) => HsTyVarBndr a -> HsTyVarBndr a
+reparenTyVar (UserTyVar x n) = UserTyVar x n
+reparenTyVar (KindedTyVar x n kind) = KindedTyVar x n (reparenLType kind)
+reparenTyVar v = v
```
RyanGlScott: See my comments above about catch-all cases.

haddock-api/src/Haddock/GhcUtils.hs
```diff
@@ -226,6 +228,95 @@ getGADTConTypeG (ConDeclH98 {}) = panic "getGADTConTypeG"
...
+
+-- | Add parenthesis around the types in a 'ConDeclField' (see 'reparenTypePrec')
+reparenConDeclField :: (XParTy a ~ NoExt) => ConDeclField a -> ConDeclField a
+reparenConDeclField (ConDeclField x n t d) = ConDeclField x n (reparenLType t) d
+reparenConDeclField c = c
```
RyanGlScott: See my comments above about catch-all cases.

hoogle-test/ref/Bug873/test.txt
harpocrates: I was hoping for this too, but I didn't see any clean way to abstract that logic across backends too.

haddock-api/src/Haddock/GhcUtils.hs
harpocrates: GHC has `funPrec == opPrec`, while Haddock has `funPrec < opPrec`. The `Note [Type operator precedence]` summarizes this difference quite well: GHC's choice of precedence is not incorrect - it will just lead to a extra parens.

In summary I'd rather not, but I'm willing to be told otherwise.

haddock-api/src/Haddock/GhcUtils.hs
harpocrates: Good point indeed! I've pushed a change for this and the remaining three comments.

haddock-api/src/Haddock/GhcUtils.hs
RyanGlScott: Ah, I missed that detail. In that case, what you've used here is fine.

hoogle-test/ref/Bug873/test.txt
RyanGlScott: That makes sense. It might be worth opening a separate issue about this, then.

harpocrates:
I've now also added a commit that fixes #877. Although the bug is reported for the HTML backend, it obviously affects all backends.

PR #884, opened 2018-07-19
Add # as a special character

msakai:
'#' has special meaning used for anchors and can be escaped using backslash.
Therefore it would be nice to be listed as special characters.


alexbiehl:
Thanks!

Issue #255, opened 2014-05-08
Rework the test method to prevent failing tests due to interface file version changes.


ghc-mirror:
**Original reporter**: _fuuzetsu@<hidden>_

Bumping up the interface version during development to something incompatible with the interface file of the developer's GHC causes Haddock to be unable to link to basic things like String. While this is expected, the HTML test-suite fails because of it: links don't get generated, formatting changes (no line breaks for `<a>` tags) and the tests fail. 

Either stripping the file comparison has to be made smarter (strip links AND tags AND reformat the file) or tests somehow made fully independent of what's on the developer's system, including base documentation. 

First option is easier and most likely what will be done. Second option is hard but we don't lose the ability to do link comparisons. Currently link comparisons aren't done due to stripping anyway (which causes problems when trying to test something like ticket #253).

A quick and effective solution would be to simply not refer to any types defined in any external modules (those not in the testing directory).

A long term solution would be to strip + reformat for tests where we don't care about links AND improve the way testing is done so that we can do fully self-contained tests with links when we do care.

PR #890, opened 2018-07-20
Let `haddock-test` bypass interface version check

harpocrates:
Partially alleviates #255 by disabling the interface version check for `haddock-test`. This means `haddock-test` might

  * crash during deserialization
  * deserialize incorrectly

Still - it means things _might_ work where they were previously guaranteed not to. For example, this is enough to keep tests fully green when bumping the interface version in #875.


alexbiehl:
We have been bitten quite often lately. While I like the backward compat way I think it might not be worth it: a) In general, GHC's Binary framework doesn't support b) we are probably the only one to benefit. I am fine with this.

Issue #462, opened 2015-12-18
Haddock uses a lot of memory


mpickering:
I've been doing some profiling of haddock so this ticket is for me to write down what I have found out.

Here is the allocation graph when running haddock on `base`. 

![haddock](https://cloud.githubusercontent.com/assets/1216657/11886013/fa83e458-a51c-11e5-930f-51c98a9eae17.png)

The only observation I have at the moment is that it seems that the specialisation code leaks quite a lot of memory more than the previous version. 

It might be worth investigating the suggestion on the old ticket to write interfaces to file but I have no intuition as to how much space this will save.

mpickering:
More graphs, hy profile and with the hyperlinker enabled

https://usercontent.irccloud-cdn.com/file/l7vmLw7e/

https://usercontent.irccloud-cdn.com/file/rN8oni3d/hyperlinked

mpickering:
Here is the .prof. http://lpaste.net/147305

mpickering:
@mrhania Is there any chance you could look at this?

mrhania:
Yeah, so the code was wrote with simplicity in mind and I haven't put too much thought into the performance (and honestly, I haven't had much experience with performance tweaking in Haskell at the time). For example, for almost each token type I run a separate query on the Haskell AST which is just plain terrible.

I will rework this at some point, I am just not sure when as I am quite a busy person right now. But I will, that's for sure.

asr:
+1

andreasabel:
@mrhania When will the day come?

alexbiehl:
@andreasabel @asr have you tried building haddock documentation with haddock-2.18.1 (it is bundled with ghc-8.2.1). Much has been done in this area. Please report back if you do a before-after comparison.

Fuuzetsu:
@alexbiehl Out of interest, what was done more concretely? The biggest issue with Haddock has always been that it has to ask GHC to re-do a lot of work, load modules, typecheck, …. Were there improvements in that area? I always suspected that improvements to Haddock code itself would always provide only a minor benefit and its the use of GHC API would have to get smarter.

alexbiehl:
- Optimization of syb-traversal in the hyperlinker (see https://github.com/haskell/haddock/pull/621).
- AttachInstances makes use of `getNameToInstancesIndex` from recent GHCs (see https://github.com/haskell/haddock/pull/636)
[...]

PR #893, opened 2018-07-23
Accumulate explicitly which modules to load for 'attachInstances'

harpocrates:
The old approach to fixing #469, while correct, consumes a lot of
memory. We ended up with a HUGE `GblRdrEnv` in `ic_rn_gbl_env`. However,
`getNameToInstancesIndex` takes that environment and compresses it down
to a much smaller `ModuleSet`.

Now, we compute that `ModuleSet` explicitly as we process modules. That
way we can just tell `getNameToInstancesIndex` what modules to load
(instead of it trying to compute that information from the interactive
context).

### DIFF
--- a/haddock-api/src/Haddock/Interface.hs
+++ b/haddock-api/src/Haddock/Interface.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE CPP, OverloadedStrings #-}
+{-# LANGUAGE CPP, OverloadedStrings, BangPatterns #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  Haddock.Interface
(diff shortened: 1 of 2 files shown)


harpocrates:
Relevant issue: #462.

This won't build without GHC side changes to `getNameToInstancesIndex`: https://phabricator.haskell.org/D5003.

This reduces the peak memory usage of Haddock on the `ghc` library back down to ~1GB from > 2GB before. It basically fixes the memory regression in deddced31cabadf62fe01fff77b094cd005e52a1 (see [this comment for graphs](https://github.com/haskell/haddock/issues/462#issuecomment-406077792)). Here is what the new graph looks like:

screen shot 2018-07-23 at 2 00 25 pm

alexbiehl:
Now we have both! Precise instances and more efficient memory usage!

harpocrates:
Ping? Is there no hope of this making it into 8.6?

alexbiehl:
Hmmm. https://phabricator.haskell.org/D5003 hasn't been merged yet. Ping @bgamari

bgamari:
@alexbiehl in general I can't merge this sort of thing until the Haddock patch has merged. Afterall, otherwise I can't bump the submodule.

Issue #332, opened 2014-10-16
haddocks list data family instances that arent exported!


cartazio:
my minimal repro is this 
the resulting haddocks list ALLLL known Unboxed Vector/MVector data family instances and type class instances!  I'm not sure how much is at fault on the haddock or GHC side respectively.

``` haskell

{-# LANGUAGE TypeFamilies #-}
module M1(MyUnit(..),UV.Vector(V_MU),UV.MVector(MV_MU)) where


import Data.Vector.Unboxed as UV



--type instance VG.Mutable MyV = MyMV

data MyUnit = MU

newtype instance UV.MVector s MyUnit  = MV_MU ()
newtype instance UV.Vector    MyUnit = V_MU ()

```

![screen shot 2014-10-16 at 2 34 45 pm](https://cloud.githubusercontent.com/assets/4290/4668183/3aa4b3ba-5563-11e4-93d4-3e09beed3e14.png)
![screen shot 2014-10-16 at 2 34 23 pm](https://cloud.githubusercontent.com/assets/4290/4668184/3aa9b8ec-5563-11e4-90c5-87c7e2a53167.png)

cartazio:
It looks like theres a GHC ticket thats related to this https://ghc.haskell.org/trac/ghc/ticket/10027

Kleidukos:
Any development on that front? cc @cartazio @Fuuzetsu

Fuuzetsu:
I don't think so. Trying to push forward #897 is probably anyone's best bet but I have no idea to the status of the project currently, I'm not sure if there's really a maintainer.

PR #897, opened 2018-08-05
Better data family rendering

harpocrates:
This changes the HTML backend to render the constructors of data families in the same way as other constructors (from regular data declarations) are rendered. One pleasant side effect of this is that docs/fixities on data family constructors end up in the final HTML output.

See [this comment](https://github.com/haskell/haddock/issues/300#issuecomment-402349593) for a preview of what this looks like. I added that example as a test case in `html-test/src/DataFamily.hs`.


harpocrates:
Note to self: try to fix https://github.com/haskell/haddock/issues/332 before merging this.

Kleidukos:

PR #905, opened 2018-08-14
Load plugins when starting a GHC session

### LINKED ISSUE #900: Haddock fails to run plugins with GHC 8.6.1 alpha
GHC has changed so that plugins are only loaded once rather than just before each time they are used.
I suspect that Haddock needs to follow this change and explicitly loads the plugins as well. 

@christiaanb suggested some code like this in a suitable place would fix things.

https://github.com/clash-lang/clash-compiler/commit/8d32505b84a4f597bd995a4c39234b74be1b1e66#diff-c50e1570ba401340dde73490fed2c8cdR189

mpickering:
Fixes #900 

This also needs to be backported to the `ghc-8.6` branch.

### DIFF
--- a/haddock-api/src/Haddock.hs
+++ b/haddock-api/src/Haddock.hs
@@ -75,6 +75,7 @@ import Packages
 import Panic (handleGhcException)
 import Module
 import FastString
+import qualified DynamicLoading
 
 --------------------------------------------------------------------------------
 -- * Exception handling
@@ -448,7 +449,10 @@ withGhc' libDir flags ghcActs = runGhc (Just libDir) $ do
   -- that may need to be re-linked: Haddock doesn't do any
   -- dynamic or static linking at all!
   _ <- setSessionDynFlags dynflags''
-  ghcActs dynflags''
+  hscenv <- GHC.getSession
+  dynflags''' <- liftIO (DynamicLoading.initializePlugins hscenv dynflags'')
+  _ <- setSessionDynFlags dynflags'''
+  ghcActs dynflags'''
   where
 
     -- ignore sublists of flags that start with "+RTS" and end in "-RTS"


mpickering:
Is anything blocking this patch? @alexbiehl @harpocrates ?

alexbiehl:
Sorry, just me not noticing the notifications..

sheaf:
This fix doesn't seem to work on Windows. I'm afraid that even with verbose output, there are no meaningful error messages (I've attached the output anyway).
(Note that I had to patch GHC 8.6.1 for plugins to work on Windows, see [GHC trac #15700](https://ghc.haskell.org/trac/ghc/ticket/15700).)

[cabal new-haddock output, with GHC 8.6.1](https://github.com/haskell/haddock/files/2501772/cabal_haddock.txt)

harpocrates:
> This fix doesn't seem to work on Windows. I'm afraid that even with verbose output, there are no meaningful error messages (I've attached the output anyway).

Yeah, the `-v3` is something that cabal doesn't end up threading to Haddock (perhaps it should...). However, I doubt that the verbose output is all that useful - the failure is surely happening in `typecheckModule` 😄.

Unfortunately I don't have a Windows machine with which to debug...

sheaf:
At first, I thought it boiled down to the fact that, when running haddock, the call to initializePlugins (from `ghc/compiler/main/DynamicLoading.hs`) determines that no loading needs to take place.
When running e.g. `cabal new-repl` or `cabal new-build`, the call to `initializePlugins` would reload plugins, whereas with `cabal new-haddock`, it would follow the code path which did not reload anything.

However, after modifying the code to force cabal new-haddock to always reload plugins, the problem persists. Let me know what I should be looking for in `typecheckModule` and I'll try to keep digging.

PR #908, opened 2018-08-18
Remove unnecessary backslashes from docs

undergroundquizscene:
At https://haskell-haddock.readthedocs.io/en/latest/markup.html#special-characters
the backslash and backtick special characters show up with an extra
backslash before them – I think the escaping is not (or is no longer)
needed for those characters in rst.

I don’t know that much about rst, so shout if these changes are incorrect! I have built the documentation locally and verified that these changes remove the extra backslashes as I expected. I have also verified this on github (https://github.com/undergroundquizscene/haddock/blob/remove-backslashes-from-docs/doc/markup.rst).

### DIFF
--- a/doc/markup.rst
+++ b/doc/markup.rst
@@ -762,14 +762,14 @@ Special Characters
 ~~~~~~~~~~~~~~~~~~
 
 The following characters have special meanings in documentation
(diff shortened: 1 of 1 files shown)


alexbiehl:
Thanks!

PR #910, opened 2018-08-20
Dead code-paths in Hi Haddock

harpocrates:
Removing this code has no impact on the output (try accepting before and after the commit). Offhand, it doesn't seem to impact any of the docs for libraries bundled with GHC either. We should either:

  * add test cases exercising the code paths this PR removes (I can't think of any such cases)
  * add test cases which used to (before Hi Haddock) exercise these code paths and are now broken
  * merge this


harpocrates:
Thanks for the quick review! :D 

There is definitely some more global cleanup to do too, but I wanted to make sure this code was really supposed to be dead. I agree with getting rid of `Maps` in a separate PR.

PR #913, opened 2018-08-21
Convert GADTs properly

harpocrates:
This does two things:

  1. makes sure that `synifyTyCon` always produces GADTs from type constructors that originally came from GADTs
  2. removes foralls on GADT constructors


harpocrates:
The second of these points is somewhat arbitrary, but tends to  be much nicer. Look at the before and after:

screen shot 2018-08-21 at 9 11 18 am screen shot 2018-08-21 at 9 11 24 am

PR #914, opened 2018-08-21
Better rendering of unboxed sums/tuples

harpocrates:
* adds space after/before the '#' marks
* properly reify 'HsSumTy' in 'synifyType'


harpocrates:
This is going to matter much more once Hi Haddock gets merged, since `synifyType` is going to be used a lot.

RyanGlScott:
This looks plausible to me. There's no associated issue to go along with this, so I can't tell how they were rendered before, but presumably things are better now than they were previously.

You checked in some test `.hs` files, but no corresponding HTML files—was this intentional?

While you're in town, have you checked to see if partially applied tuple types display correctly?

harpocrates:
> This looks plausible to me. There's no associated issue to go along with this, so I can't tell how they were rendered before, but presumably things are better now than they were previously.

I'm working off of the WIP Hi Haddock branch which uses `synifyType` more widely. Here are some before after screenshots:

Before:
screen shot 2018-08-21 at 1 21 16 pm
After:
screen shot 2018-08-21 at 12 51 31 pm

Before:
screen shot 2018-08-21 at 1 19 32 pm
After:
screen shot 2018-08-21 at 12 54 00 pm

> You checked in some test .hs files, but no corresponding HTML files—was this intentional?

Right. Running tests only works from inside an inplace GHC - and I only have the Hi Haddock branch built right now.

> While you're in town, have you checked to see if partially applied tuple types display correctly?

Partially applied (and possible promoted) tuple types fall back to the prefix notation since they don't satisfy the length guard.

screen shot 2018-08-21 at 1 41 22 pm

PR #922, opened 2018-08-27
Properly synify promoted list types

harpocrates:
We reconstruct promoted list literals whenever possible. That means
that `synifyType` produces

    '[Int, Bool, ()]

instead of

    (Int ': (() ': (Bool ': ([] :: [Type]))))


harpocrates:
For the Hi-Haddock branch (which uses `synifyType` much more), this helps quite a bit:

Before:
screen shot 2018-08-27 at 2 29 00 pm

After:
screen shot 2018-08-27 at 2 29 16 pm

/cc @sjakobi

PR #931, opened 2018-09-04
More synify-ing improvements

harpocrates:
This cherry-picks from `ghc-head` some of the recent changes in `haddock-api/src/Haddock/Convert.hs`, then goes on to fix some more stuff on top of that:

  * filter almost all `forall`s on class methods and instance methods (visible improvements in `Bug548`, `Bug613`, and `SpuriousSuperclassConstraints`)
  * filter almost all `forall`s on pattern synonyms (visible improvements in `BundledPatterns` and `BundledPatterns2`)
  * fix a bug when computing the inferred type variable order (visible improvements in `Test`, but also all over the docs for the `Prelude`)
  * synify default methods (looks like handling these is now quite within reach since GHC internally gives these distinct names)
  * synify associated type defaults

### DIFF
--- a/haddock-api/src/Haddock/GhcUtils.hs
+++ b/haddock-api/src/Haddock/GhcUtils.hs
@@ -21,6 +21,7 @@ import Control.Arrow
 import Haddock.Types( DocNameI )
 
 import Exception
+import FV
 import Outputable
 import Name
 import NameSet
@@ -30,6 +31,9 @@ import HscTypes
 import GHC
 import Class
 import DynFlags
+import Var       ( TyVarBndr(..), TyVarBinder, tyVarKind )
+import VarSet    ( VarSet, emptyVarSet )
+import TyCoRep   ( Type(..) )
 
 import HsTypes (HsType(..))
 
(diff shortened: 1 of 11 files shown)


haddock-api/src/Haddock/GhcUtils.hs
```diff
@@ -436,3 +440,49 @@ setStubDir    f d = d{ stubDir    = Just f
 setOutputDir  f = setObjectDir f . setHiDir f . setStubDir f
 
 
+-------------------------------------------------------------------------------
+-- * Free variables of a 'Type'
+-------------------------------------------------------------------------------
+
+-- | Get free type variables in a 'Type' in their order of appearance.
+-- See [Ordering of implicit variables].
```
RyanGlScott: I must be missing some context here. Why would we ever want to retrieve free type variables in _reverse_ order?

harpocrates: We want them in order. Unfortunately, that's not what GHC does.

`unionFV` starts by running the second `FV` on the accumulator, then passing that to the first `FV`. That means it adds variables from the second argument to the `in_scope` set before adding variables from the first argument. All this is bad news if the first and second arguments share some variables.

Here is a snippet demonstrating the somewhat surprising behaviour:

```haskell
ghci -package ghc
ghci> import FV
ghci> import TysWiredIn
ghci> import Unique
ghci> import Name
ghci> import OccName
ghci> import Id
ghci> x0 = mkLocalId (mkSystemName (mkUniqueGrimily 0) (mkVarOcc "x0")) unitTy
ghci> x1 = mkLocalId (mkSystemName (mkUniqueGrimily 1) (mkVarOcc "x1")) unitTy
ghci> tvs = fvVarList (unitFV x0 `unionFV` unitFV x1 `unionFV` unitFV x0)
ghci> map (getOccString . idName) tvs
["x1","x0"]
```

My hack here is to traverse everything in reverse order, then reverse the list of type variables at the end.

RyanGlScott: Oh dear, that's absolutely terrible. I really thought that `tyCoVarsOfTypesWellScoped` would be enough to accomplish this task, but your counterexample shows that I was mistaken.

Still, I'm incredibly bothered by the fact that we have to duplicate so much machinery just to solve this problem. I've sent an [email to the ghc-devs list](https://mail.haskell.org/pipermail/ghc-devs/2018-September/016206.html) asking about this, so I'm hoping I can get some clarification on this point.

RyanGlScott: Based on Simon's response [here](https://mail.haskell.org/pipermail/ghc-devs/2018-September/016210.html), it looks like the `FV`-returning functions really don't guarantee any particular order (only that it's deterministic), which is a huge bummer.

Ideally, we'd rewrite `tyCoFVsOfType` and friends in such a way that they always return things in left-to-right order. Your hacked version accomplishes this, but at the cost of reversing the order at the very end, which feels wasteful. Is there a way to accomplish the same thing with only one pass?

harpocrates: > Is there a way to accomplish the same thing with only one pass?

In GHC: yes. We'd need to change `unitFV` to append to the var list. In order to keep that operation efficient, that means swapping occurrences of `[Var]` in `FV` for difference lists of `Var`. It looks like the `FV` code has already been carefully performance tuned, so we'd have to be very careful there.

In Haddock: not really, short of porting over most of `FV` (and then making the changes we'd otherwise make in GHC).

What frustrates me the most is not the `reverse`: it's having to rely on what is accidental behaviour in GHC.

RyanGlScott: Ugh, this is all so nasty.

Ultimately, I think the `orderedFVs` hack is probably the way to go for now. I'd only request that you expand the documentation for it to include an example of why `tyCoVarsOfTypeWellScoped` doesn't give the behavior you desire (e.g., your example in https://github.com/haskell/haddock/pull/931#discussion_r215013166).

Ideally, we'd investigate making the `FV` functions in GHC return things in a reliable left-to-right order. But as you've noted, they're quite performance intensive, so I can understand your reservations about making things slower by changing the underlying data structures involved.

harpocrates: I've added an example illustrating GHC's behaviour (I didn't know about `alphaTyVar` and `betaTyVar` - thanks!).

I agree we should eventually look into making `FV` return things in a reliable left-to-right order. I'm a bit swamped at the moment, but I'll try to circle back when all of this Hi Haddock stuff settles down.

RyanGlScott:
`FV` issues aside, this all looks plausible to me.

There's a lot of improvements mentioned in your description, but surprisingly few added tests. Do the existing tests cover this already? If not, it would be good to add some new ones.

harpocrates:
> There's a lot of improvements mentioned in your description, but surprisingly few added tests. Do the existing tests cover this already? If not, it would be good to add some new ones.

The existing tests cover the changes. In fact, pretty much all changes were motivated by the 20 currently failing HTML tests on the Hi Haddock branch. I'm trying to get the test outputs close enough that they can just be accepted.

As far as I'm concerned, the following test cases are ready to be accepted:

  * Bug294
  * Bug548 (after this PR)
  * Bug613 (after this PR)
  * Bug85
  * BundledPatterns (after this PR)
  * BundledPatterns2 (after this PR)
  * DeprecatedTypeFamily
  * FunArgs
  * Operators
  * PromotedTypes (after this PR)
  * SpuriousSuperclassConstraints (after this PR)

There are still 9 test cases that I don't consider ready to accept though. I've dumped out my TODO list below:

 * ConstructorArgs
      - strict non-record constructor args (see https://phabricator.haskell.org/D5067#inline-39368)
 * Instances
      - `data family Thud a c ` is leaking out to the top-level (instead of just being a plain associated data family)
 * PatternSyns
     - there is a stray `forall` on `pattern Blub`
 * Test
      - `r, s :: Int` fields get expanded out to two fields (maybe punt?)
      - contents of hidden module are not expanded out - only broken `module Hidden` link appears
 * Ticket112
      - Hyperlinking identifiers ending in `#` (see https://phabricator.haskell.org/D5067#inline-39282)
 * Ticket61
      - Contents of hidden module are not expanded out - only broken `module Ticket61_Hidden` link appears
 * TypeFamilies
      - missing docs on instances (see https://phabricator.haskell.org/D5067#anchor-inline-39369)
 * TypeFamilies2
      - missing docs on instances (see https://phabricator.haskell.org/D5067#anchor-inline-39369)
 * TypeOperators
     - try to print infix type operators that were declared infix (ie. ```data a `Op` b``` or ```newtype (g `O` f) a ...```)

RyanGlScott:
Ah, I missed that this was a part of the larger Hi Haddock project (in which case your goal is to un-break existing tests). I'll have to let @sjakobi review the Hi Haddock-specific bits, but otherwise, things here LGTM.

harpocrates:
@sjakobi Thanks for the review - a week's delay is hardly late!

PR #938, opened 2018-09-11
Default signatures

harpocrates:
Adds back rendering of default methods. While I was at it, I also cherry-picked some old un-merged changes to improve the rendering of: default methods & default associated types (in the XHTML backend) and default methods (in the LaTeX backend).

Note there is still some GHC-side work that needs to be done to collect docs on default methods and default associated types. However, this solves the more urgent problem that we currently aren't rendering _anything_ related to defaults.

### DIFF
--- a/haddock-api/src/Haddock/GhcUtils.hs
+++ b/haddock-api/src/Haddock/GhcUtils.hs
@@ -17,6 +17,7 @@
 module Haddock.GhcUtils where
 
 
+import Control.Applicative (liftA2)
 import Control.Arrow
 import Haddock.Types( DocNameI )
 
@@ -36,7 +37,9 @@ import VarSet    ( VarSet, emptyVarSet )
 import TyCoRep   ( Type(..) )
 
 import HsTypes (HsType(..))
+import Unique (deriveUnique, getKey)
 
+import Haddock.Types (SetName(..))
 
 moduleString :: Module -> String
 moduleString = moduleNameString . moduleName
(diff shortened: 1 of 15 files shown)


harpocrates:
In case it isn't clear, this really a bug fix (although I ended up improving the output along the way too). This is what `wip/hi-haddock` currently produces for [the `Binary` class from `binary`](http://hackage.haskell.org/package/binary-0.8.6.0/docs/Data-Binary.html#t:Binary):

screen shot 2018-09-10 at 7 44 24 pm

Here is what this branch produces instead:

screen shot 2018-09-05 at 6 50 50 am

----

Here is the output of the test cases I added:

### XHTML default signatures
screen shot 2018-09-10 at 7 31 38 pm

### XHTML default associated types
screen shot 2018-09-10 at 7 31 32 pm

### LaTeX default signatures
screen shot 2018-09-10 at 7 31 10 pm

haddock-api/src/Haddock/GhcUtils.hs
```diff
@@ -63,6 +66,13 @@ getMainDeclBinder (ForD _ (ForeignImport _ name _ _)) = [unLoc name]
 getMainDeclBinder (ForD _ (ForeignExport _ _ _ _)) = []
 getMainDeclBinder _ = []
 
+uniquifyClassSig :: (NamedThing name, SetName name) => Bool -> name -> name
+uniquifyClassSig False = id
+uniquifyClassSig _     = liftA2 setName (updateName . getName) id
+  where
+    updateName = liftA2 setNameUnique id $
+        liftA2 deriveUnique id ((+1) . getKey) . nameUnique
```
sjakobi: I see several issues with this function but it doesn't appear to be used at all?!

Issue #838, opened 2018-05-23
Type signature of raise# is wrong


andrewthad:
The original discussion of this issue is on [trac issue 15181](https://ghc.haskell.org/trac/ghc/ticket/15181). At GHCi, we can inspect the type of `GHC.Exts.raise#`:

```
>>> import GHC.Exts
>>> :set -XMagicHash
>>> :info raise#
raise# :: forall b (q :: RuntimeRep) (a :: TYPE q). b -> a
        -- Defined in ‘GHC.Prim’
```

However, the type of `raise#` given in the haddocks is missing the levity polymorphic kinds:

```
raise# :: a -> o
```

RyanGlScott:
This is not simple to fix. The problem is that `raise#` is defined in `GHC.Prim`, which involves a fair amount of magic. In particular, GHC wires in extra typing information for a good number of functions defined in `GHC.Prim`, which means that even though the type signature that Haddock sees for `raise#` is `a -> o` (as evidenced [here](https://hackage.haskell.org/package/ghc-prim-0.5.2.0/docs/src/GHC.Prim.html#raise%23)), in reality, GHC sneaks in some extra levity polymorphism under the hood.

I don't know a simple way to work around this.

harpocrates:
Note: this is fixed in the Hi Haddock branch:

<img width="529" alt="screen shot 2018-09-16 at 5 04 55 pm" src="https://user-images.githubusercontent.com/10766081/45602416-c4279800-b9d2-11e8-9675-cfa106c16abf.png">

RyanGlScott:
Wow! Out of curiosity, how does Hi Haddock figure out what the correct type for `raise#` is?

harpocrates:
> Wow! Out of curiosity, how does Hi Haddock figure out what the correct type for raise# is? 

That's the type that GHC uses internally, no? Hi Haddock gets its types by querying GHC (instead of by looking at surface syntax in the initial source file, as today's Haddock does). Even TH can figure out the right type for `raise#`:

```haskell
{-# LANGUAGE TemplateHaskell #-}

import Language.Haskell.TH.Syntax
import Language.Haskell.TH.Ppr
import GHC.Prim

main = putStrLn $(qReify (mkName "raise#") >>= lift . pprint)
-- GHC.Prim.raise# :: forall (b_0 :: *) (q_1 :: GHC.Types.RuntimeRep) (a_2 :: GHC.Prim.TYPE q_1) .
--                    b_0 -> a_2
```

[...]

PR #941, opened 2018-09-17
Default 'RuntimeRep' arguments unless otherwise specified

harpocrates:
This adds a module level Haddock option called `print-runtime-reps` which behaves like the GHC `-fprint-explicit-runtime-reps` flag. Hopefully this will let us restore beginner-friendliness to the `Prelude` types, without having to compromise for simplified types in `GHC.Prim`.

See the discussion in https://github.com/haskell/haddock/issues/838 for more.

### DIFF
--- a/haddock-api/src/Haddock/Convert.hs
+++ b/haddock-api/src/Haddock/Convert.hs
@@ -99,7 +99,7 @@ tyThingToLHsDecl prr t = case t of
              :: ClassATItem
              -> Either ErrMsg (LFamilyDecl GhcRn, Maybe (LTyFamDefltEqn GhcRn))
            extractAtItem (ATI at_tc def) = do
-             tyDecl <- synifyTyCon Nothing at_tc
+             tyDecl <- synifyTyCon prr Nothing at_tc
              famDecl <- extractFamilyDecl tyDecl
              let defEqnTy = fmap (noLoc . extractFamDefDecl famDecl . fst) def
              pure (noLoc famDecl, defEqnTy)
@@ -127,7 +127,7 @@ tyThingToLHsDecl prr t = case t of
          , tcdDocs = [] --we don't have any docs at this point
          , tcdCExt = placeHolderNamesTc }
     | otherwise
-    -> synifyTyCon Nothing tc >>= allOK . TyClD noExt
+    -> synifyTyCon prr Nothing tc >>= allOK . TyClD noExt
 
   -- type-constructors (e.g. Maybe) are complicated, put the definition
   -- later in the file (also it's used for class associated-types too.)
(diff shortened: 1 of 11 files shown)


harpocrates:
We'll have to remember to add `{-# OPTIONS_HADDOCK print-runtime-reps #-}` to the top of some of the internal files of `base` once Hi Haddock is merged.

---

Here is what the test cases look like.

screen shot 2018-09-17 at 8 43 22 am

screen shot 2018-09-17 at 8 43 12 am

haddock-api/src/Haddock/GhcUtils.hs
```diff
@@ -499,3 +502,45 @@ tyCoFVsOfTypes' []       fv_cand in_scope acc = emptyFV fv_cand in_scope acc
...
+        in go subs' ty
+      | otherwise
+      = ForAllTy (TvBndr (updateTyVarKind (go subs) var) flg)
+                 (go subs ty)
+
+    go subs (TyVarTy tv)
+      | tv `elemVarEnv` subs
+      = TyConApp liftedRepDataConTyCon []
+      | otherwise
+      = TyVarTy (updateTyVarKind (go subs) tv)
```
RyanGlScott: I think it's unnecessary to recurse into `tv`'s kind here. (At least, `IfaceType.defaultRuntimeRepVars` doesn't do this.)

haddock-api/src/Haddock/Interface/Create.hs
```diff
@@ -284,6 +285,8 @@ parseOption "hide"            = return (Just OptHide)
 parseOption "prune"           = return (Just OptPrune)
 parseOption "not-home"        = return (Just OptNotHome)
 parseOption "show-extensions" = return (Just OptShowExtensions)
+parseOption "print-runtime-reps" = return (Just OptPrintRuntimeRep)
+parseOption "print-explicit-runtime-reps" = return (Just OptPrintRuntimeRep)
```
RyanGlScott: Is there a reason for having two flags with the same effect? Why not just standardize on a single flag?

haddock-api/src/Haddock/Convert.hs
```diff
@@ -756,7 +771,7 @@ synifyInstHead (vs, preds, cls, types) = specializeInstHead $ InstHead
     ts' = map (synifyType WithinType vs) ts
     annot_ts = zipWith3 annotHsType is_poly_tvs ts ts'
     is_poly_tvs = mkIsPolyTvs (tyConVisibleTyVars cls_tycon)
-    synifyClsIdSig = synifyIdSig DeleteTopLevelQuantification vs
+    synifyClsIdSig = synifyIdSig True DeleteTopLevelQuantification vs
```
RyanGlScott: Why do you hardcode `PrintRuntimeReps` to `True` here instead of consulting the flag?

haddock-api/src/Haddock/Interface/Create.hs
harpocrates: No particular reason. I want this flag to be called `print-runtime-reps`, but I anticipate some confusion with GHC's `-fprint-explicit-runtime-reps` (cf. Cabal's `hyperlink-source` option vs Haddock's `hyperlinked-source` option). Do you have an opinion one way or another?

haddock-api/src/Haddock/GhcUtils.hs
harpocrates: It is necessary, since we do recurse into type variable's kinds when computing the inferred order of type variables (in `tyCoFVsBndr'`). Perhaps we should not recurse into the type variables kind there?

haddock-api/src/Haddock/Interface/Create.hs
RyanGlScott: My inclination would be to have `-fprint-explicit-runtime-reps` be the shared name of the flag across GHC and Haddock. I suppose I would be fine with picking `-fprint-runtime-reps` to be the name of the Haddock flag as well, but I definitely think we shouldn't come up with two different ways of referring to the same flag—that just seems like a recipe for confusion.

haddock-api/src/Haddock/Convert.hs
harpocrates: Yeah. So this brings up a bigger question: should we thread through the `RuntimeRep` defaulting logic to everything involving `Type` in this module? Perhaps so, but that is going to be a fair bit of plumbing.

I think it is simpler to start with defaulting only top-level signatures, but I'm willing to be told otherwise.

haddock-api/src/Haddock/GhcUtils.hs
RyanGlScott: Upon further thought, maybe it's better if we do recurse here.

For pretty-printing purposes, it's probably overkill, since if you have a type like:

```
forall (r :: RuntimeRep) (a :: TYPE r). Int -> a
```

Then after defaulting, it'll still display as:

```
forall a. Int -> a
```

Regardless of whether you recurse into the kind of `a` (in `Int -> a`) or not. On the other hand, if you _don't_ recurse into the kind of `a`, then you have a strange situation where the kind of `a` in `forall a` is `Type`, but the kind of `a` in `Int -> a` is `TYPE r`. I suppose the only way that could ever really bite you is if you were to run Core Lint on these types (which I certainly don't imagine happens today), but for the sake of consistency, it might be best just to update the kind of `a` in `Int -> a` anyways.

haddock-api/src/Haddock/Convert.hs
RyanGlScott: If your goal is to avoid having any `RuntimeRep` quantification unless explicitly enabled, I think you'll eventually need to tackle class instances as well. After all, you can define levity polymorphic instances like so:

```haskell
λ> class C a
λ> instance forall (r :: RuntimeRep) a (b :: TYPE r). C (a -> b)
λ> :i C
class C (a :: k)        -- Defined at <interactive>:2:1
instance [safe] forall a b. C (a -> b)
  -- Defined at <interactive>:3:10
λ> :set -fprint-explicit-runtime-reps 
λ> :i C
class C (a :: k)        -- Defined at <interactive>:2:1
instance [safe] forall (r :: RuntimeRep) a (b :: TYPE r).
                C (a -> b)
  -- Defined at <interactive>:3:10
```

harpocrates: Gah. Alright. Thanks for the motivating example. 😄

harpocrates: Upon further inspection: I can't convince Haddock to ever print out runtime reps for instances. I'm inclined to leave this be for the moment. Plus this part of the `wip/hi-haddock` branch, so I'm trying to stay away from far-reaching changes.

sjakobi:
Thanks! :)

PR #949, opened 2018-10-18
Introduce NewOcean theme (take 3)

harpocrates:
This is an updated version of https://github.com/haskell/haddock/pull/782 (which was itself an updated version of https://github.com/haskell/haddock/pull/721), rebased against `ghc-8.6`. Since it is a branch on the official Haddock repo, anyone can contribute to it by making a PR that targets `wip/new-ocean`.

Assuming you have `ghc-8.6.1` and `cabal` installed an on `PATH`, building/testing should be as simple as:

    $ git clone https://github.com/haskell/haddock.git && cd haddock
    haddock$ git checkout wip/new-ocean
    haddock$ cabal new-build -w ghc-8.6.1
    haddock$ cabal new-test -w ghc-8.6.1

/cc @NunoAlexandre @gbaz @hvr


harpocrates:
TODOs from #782, which I'm about to close:

  - [ ] Top `Source` link can be misaligned (but only on certain page widths) ([screenshot + comment](https://github.com/haskell/haddock/pull/782#issuecomment-374871262))
  - [ ] NewOcean files need to be added to `data-files` in `haddock-api.cabal` ([comment](https://github.com/haskell/haddock/pull/782#issuecomment-376202551))
  - [x] Port some CSS changes from https://github.com/haskell/hackage-server/pull/731 ([comment](https://github.com/haskell/haddock/pull/782#issuecomment-376278712))
  - [x] Typeclasses are misindented ([comment](https://github.com/haskell/haddock/pull/782#issuecomment-376284937))
  - [ ] Figure out what to do with the style switcher
  - [ ] Synopsis button is buggy (still the old button and is it supposed to be left aligned?) ([screenshot + comment](https://github.com/haskell/haddock/pull/782#issuecomment-381425438))
  - [ ] `Source` link gets nudged to another line, but without the signature background, when there is a long type signature ([screenshot + comment](https://github.com/haskell/haddock/pull/782#issuecomment-393932016))
  - [ ] `Synopsis` button drops down and has a dotted line above it whenever the page is narrowed to the point that the "Contents" box takes up the whole width ([screenshot + comment](https://github.com/haskell/haddock/pull/782#issuecomment-393932016))

Some additionally slightly less specific concerns (which @sjakobi has also voiced [here](https://github.com/haskell/haddock/pull/782#issuecomment-393932016)).

 - [ ] Class docs and function docs probably should be indented a bit w.r.t their headers/types, otherwise it isn't as easy to spot where declarations start/end
 - [ ] Perhaps a slightly smaller font size? Even discounting the tighter left/right margins, I get just over half as many lines per screen as I used to...

      Small font sizes and compact layouts seems to be pretty typical for generated docs too: check out [javadoc](https://docs.oracle.com/javase/7/docs/api/java/nio/ByteBuffer.html), [scaladoc](https://www.scala-lang.org/api/current/scala/collection/immutable/List.html), and even [rustdoc](https://doc.rust-lang.org/std/vec/struct.Vec.html) (and even _that_ is a bit big for my liking).

Seems to me like this is actually a pretty small list of things to fix!

NunoAlexandre:
@harpocrates I'm happy if you take this on. Since I became a father I have had no time to get back to this properly and won't, so I'm glad to see you taking this further.

NunoAlexandre:
> Small font sizes and compact layouts seems to be pretty typical for generated docs too: check out javadoc, scaladoc, and even rustdoc (and even that is a bit big for my liking).

javadoc and scaladoc are not an example to follow per se. Rustdoc is a good example of good font size and of balance.

haddock-api/haddock-api.cabal
```diff
@@ -34,6 +34,11 @@ data-files:
   html/Ocean.std-theme/ocean.css
   html/Ocean.std-theme/plus.gif
   html/Ocean.std-theme/synopsis.png
+  html/NewOcean.std-theme/hslogo-16.png
+  html/NewOcean.std-theme/minus.gif
+  html/NewOcean.std-theme/new-ocean.css
+  html/NewOcean.std-theme/plus.gif
+  html/NewOcean.std-theme/synopsis.png
```
NunoAlexandre: Sorry if I am annoying (it's much easier to comment than doing it myself), but for the synopsis and "minus" I moved to use unicode instead of an actual imagine.

harpocrates: Not at all - please continue to drop in and comment on this sort of thing!

I did notice that the `synopsis` button is still the one of old. (And looks funky, since the old button was designed to hang off the right side of the screen). I'll investigate later.

fredefox:
So happy to see that this hasn't died. I just tried 

    cabal new-install haddock

But got some errors due to some missing data-files. Diff:

```diff
diff --git a/haddock-api/haddock-api.cabal b/haddock-api/haddock-api.cabal
index 70e4c2cc..dc69d857 100644
--- a/haddock-api/haddock-api.cabal
+++ b/haddock-api/haddock-api.cabal
@@ -20,25 +20,6 @@ extra-source-files:
 data-dir:
   resources
 data-files:
-  html/quick-jump.min.js
-  html/haddock-bundle.min.js
-  html/quick-jump.css
-  html/solarized.css
-  html/highlight.js
-  html/Classic.theme/haskell_icon.gif
-  html/Classic.theme/minus.gif
-  html/Classic.theme/plus.gif
-  html/Classic.theme/xhaddock.css
-  html/Ocean.std-theme/hslogo-16.png
-  html/Ocean.std-theme/minus.gif
-  html/Ocean.std-theme/ocean.css
-  html/Ocean.std-theme/plus.gif
-  html/Ocean.std-theme/synopsis.png
-  html/NewOcean.std-theme/hslogo-16.png
-  html/NewOcean.std-theme/minus.gif
-  html/NewOcean.std-theme/new-ocean.css
-  html/NewOcean.std-theme/plus.gif
-  html/NewOcean.std-theme/synopsis.png
   latex/haddock.sty
 
 library
```

I say this with the caveat that I'm running dev versions of both GHC and cabal.

This is with e2cd0a8095e6fe3c5081e73beff0d2e2a1cc00f8.

fredefox:
The error I mentioned [here](https://github.com/haskell/cabal/issues/5637#issuecomment-432815956) looks self-inflicted given the diff I provided above.

fredefox:
So this should fix the above referenced issue:

```diff
diff --git a/haddock-api/haddock-api.cabal b/haddock-api/haddock-api.cabal
index 70e4c2cc..b812bfae 100644
--- a/haddock-api/haddock-api.cabal
+++ b/haddock-api/haddock-api.cabal
@@ -29,12 +29,11 @@ data-files:
   html/Classic.theme/minus.gif
   html/Classic.theme/plus.gif
   html/Classic.theme/xhaddock.css
-  html/Ocean.std-theme/hslogo-16.png
-  html/Ocean.std-theme/minus.gif
-  html/Ocean.std-theme/ocean.css
-  html/Ocean.std-theme/plus.gif
-  html/Ocean.std-theme/synopsis.png
-  html/NewOcean.std-theme/hslogo-16.png
+  html/Ocean.theme/hslogo-16.png
+  html/Ocean.theme/minus.gif
+  html/Ocean.theme/ocean.css
+  html/Ocean.theme/plus.gif
+  html/Ocean.theme/synopsis.png
   html/NewOcean.std-theme/minus.gif
   html/NewOcean.std-theme/new-ocean.css
   html/NewOcean.std-theme/plus.gif
```

For some reason it didn't get copied to the install location for haddock-api, but I think that may just be me that's a bit too tired right now. Hope this helps someone.

fredefox:
I manage to get it to work, even with `cabal new-haddock`. Thank you @harpocrates!

fredefox:
Is the front-page you see on hackage specific to hackage? Was it removed or was it never there?

harpocrates:
> Is the front-page you see on hackage specific to hackage? Was it removed or was it never there?

That is specific to `hackage-server`. The new purple style on Hackage comes from https://github.com/haskell/hackage-server/pull/693. This PR is about moving that theme into the actual docs too.

fredefox:
Sorry @harpocrates I deleted the question cause I realized I was asking something I knew the answer to.

Another thing, the side-bar seems to have disappeared, is this by design? I couldn't find any documentation on the design-choices in the repo...

harpocrates:
> Another thing, the side-bar seems to have disappeared, is this by design? I couldn't find any documentation on the design-choices in the repo...

I'm not the person to ask - I too wish we had documentation about these sort of design choices. Please contribute input! I think the side bar is now on the left hand side. If your window isn't wide enough, the side bar also turns into a full-width component at the top of the page.

haddock-api/resources/html/NewOcean.std-theme/new-ocean.css
```diff
@@ -0,0 +1,896 @@
...
+#synopsis summary {
+  display: block;
+  float: left;
+  width: 29px;
+  color: rgba(255,255,255,0);
+  height: 110px;
+  margin: 0;
+  font-size: 1px;
+  padding: 0;
+  background: url(synopsis.png) no-repeat 0px -8px;
```
harpocrates: @NunoAlexandre  Doesn't `url(synopsis.png)` (and `url(plus.gif)` and `url(minus.gif)` elsewhere in this file) indicate you are still using these assets? Are we missing some of your newer changes? Forgive the potentially stupid question - I am not familiar at all with CSS...

NunoAlexandre: @harpocrates not a stupid question at all. Either some of my changes are missing indeed, or those styling elements referencing `synopsis` and `minus` were not taking any effect in the html pages I tested the style on and therefore I could have forgotten to delete them. They seem to be used though (considering your screenshots), therefore something's off. I will have a look 👍

harpocrates: I did something with the synopsis before you commented your response. Feel free to revert that commit if you feel it clashes with everything else too much. I would love it if you took a look at the `minus` - there's one case that really feels like a bit of an eyesore:

screen shot 2018-10-25 at 7 48 36 am

The old minus clashes with the new vertical indent line as well as with the new unicode pluses in the instances list. That said, aligning the `Instances` and `Constructors` section is obviously something we'd also like.

harpocrates:
I'm mostly satisfied with these changes having addressed (I think?) the comments from the previous threads. I'd appreciate folks giving things another careful look over so we can catch less obvious issues. Here are some sample docs (generated with 88f9060):

  * [`lens-4.17-docs`](https://harpocrates.github.io/lens-4.17-docs/)
  * [`singletons-2.5-docs`](https://harpocrates.github.io/singletons-2.5-docs/)
  * [`these-0.7.5-docs`](https://harpocrates.github.io/these-0.7.5-docs/)

My one issue is that we are still using `plus.gif` and `minus.gif` - but I don't know how to get rid of them nicely (things no longer align as nicely with the unicode plus.. maybe that's ok?).

fredefox:
> Perhaps we aren't referring to the same things. By the "sidebar", do you mean the "Contents" box? You'll need that box to be non-empty for it to show up at all.

Sorry for the delay @harpocrates, I was having some issue getting the haddocks to display with the new theme. This is the screen I'm referring to:

2018-10-27-131140_1920x1026_scrot

This is what it looks like on hackage:

2018-10-27-131229_1920x1026_scrot

As you can see the dependencies etc. on the right is not there anymore.

fredefox:
Ah ok. Thank you for the clarification @harpocrates! I was not aware. Perhaps `stack` sets some other defaults. This is what I usually use.

harpocrates:
@NunoAlexandre @gbaz @hvr I'm pretty much ready to pull the trigger and merge this! Barring any new problems folks find, I'm gonna merge Friday night. So now is the moment to comment if you want something changed!

[Here is a link to the Haskell Hierarchical Libraries built with this new Haddock. I enabled the newly-fixed style switcher, so you can easily compare the new theme to the old one via the "Style" drop down in the top right.](https://harpocrates.github.io/ghc-head-libraries/)

Please feel free to forward/share the above link with folks who might be interested.

fredefox:
Could we change

    #table-of-contents {
      overflow-y: scroll;
    }

to `auto`?

fredefox:
I also have an issue that the main-content (aka. some things under `#interface`) doesn't seem to respect the `max-width` set on `#content` resulting in a very wide page with a horizontal scroll bar. Otherwise I'm really digging the new styles.

fredefox:
I also personally don't really like that the documentation for instance methods are laid out to the right of the signature (on the same line) but I realize that this is most probably outside the scope of this (already significant PR). Perhaps this deserves a new issue?

fredefox:
I realize that dropping the `nowrap` places some of the "# Source" tags on a new line even though they fit on the same on. Perhaps a `min-width` will suffice. CSS is so filled with corner-cases. Quite infuriating in some circumstances...

harpocrates:
> Could we change
> 
> ```
> #table-of-contents {
>   overflow-y: scroll;
> }
> ```
> 
> to `auto`?

Sure! I'm a CSS noob and do not know what difference that makes though. 😄 

> I also have an issue that the main-content (aka. some things under `#interface`) doesn't seem to respect the `max-width` set on `#content` resulting in a very wide page with a horizontal scroll bar. Otherwise I'm really digging the new styles.

Yeah, that's not due to the new theme though. The problem is the way instance lists currently get formatted. They get quite wide in degenerate cases. See https://github.com/haskell/haddock/issues/300.

> I also personally don't really like that the documentation for instance methods are laid out to the right of the signature (on the same line) but I realize that this is most probably outside the scope of this (already significant PR). Perhaps this deserves a new issue?

I'm not sure what you mean by this. If it is a problem you can't see on the old theme (select "ocean" under the "style" menu in the top right), let's address it here. Otherwise make a separate issue for it. In either case, a screenshot will help explain what you mean by this.

fredefox:
Yeah it is also like this with the old theme. That's what I was hinting at (could've been more explicit):

2018-11-08-024723_1454x530_scrot

I think it would be nicer if the 2nd `td` was displayed below the former. Ideally we wouldn't use a table here in the first place. But I can understand from previous discussions that some people really value their horizontal screen real estate...

fredefox:
> Sure! I'm a CSS noob and do not know what difference that makes though.

The difference is that the scroll bar is only visible when needed.

dfordivam:
Hi
Thanks for work on this!

I noticed that in some cases the "Source #" is coming in multiple lines at 100% zoom. This screenshot is from Data.Text

image

MaxGabriel:
This sticky content box is great!

image

I find it a little harder to parse these as distinct items compared to the "Ocean" theme. It seems like they're almost all at the same level of indentation, or that they don't feel like distinct items in a list. Maybe more top/bottom margin would help?

image

Compare to the Elm docs, for example:

image

cdepillabout:
I think the new redesign looks really good.

This is probably too late (and likely deserves its own issue), but if it would be easy to add, it would be great if individual sections in the haddocks were collapsible.  This would really help for huge haddock pages, like [accelerate's](http://hackage.haskell.org/package/accelerate-1.2.0.1/docs/Data-Array-Accelerate.html).

harpocrates:
@dfordivam That's unfortunately an existing issue (just more easily triggered now that font sizes are a bit larger. Not to say we shouldn't fix it, but I'm not going to let it hold this PR back.

screen shot 2018-11-08 at 8 11 25 pm

@MaxGabriel I've added some extra space before every element, hopefully that helps address your concern. It isn't as much as elm, but I'm concerned that going that far would upset too many folks who are accustomed to much denser docs.

@ivan-m Great idea! Done.

@cdepillabout That seems like a fine feature request (probably worth an issue on its own), but what does this bring you that you don't already get from the contents tab on the left?

cdepillabout:
@harpocrates 

> That seems like a fine feature request (probably worth an issue on its own), but what does this bring you that you don't already get from the contents tab on the left?

After playing around with the contents tab for a little bit, I think it might very-well give me most of what I want with folding.

I guess with folding what I would want is a way to hide everything I don't care about, so I could focus only on one or two sections.

However I think this is out of scope for this PR, so should probably be a separate feature request.

NunoAlexandre:
@harpocrates 

> but I'm not going to let it hold this PR back.

The way to do, in my opinion. Now you will get drown with bikeshedding with suggestions to improve every single detail. Better release it asap and keep on improving later in small iterations.

hvr:
> Better release it asap and keep on improving later in small iterations.

That's the thing... `haddock` is tied to GHC releases and it's tricky to update `haddock` alone; so releasing it before GHC isn't really going to help us here much. We basically have to wait until the next GHC release.

lthms:
That being said, merging it would probably have the good property of having a sense of achievement. Further improvements can happen then, until GHC makes a release. Right?

vlatkoB:
What I would like is a `switch` somewhere on the module page `that would collapse all instances`. When I'm just browsing a new package, the list of instances, which sometimes can be very long, only makes you lose focus. My two cents.

domenkozar:
Really great work. I have one suggestion in case you find it interesting. 

Right now `# source` is placed on the right side of the screen, completely separate from the type signature.

What if the function/declaration was a `#` link itself? And then maybe place `(source)` in grey at the end of the declaration.

Like so:
2018-11-09-114618_665x456_scrot

domenkozar:
I like the suggestion from @MaxGabriel, here it is in pictures:

Before:

2018-11-09-115102_1082x246_scrot


After:

2018-11-09-115044_945x222_scrot

fredefox:
@domenkozar:

> What if the function/declaration was a # link itself?

I like this idea. The problem is what do you do for e.g. instance decls? The nice thing about the hash tag (#) is that you can get something consistent across the board for various types of link-able gadgets.

fredefox:
What I personally find more confusing is that the # source thing feels a bit counter-intuitive to me. I always think that they do the same. I always have to stop up and think that one gives me a deep-link and the other gives me the sources. Perhaps if "Source" was replaced with a symbol... May just be me...

harpocrates:
@NunoAlexandre @lthms This PR is going to be merged as soon as activity dies down. I asked for feedback because I want as many eyes looking for regressions and small issues. We're no longer calling into question the whole theme 😄, or addressing every concern. The focus is small, __actionable__ suggestions/concerns.

@vlatkoB Others have wanted this before. Could you add your suggestion to https://github.com/haskell/haddock/issues/698? Also, it might help if you mentioned a package which displayed this behaviour... 

@domenkozar @fredefox Changing how `# Source` works (and fixing some of the existing problems with it, like https://github.com/haskell/haddock/pull/949#issuecomment-437197912) would be great, but probably belongs as a seperate issue/PR. Would either of you like to open one? Haddock is always looking for new contributors with ideas and time...

domenkozar:
@fredefox I agree. That would be the most minimal change to fix the most annoying bit. Moving it to the left (closer to what the action is about) would be an additional step. 

> The problem is what do you do for e.g. instance decls

That's a good point. I wonder why it the point of instance decls, all anchors lead to the type itself so I wonder what's the point? :) You can already click on the class or instance. Then you'll be able to see the source.

@harpocrates I agree, this shouldn't block your PR, I apologise.

domenkozar:
See https://github.com/haskell/haddock/issues/966

vlatkoB:
@harpocrates I searched for existing issue, but couldn't find one. :-(

What do you mean by 

> a package which displayed this behaviour

Most of slightly larger packages have many types with lots of instances.

harpocrates:
@vlatkoB

> @harpocrates I searched for existing issue, but couldn't find one. :-(

I thought I had referenced it? https://github.com/haskell/haddock/issues/698 is the issue in question.

> What do you mean by
> 
> > a package which displayed this behaviour
> 
> Most of slightly larger packages have many types with lots of instances.

Oh OK, that's fine! I thought perhaps you had a package in mind that had egregiously long instance lists. I was asking mostly because this hasn't bothered me personally much and because it was possible that this was the case only because I hadn't been viewing the same packages as you.

mcschroeder:
I strongly dislike the use of the circled plus/minus symbols as disclosure indicators. I think they are very confusing, especially since ⊕ is used pretty frequently as a mathematical operator. They are also a bit difficult to tell apart.

Might I suggest something more standard and unambiguous, like the commonly used disclosure triangles? For example:

screenshot 2018-11-09 at 23 25 13

This uses the unicode symbols ▿ and ▹ (U+25BF White Down-Pointing Small Triangle and U+25B9 White Right-Pointing Small Triangle, respectively), but I'm sure there's other possibilities.

harpocrates:
@mcschroeder Your objection (and proposed alternative) seems reasonable; I know several others have brought this up too (a couple more have messaged me privately). Given that this really changes something core about the style, I'm going to defer to the original author. @NunoAlexandre what do you think of using triangles instead of circles for elements that can be expanded/collapsed?

NunoAlexandre:
@harpocrates respect for picking this up so well and get it out 👍 thanks!

PR #954, opened 2018-10-25
Avoid more conflicts in generated ids 

### LINKED ISSUE #953: Expander/collapser ids can conflict to easily
We do some work to generate HTML id's in a deterministic fashion. However, we're not trying hard enough to make them unique. Here's an example: 

```haskell
module Bug953 where

{- | A foo

==== __Examples__

Foo example body
-}
data Foo = Foo'

{- | A bar

==== __Examples__

Bar example body
-}
data Bar = Bar'
```

It is impossible to expand the first example section (clicking on the plus expands the second section!) because the two generated HTML elements are given the same `ch:noid:0` id.

harpocrates:
This fixes #953 by passing more names into the generated ids.


sjakobi:
Nice! :)

PR #960, opened 2018-11-05
Only run MathJax on entities with "mathjax" class

### LINKED ISSUE #959: Bizarre HTML formatting of symbolic name
I originally discovered this oddity in the [Haddocks for `singletons-2.5`](http://hackage.haskell.org/package/singletons-2.5/docs/Data-Singletons-TypeError.html#t::-36--36-:-64--35--64--36--36-). Here is a minimal example of how to trigger the issue:

```haskell
{-# LANGUAGE TypeOperators #-}
module Bug where

data (:$$:@#@$$)
```
```
$ haddock Bug.hs --html -o bug
Haddock coverage:
Warning: Package name is not available.
   0% (  0 /  2) in 'Bug'
  Missing documentation for:
    Module header
    :$$:@#@$$ (Bug.hs:4)
```
haddock

harpocrates:
Correspondingly, we wrap all inline/diplay math in

    <span class="mathjax"> ... the math .... </span>

This fixes #959.

### DIFF
--- a/haddock-api/src/Haddock/Backends/Xhtml.hs
+++ b/haddock-api/src/Haddock/Backends/Xhtml.hs
@@ -120,17 +120,23 @@ copyHtmlBits odir libdir themes withQuickjump = do
 
 headHtml :: String -> Themes -> Maybe String -> Html
 headHtml docTitle themes mathjax_url =
-  header << [
-    meta ! [httpequiv "Content-Type", content "text/html; charset=UTF-8"],
-    thetitle << docTitle,
-    styleSheet themes,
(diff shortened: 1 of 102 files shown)


harpocrates:
A test case isn't useful here since the interesting bit is whether MathJax decides to run or not. Perhaps a screenshot will be more convincing.

```haskell
{-# LANGUAGE TypeOperators #-}
module Bug where

 -- | Some actual math \[\int_{\partial\Omega} \omega = \int_\Omega \mathrm{d}\omega\] not
 -- particularly associated with this operator.
data (:$$:@#@$$)
```

Now produces:

screen shot 2018-11-05 at 9 31 07 am

PR #963, opened 2018-11-08
Remove unused imports.

DavidEichmann:
Due to a bug in ghc, some unused imports do not yield warnings.
This commit will remove such unused imports in preparation for
the ghc bug fix (see https://ghc.haskell.org/trac/ghc/ticket/13064).


DavidEichmann:
I've submitted this too early. Please consider this still a WIP PR. It may turn out unneeded in the end, but it depends on the changes for [ghc issue 13064](https://ghc.haskell.org/trac/ghc/ticket/13064).

Issue #927, opened 2018-08-30
add option to turn warnings into errors


ghorn:
In my project whenever I build haddocks I seem to have many warnings of the form

`Warning: Some.Module: could not find link destinations for:`

Every once in a while I go through and fix them all, but I would prefer if I could have my continuous integration prevent these sort of regressions from going in in the first place. Is it currently possible to turn these warnings into errors? If not, what would it look like to add an option to enable this?

ghorn:
If I could get some guidance on this I'd be happy to contribute a patch.

harpocrates:
Thanks for the feature request!

The particular warnings you mention come from here:

https://github.com/haskell/haddock/blob/1868443b01232d57ec11dfc831ac0a6915a2b337/haddock-api/src/Haddock/Interface.hs#L112-L114

But there are also warnings coming from here:

https://github.com/haskell/haddock/blob/1868443b01232d57ec11dfc831ac0a6915a2b337/haddock-api/src/Haddock/Interface.hs#L191

It would be quite nice to have a better reporting mechanism than a direct `putStrLn`. That mechanism could then be configured with a flag to error out when it encounters warnings (and perhaps even carry along more information with it than just a string). Since all warnings are emitted from inside `Ghc`, we might even be able to re-use some of GHC's machinery.

harpocrates:
Here is what would be nice:

  * we add a new `Opt_WarnMalformedHaddock` constructor to GHC's `WarningFlag` for Haddock related warnings (perhaps we'll be able to re-use a future GHC warning for misplaced Haddocks)
  * Haddock can then generate warnings in any `GhcMonad` monad with

      ```haskell
      let msg = makeIntoWarning (Reason Opt_WarnMalformedHaddock) $
                mkWarnMsg dflags loc unqual $
                text "Looks like you might've done something wrong..."
      logWarnings (unitBag msg)
      ```

We'd get two things for free: nicely coloured error messages (matching GHC's!) and the ability to set `-Werror=malformed-haddock`. This is also a good opportunity to add locations to error/warning messages.

peddie:
@harpocrates, I submitted a new pull request at #969 that approaches this in the way we discussed on IRC.

PR #969, opened 2018-11-13
Use the Ghc monad to emit warnings

peddie:
Previously Haddock warnings were emitted directly to `stdout` via
`putStrLn`.  This commit changes the output to go via the `Ghc` monad
as a new type of Haddock warning.  To take advantage of using `Ghc` in
this way, source locations are now tracked and emitted as part of most
warning messages.

This is an attempt to address #927.

The corresponding GHC commits are [here](https://github.com/peddie/ghc/commit/a80ea5d11da07d20207f1137b5f11113fd3dba9c) and [here](https://github.com/peddie/ghc/commit/d4d7adaa17a4b8123e93731d376003d54cda6d00).

### DIFF
--- a/haddock-api/src/Haddock/Interface/Create.hs
+++ b/haddock-api/src/Haddock/Interface/Create.hs
@@ -100,14 +100,14 @@ createInterface tm flags modMap instIfaceMap = do
   -- Cabal can be trusted to pass the right flags, so this warning should be
   -- mostly encountered when running Haddock outside of Cabal.
   when (isNothing pkgName) $
-    liftErrMsg $ tell [L moduleLocation "Warning: Package name is not available." ]
+    liftErrMsg $ tell [L moduleLocation "Package name is not available." ]
 
   -- The renamed source should always be available to us, but it's best
   -- to be on the safe side.
   (group_, imports, mayExports, mayDocHeader) <-
     case renamedSource tm of
       Nothing -> do
-        liftErrMsg $ tell [L moduleLocation "Warning: Renamed source is not available." ]
+        liftErrMsg $ tell [L moduleLocation "Renamed source is not available." ]
         return (emptyRnGroup, [], Nothing, Nothing)
       Just x -> return x
 
@@ -147,7 +147,7 @@ createInterface tm flags modMap instIfaceMap = do
 
   -- The MAIN functionality: compute the export items which will
   -- each be the actual documentation of this module.
-  exportItems <- mkExportItems is_sig modMap pkgName mdl sem_mdl allWarnings gre
+  exportItems <- mkExportItems moduleLocation is_sig modMap pkgName mdl sem_mdl allWarnings gre
                    exportedNames decls maps fixMap unrestrictedImportedMods
                    splices exports all_exports instIfaceMap dflags
 
@@ -323,8 +323,8 @@ mkDocOpts :: SrcSpan -> Maybe String -> [Flag] -> Module -> ErrMsgM [DocOption]
 mkDocOpts loc mbOpts flags mdl = do
   opts <- case mbOpts of
     Just opts -> case words $ replace ',' ' ' opts of
-      [] -> tell [L loc "No option supplied to DOC_OPTION/doc_option"] >> return []
-      xs -> liftM catMaybes (mapM parseOption xs)
+      [] -> tell [L loc "No option supplied to OPTIONS_HADDOCK/haddock-opts"] >> return []
+      xs -> liftM catMaybes (mapM (parseOption loc) xs)
     Nothing -> return []
   pure (foldl go opts flags)
   where
@@ -338,13 +338,13 @@ mkDocOpts loc mbOpts flags mdl = do
             | m == Flag_ShowExtensions mdlStr = OptIgnoreExports : os
             | otherwise                       = os
 
-parseOption :: String -> ErrMsgM (Maybe DocOption)
-parseOption "hide"            = return (Just OptHide)
-parseOption "prune"           = return (Just OptPrune)
-parseOption "ignore-exports"  = return (Just OptIgnoreExports)
-parseOption "not-home"        = return (Just OptNotHome)
-parseOption "show-extensions" = return (Just OptShowExtensions)
-parseOption other = tell [dislocatedErrMsg $ "Unrecognised option: " ++ other] >> return Nothing
+parseOption :: SrcSpan -> String -> ErrMsgM (Maybe DocOption)
+parseOption _ "hide"            = return (Just OptHide)
+parseOption _ "prune"           = return (Just OptPrune)
+parseOption _ "ignore-exports"  = return (Just OptIgnoreExports)
+parseOption _ "not-home"        = return (Just OptNotHome)
+parseOption _ "show-extensions" = return (Just OptShowExtensions)
+parseOption loc other           = tell [L loc $ "Unrecognised option: " ++ other] >> return Nothing
 
 
 --------------------------------------------------------------------------------
@@ -635,7 +635,8 @@ collectDocs = go Nothing []
 -- We create the export items even if the module is hidden, since they
 -- might be useful when creating the export items for other modules.
 mkExportItems
-  :: Bool               -- is it a signature
+  :: SrcSpan            -- module location, needed for error messages
+  -> Bool               -- is it a signature
   -> IfaceMap
   -> Maybe Package      -- this package
   -> Module             -- this module
@@ -654,12 +655,12 @@ mkExportItems
   -> DynFlags
   -> ErrMsgGhc [ExportItem GhcRn]
 mkExportItems
-  is_sig modMap pkgName thisMod semMod warnings gre exportedNames decls
+  loc is_sig modMap pkgName thisMod semMod warnings gre exportedNames decls
   maps fixMap unrestricted_imp_mods splices exportList allExports
   instIfaceMap dflags =
   case exportList of
     Nothing      ->
-      fullModuleContents is_sig modMap pkgName thisMod semMod warnings gre
+      fullModuleContents loc is_sig modMap pkgName thisMod semMod warnings gre
         exportedNames decls maps fixMap splices instIfaceMap dflags
         allExports
     Just exports -> liftM concat $ mapM lookupExport exports
@@ -673,7 +674,7 @@ mkExportItems
       return [ExportDoc doc]
 
     lookupExport (IEDocNamed _ str, _)      = liftErrMsg $
-      findNamedDoc str [ unL d | d <- decls ] >>= \case
+      findNamedDoc loc str [ unL d | d <- decls ] >>= \case
         Nothing -> return  []
         Just docStr -> do
           doc <- processDocStringParas dflags pkgName gre docStr
@@ -694,7 +695,8 @@ mkExportItems
       availExportItem is_sig modMap thisMod semMod warnings exportedNames
         maps fixMap splices instIfaceMap dflags avail
 
-availExportItem :: Bool               -- is it a signature
+availExportItem :: SrcSpan            -- module location, for error messages
+                -> Bool               -- is it a signature
                 -> IfaceMap
                 -> Module             -- this module
                 -> Module             -- semantic module
@@ -707,7 +709,7 @@ availExportItem :: Bool               -- is it a signature
                 -> DynFlags
                 -> AvailInfo
                 -> ErrMsgGhc [ExportItem GhcRn]
-availExportItem is_sig modMap thisMod semMod warnings exportedNames
+availExportItem loc is_sig modMap thisMod semMod warnings exportedNames
   (docMap, argMap, declMap, _) fixMap splices instIfaceMap
   dflags availInfo = declWith availInfo
   where
@@ -728,8 +730,8 @@ availExportItem is_sig modMap thisMod semMod warnings exportedNames
               -- parents is also exported. See note [1].
               | t `notElem` declNames,
                 Just p <- find isExported (parents t $ unL decl) ->
-                do liftErrMsg $ tell [ dislocatedErrMsg $
-                     "Warning: " ++ moduleString thisMod ++ ": " ++
+                do liftErrMsg $ tell [ L loc $
+                     moduleString thisMod ++ ": " ++
                      pretty dflags (nameOccName t) ++ " is exported separately but " ++
                      "will be documented under " ++ pretty dflags (nameOccName p) ++
                      ". Consider exporting it together with its parent(s)" ++
@@ -768,7 +770,7 @@ availExportItem is_sig modMap thisMod semMod warnings exportedNames
               case M.lookup (nameModule t) instIfaceMap of
                 Nothing -> do
                    liftErrMsg $ tell
-                      [dislocatedErrMsg $ "Warning: Couldn't find .haddock for export " ++ pretty dflags t]
+                      [L loc $ "Couldn't find .haddock for export " ++ pretty dflags t]
                    let subs_ = availNoDocs avail
                    availExportDecl avail decl (noDocForDecl, subs_)
                 Just iface ->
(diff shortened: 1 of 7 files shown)


haddock-api/src/Haddock/Interface/Create.hs
```diff
@@ -99,18 +100,18 @@ createInterface tm flags modMap instIfaceMap = do
   -- Cabal can be trusted to pass the right flags, so this warning should be
   -- mostly encountered when running Haddock outside of Cabal.
   when (isNothing pkgName) $
-    liftErrMsg $ tell [ "Warning: Package name is not available." ]
+    liftErrMsg $ tell [L moduleLocation "Warning: Package name is not available." ]
```
harpocrates: The `"Warning: "` is superfluous. You end up displaying messages that look like

```
Example.hs:1:1: warning: Warning: Package name is not available.
  |
1 | module Example where
  | ^
```

haddock-api/src/Haddock/Interface/Create.hs
```diff
@@ -99,18 +100,18 @@ createInterface tm flags modMap instIfaceMap = do
...
 
   -- The renamed source should always be available to us, but it's best
   -- to be on the safe side.
   (group_, imports, mayExports, mayDocHeader) <-
     case renamedSource tm of
       Nothing -> do
-        liftErrMsg $ tell [ "Warning: Renamed source is not available." ]
+        liftErrMsg $ tell [L moduleLocation "Warning: Renamed source is not available." ]
```
harpocrates: As before, the "Warning: " is superfluous.

haddock-api/src/Haddock/Interface/Create.hs
```diff
@@ -318,11 +319,11 @@ parseWarning dflags gre w = case w of
...
 
 
-mkDocOpts :: Maybe String -> [Flag] -> Module -> ErrMsgM [DocOption]
-mkDocOpts mbOpts flags mdl = do
+mkDocOpts :: SrcSpan -> Maybe String -> [Flag] -> Module -> ErrMsgM [DocOption]
+mkDocOpts loc mbOpts flags mdl = do
   opts <- case mbOpts of
     Just opts -> case words $ replace ',' ' ' opts of
-      [] -> tell ["No option supplied to DOC_OPTION/doc_option"] >> return []
+      [] -> tell [L loc "No option supplied to DOC_OPTION/doc_option"] >> return []
```
harpocrates: I know this isn't your doing, but that error message needs to be revised (how old is it!?). The pragma is called `OPTIONS_HADDOCK` and the GHC option is called `-haddock-opts`.

haddock-api/src/Haddock/Interface/Create.hs
```diff
@@ -343,7 +344,7 @@ parseOption "prune"           = return (Just OptPrune)
 parseOption "ignore-exports"  = return (Just OptIgnoreExports)
 parseOption "not-home"        = return (Just OptNotHome)
 parseOption "show-extensions" = return (Just OptShowExtensions)
-parseOption other = tell ["Unrecognised option: " ++ other] >> return Nothing
+parseOption other = tell [dislocatedErrMsg $ "Unrecognised option: " ++ other] >> return Nothing
```
harpocrates: This error message should look almost exactly like the `mkDocOpts` one since they are referring to the same options. That includes passing in the same module location (instead of `dislocatedErrMsg`).

haddock-api/src/Haddock/Interface/Rename.hs
```diff
@@ -68,20 +68,23 @@ renameInterface dflags renamingEnv warnings iface =
...
+             | n <- missingNames
+             , not (isSystemName n)
+             , not (isBuiltInSyntax n)
+             , Exact n /= eqTyCon_RDR
+             ]
+
+      warnMsg (name, msg) =
+        L (nameSrcSpan name) $
+        "Warning: " ++ moduleString (ifaceMod iface) ++
+        ": could not find link destinations for: " ++ msg
```
harpocrates: I think this is the biggest issue in this PR, and it is a totally non-obvious one. `nameSrcSpan name` is unfortunately _not_ the location we want. Try it out. We get these names from querying the `tcg_rdr_env` in GHC, so the location attached to the names is pretty much never useful.

There's probably some interesting location to attach to this warning (maybe the location of the initial docstring?) but piping that location through is going to take some care. If you want, you can duck out of this with a plain module location (if you take this path, let's go back to generating just one warning per module).

The warning message shouldn't have `Warning:` or  `moduleString (ifaceMod iface)` though...

haddock-api/src/Haddock/Interface/Create.hs
```diff
@@ -727,7 +728,7 @@ availExportItem is_sig modMap thisMod semMod warnings exportedNames
               -- parents is also exported. See note [1].
               | t `notElem` declNames,
                 Just p <- find isExported (parents t $ unL decl) ->
-                do liftErrMsg $ tell [
+                do liftErrMsg $ tell [ dislocatedErrMsg $
```
harpocrates: As before, the `"Warning: "` is superfluous. Let's also pass the `moduleLocation` here.

haddock-api/src/Haddock/Interface/Create.hs
```diff
@@ -767,7 +768,7 @@ availExportItem is_sig modMap thisMod semMod warnings exportedNames
               case M.lookup (nameModule t) instIfaceMap of
                 Nothing -> do
                    liftErrMsg $ tell
-                      ["Warning: Couldn't find .haddock for export " ++ pretty dflags t]
+                      [dislocatedErrMsg $ "Warning: Couldn't find .haddock for export " ++ pretty dflags t]
```
harpocrates: As before, the `"Warning: "` is superfluous. Let's also pass the `moduleLocation` here.

haddock-api/src/Haddock/Interface/Create.hs
```diff
@@ -964,7 +965,8 @@ moduleExport thisMod dflags ifaceMap instIfaceMap expMod =
           Just iface -> return [ ExportModule (instMod iface) ]
           Nothing -> do
             liftErrMsg $
-              tell ["Warning: " ++ pretty dflags thisMod ++ ": Could not find " ++
+              tell [dislocatedErrMsg $
+                    "Warning: " ++ pretty dflags thisMod ++ ": Could not find " ++
```
harpocrates: The `"Warning: " ++ pretty dflags thisMod ++ ` is superfluous. Let's also pass the `moduleLocation` here.

haddock-api/src/Haddock/Interface/Create.hs
```diff
@@ -1200,14 +1202,15 @@ mkMaybeTokenizedSrc dflags flags tm
             tokens <- liftGhcToErrMsgGhc (liftIO (mkTokenizedSrc dflags summary src))
             return $ Just tokens
         Nothing -> do
-            liftErrMsg . tell . pure $ concat
+            liftErrMsg . tell . pure . L moduleLocation $ concat
                 [ "Warning: Cannot hyperlink module \""
```
harpocrates: As before, lets get rid of the `"Warning: "` in the message.

haddock-api/src/Haddock/Interface/Create.hs
```diff
@@ -1225,7 +1228,7 @@ findNamedDoc :: String -> [HsDecl GhcRn] -> ErrMsgM (Maybe HsDocString)
 findNamedDoc name = search
   where
     search [] = do
-      tell ["Cannot find documentation for: $" ++ name]
+      tell [dislocatedErrMsg $ "Cannot find documentation for: $" ++ name]
```
harpocrates: Let's also pass the `moduleLocation` here.

haddock-api/src/Haddock/Interface/LexParseRn.hs
```diff
@@ -164,9 +164,10 @@ outOfScope dflags x =
     Exact name -> warnAndMonospace name  -- Shouldn't happen since x is out of scope
   where
     warnAndMonospace a = do
-      tell ["Warning: '" ++ showPpr dflags a ++ "' is out of scope.\n" ++
-            "    If you qualify the identifier, haddock can try to link it\n" ++
-            "    it anyway."]
+      tell $ fmap (rdrNameErrMsg x)
+        ["Warning: '" ++ showPpr dflags a ++ "' is out of scope.\n" ++
```
harpocrates: I'd get rid of `rdrNameErrMsg` here - it is always going to `noLoc` (as the comment on the `Exact` case indicates).

As before, the `"Warning: ` prefix in the error message is redundant.

haddock-api/src/Haddock/Interface/LexParseRn.hs
```diff
@@ -192,7 +193,7 @@ ambiguous dflags x gres = do
   -- of the same name, but not the only constructor.
   -- For example, for @data D = C | D@, someone may want to reference the @D@
   -- constructor.
-  when (length noChildren > 1) $ tell [msg]
+  when (length noChildren > 1) $ tell $ [rdrNameErrMsg x msg]
```
harpocrates: As before, the `"Warning: "` prefix in `msg` is redundant.

haddock-api/src/Haddock/Interface/Create.hs
peddie: Would you prefer that I make this change in a separate commit so the commits remain atomic / minimal?

harpocrates: I intend to squash before merging, so it shouldn't make a difference.

haddock-api/src/Haddock/Interface/Rename.hs
peddie: I see why this name is not useful.  I'm still trying to figure out where I might find the initial docstring's location.  It can't be in the `Interface`, I don't think, because that's where we got the problematic `Name` from in the first place (e.g. `ifaceDocMap`, `ifaceArgMap`).  Should I be seeking a way to attach `SrcSpan` info to the `Maps` in `Haddock.Interface.Create.mkMaps`, using the info in the given `LhsDecl GhcRn`s?

Can you provide any hints off the top of your head?

peddie: Sorry, I meant the `HsDocString`s corresponding to the `LhsDecl GhcRn`s.

harpocrates: Yeah, I even did a bunch of this a while ago. This should serve as a pretty good inspiration: https://github.com/harpocrates/haddock/commit/2b508013bc0fb348c909e6168b5c57ce617f4282 (it's almost exactly what you need). Basically, try to replace as many `HsDocString`'s with `LHsDocString`'s.

peddie: I'm still struggling to get this running, but I'm getting over my head when it comes to the parser side of things.  I started out when I last posted by changing `HsDocString` to `LHsDocString` everywhere, but it seems like in `LexParseRn.hs` and `Parser.hs` a bunch of individual docstrings get merged into a `MetaDoc` of some kind, so I can't figure out a way to avoid digging in and instead attach location information afterwards.  Your branch makes a lot of the same changes, but it modifies and relies on a type that doesn't seem to exist for me on `ghc-head` (`NsRdrName`) when I get into this part of the code (I'm not sure it compiles; I think there is some missing namespace plumbing as well?).

Having spent all afternoon chugging away at this, I think I'm out of time to keep poking, and I'm going to just report one warning per module and give the module location.

harpocrates: That sounds fine to me. I agree that this is a pretty deep rabbit hole too - the one warning per module with the module location is already a step forwards. 😄

peddie:
Sorry for the delay.  I think I've addressed all the requested changes.  In testing the changes, I realised that the out-of-scope and ambiguous identifier warnings changed from
```
Warning: 'mapAccumR' is out of scope.
    If you qualify the identifier, haddock can try to link it anyway.
```
to
```
<no location info>: warning:
    'mapAccumR' is out of scope.
    If you qualify the identifier, haddock can try to link it anyway.
```
If I were a user, I'd be annoyed to receive the latter ("thanks for pointing out that neither of us knows where this problem is, Haddock"), so I added the last commit in the series, which plumbs through module location info.  Now the messages look like:
```
libraries/containers/Data/IntMap/Internal.hs:1:1: warning:
    'mapAccumR' is out of scope.
    If you qualify the identifier, haddock can try to link it anyway.
  |
1 | {-# LANGUAGE CPP #-}
  | ^
```
Unfortunately, as you can see, this always grabs the source span at the beginning of the file in question, which usually has nothing to do with the link destinations.  What do you think?

peddie:
I've made my patch against the `ghc-head` branch, but it looks like the tests are failing due to lack of a field introduced to GHC a few weeks ago.  Is it possible the test suite is testing against 8.6 or an outdated `HEAD` branch of the compiler?

harpocrates:
> I've made my patch against the ghc-head branch, but it looks like the tests are failing due to lack of a field introduced to GHC a few weeks ago. Is it possible the test suite is testing against 8.6 or an outdated HEAD branch of the compiler?

Don't worry about CI on `ghc-head`; it hardly ever works. I'll test locally when I review (which should be in the next couple of days - I did notice that you addressed my comments, I've just not found time to give the final thorough review it deserves).

harpocrates:
This looks great!

I'm not going to merge just yet because I'd like to wait for the next GHC release cycle in order to dogfood on HEAD for a while (the new 8.8 release _just_ got cut). I'm a bit concerned that we might end up breaking the workflows of people who are using `-Werror`. That is especially a problem if we don't have a specific flag with which to disable just Haddock warnings or if there are some warnings in Haddock which cannot be avoided (I think there might be some nasty edge cases... for instance missing links to methods which are mentioned in `MINIMAL` pragmas but are otherwise not exported).

peddie:
Thanks for the review and thanks again for your guidance while working on this feature.  Is there anything else you need from me now or in the future to ensure that this goes in smoothly?

I assume once `Opt_WarnMalformedHaddock` makes it into `DynFlags.hs`, it's possible to disable only Haddock warnings or prevent only Haddock warnings from being errors?

harpocrates:
I don’t think so. It might be necessary to rebase at some point (depending on how long it takes me to finish the motivating patch for `Opt_WarnMalformedHaddock`). I’ll ping you if so.

Yes, once `Opt_WarnMalformedHaddock` makes it into DynFlags, it’ll be possible to disable Haddock warnings, or prevent just Haddock warnings from being errors. I think it should even be possible to do this on a module-by-module basis with `OPTIONS_GHC`.

Thanks again for all your hard work!

guibou:
Hello.

Any news about this?

I'm regularly rebasing these changes on top of my current GHC and that's a painful process. What can I do to help getting this merged?

simonmichael:
Ping ? Any movement on haddock's logging would be great. Perhaps this would also fix the -w/--no-warnings flag, which I think haddock currently ignores ?

harpocrates:
We are actually very close to this! There is a GHC patch that will introduce a Haddock related warning flag to GHC's flags, and it is close to being merged: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/2377.

@guibou I apologize for the delay here - the blocker was that without a flag under which to register these warnings, it won't be possible to turn them on or off directly. This would be very annoying in the presence of `-Werror`. Also, don't worry about rebasing changes - I can help with that part. Thanks again for putting this together.

guibou:
@harpocrates what is the status of this pull request.

At work, I'm rebasing this on each GHC release. I'm starting to think about another approach (using a post processing of the haddock cli output) to detect error and I'm wondering where my motivation will be best used.

So what is missing for that PR to be merged? Are you interested by a rebase on top of haddock master? If not, are we closing this PR as "won't fix"?

Kleidukos:
@peddie Does this PR affect non-GHC consumers? If no, could you please re-open it targeting the `ghc-9.0` branch so that all may benefit from it?

peddie:
@Kleidukos I think this PR could affect anything that's used to the current format of Haddock errors -- is that what you're asking?

I made this patch over 2 years ago, and unfortunately I do not currently have the time to continue working on it.  It's difficult for me to prioritize tasks that have already been "accepted" and then left untouched by the maintainers for 2 years and several GHC releases.  I hope someone with more time and/or the ability to directly merge the patch when it's ready can find the time to update it and land it so that all may benefit, as you say.

Kleidukos:
> @Kleidukos I think this PR could affect anything that's used to the current format of Haddock errors -- is that what you're asking?

Yes. The `ghc-head` branch is reserved for the GHC-specific patches, like glue code. Patches to the documentation generator should go toward the current GHC release branch, which is currently `ghc-9.0`. 

> I made this patch over 2 years ago, and unfortunately I do not currently have the time to continue working on it.  It's difficult for me to prioritize tasks that have already been "accepted" and then left untouched by the maintainers for 2 years and several GHC releases.  I hope someone with more time and/or the ability to directly merge the patch when it's ready can find the time to update it and land it so that all may benefit, as you say.


That's understandable. Thanks for your contribution, I am certain it will live under one shape or another in the codebase one day. 

Have a nice day!

PR #970, opened 2018-11-15
Follow GHC HEAD's HsTypes.Promoted -> BasicTypes.PromotionFlag change

alpmestan:
It got introduced in ghc/ghc@ae2c9b40f5b6bf272251d1f4107c60003f541b62 and without the patch from this PR, we cannot build haddock.


harpocrates:
Thanks!

We should probably also port over to the various backends the pretty-printing logic that SPJ added, but let's start by unbreaking things...

Issue #441, opened 2015-08-27
LINE pragmas causes missing documentation


mjakob:
Dear Haddock developers,

I ran into a problem with Haddock while using [c2hs](https://github.com/haskell/c2hs), but I think this issue can be applicable for any pre-processor using LINE pragmas. I managed to simplify the issue to the following: say we have a source file _LinePragmaIssue.chs_ and we run a pre-processor on it to produce _LinePragmaIssue.hs_:

```haskell
{-# LINE 1 "LinePragmaIssue.chs" #-}
-- |Module LinePragmaIssue
module LinePragmaIssue ( B ) where
-- This and the following two lines were generated by the pre-processor:
data A = SomethingLongerThanTheAnnotationForB

{-# LINE 4 "LinePragmaIssue.chs" #-}
-- |Data type B.
data B = B
```

If Haddock is applied to this file, the output is:

```
$ haddock LinePragmaIssue.hs
Haddock coverage:
  50% (  1 /  2) in 'LinePragmaIssue'
  Missing documentation for:
    B (LinePragmaIssue.chs:5)
```

Since the (generated) line `data A = SomethingLongerThanTheAnnotationForB` has the same line number as the Haddock annotation for `B` when the LINE pragmas are taken into account and since the declaration is longer, it is sorted after the latter, i.e., the sorted document structure that Haddock uses is:

```
-- |Data type B.
data A = SomethingLongerThanTheAnnotationForB
data B = B
```

This results in B being undocumented, which was unexpected, at least to me.

If this is considered a bug, one solution I can think of is to ignore the LINE pragmas while matching comments to declarations, but this might be built into the GHC API and perhaps cause other problems. I couldn't find any related issues in this tracker, but I appologize if I missed them.

Specs according to the guidelines:

  * GHC: 7.10.1
  * Platform: Arch Linux x86_64
  * Cabal: 1.22.2.0

Regards, Mattias

Fuuzetsu:
We don't do any processing of the files ourselves which means either c2hs has to be taught about using LINE better in presence of Haddock comments or GHC to do the thing you expect here. Either way, the ticket should probably go into an issue tracker for one of those projects. Personally I'd file it on c2hs and see what they say. If it is GHC that learns to do this and we need to adapt in how we call it, then it becomes our concern.

Let me know if you disagree and still think it's our ticket.

mjakob:
[...]

PR #977, opened 2018-12-12
Use `.hie` files for the Hyperlinker backend

### LINKED ISSUE #496: support inter-*package*  --hyperlinked-source links, currently only work for *intra* package references,
i'm not sure if this is a bug report or a feature request, but it looks like haddock's --hyperlinked-source source files currenlty only support intrapackage references?

cc @hvr @thoughtpolice @mpickering  @bgamari, thoughts?

### LINKED ISSUE #715: Hyperlinker: Hovering a name should show its type
An interesting feature for our hyperlinked source backend. Hovering over a name should show its type:

hover

Implementation should be straight forward. In https://github.com/haskell/haddock/blob/master/haddock-api/src/Haddock/Backends/Hyperlinker/Ast.hs we already collect all the information for each binder which we can use in the renderer.

In case of parameterized functions we could show the original type and the instantiated types on hover. This might give a nice experience when browsing sources.

/cc @harpocrates Interested?

harpocrates:
Now that `.hie` files can be generated by GHC, Haddock should use them in the hyperlinker backend instead of going via the `TypecheckedModule` (which will have to go away for Hi Haddock to work).

Fixes  #496.
Fixes  #715.


harpocrates:
Some bigger TODO items:

 - [ ] `.hie` files shouldn't be put beside `.hs` files
 - [ ] `.hie` files should be generated only when hyperlinked sources are requested 
 - [ ] get to the bottom of `Typeable` warnings the get produced on `base`:

       GHC error in desugarer lookup in Ghci1:
         Can't find interface-file declaration for type constructor or class Typeable
           Probable cause: bug in .hi-boot file, or inconsistent .hi file
           Use -ddump-if-trace to get an idea of which file caused the error

  - [ ] make sure we don't have performance regressions (in particular: if the new type annotation stuff is causing a huge performance regression, maybe it should be behind a flag?) 
  - [ ] accept the output

wz1000:
There are a few more issues about the interaction between the haddock tokenizer and `.hie` files.

- The haddock tokenizer classifies `` `foo` `` and `(+)` as 3 separate tokens, while they are treated as a single token by ghc, so hyperlinking is broken for these.

- The haddock tokenizer doesn't understand `{-# LINE ... #-}` pragmas, so hyperlinking is broken for files that include this pragma(see [this](https://hackage.haskell.org/package/base-4.12.0.0/docs/src/GHC.Clock.html)). This is not a new issue created by this PR, but also affects hyperlinked source generated by haddock without using `.hie` files.

harpocrates:
> There are a few more issues about the interaction between the haddock tokenizer and `.hie` files.
> 
>    * The haddock tokenizer classifies `` `foo` `` and `(+)` as 3 separate tokens, while they are treated as a single token by ghc, so hyperlinking is broken for these.
> 
>    * The haddock tokenizer doesn't understand `{-# LINE ... #-}` pragmas, so hyperlinking is broken for files that include this pragma(see [this](https://hackage.haskell.org/package/base-4.12.0.0/docs/src/GHC.Clock.html)). This is not a new issue created by this PR, but also affects hyperlinked source generated by haddock without using `.hie` files. Relevant issue: #441

Good point for infix identifiers! Haddock uses GHC's tokenizer - the problem is that `` `foo` `` (and `(+)`) is actually three tokens (see https://github.com/ghc/ghc/blob/1ebe8438cd29448291229d5ce23c9f73216e9650/compiler/parser/Parser.y#L3354).

However, I don't think we should let this be blocked by position-altering pragmas. I think that's a more fundamental issue about how GHC conflates positions in files with positions used for error reporting.

wz1000:
@harpocrates Is this for building haddock docs with `--hyperlinked-source`?

With an SSD, I see a slowdown of ~15% compiling ghc stage2 with `-fwrite-ide-info`: https://gist.github.com/wz1000/8678464b4c88be8359d39462ae7f0a90

harpocrates:
Yes this is building full docs with --hyperlinked-source.

Performance won’t necessarily get that much better. We’ll still be calling out to GHC - just in a way that GHC can avoid compiling modules if it already has what it needs. I suspect that when folks do build docs with hyperlinked sources, they’ll probably still need compilation for HIE files (unless those are already produced and up to date - maybe for IDE reasons).

Until the performance for generating HIE files gets better, I don’t think we should enable them by default in Cabal either (the way I think we will want to enable `-haddock` by default).

wz1000:
It would be useful to know whether the bulk of the slowdown in in the `ghc` stage that generates .hie files or in the `haddock` stage that writes hyperlinked sources. Since on my machine, I get <15% slowdown when building stage2 with `fwrite-ide-info`, wherease you get a 50% overall slowdown.

wz1000:
I meant I get a 15% slowdown building `ghc` with `-fwrite-ide-info`, without building any haddock docs at all. It would be good to the disparity is due to

- Using an SSD, since `.hie` files naturally lead to a substantial and largely unavoidable increase in disk IO, which would impact setups with slower storage formats disproportionately.
- The new hyperlinker code added to haddock by this PR being slow
- Some other reason.

harpocrates:
Given the first lines of pretty much all the new `.prof` files, it looks like generating `.hie` files is the slow part. The new hyperlinker code could probably be sped up a bit too though. 

```
        Sat Dec 22 19:54 2018 Time and Allocation Profiling Report  (Final)

           haddock +RTS -p -t_build/stage1/libraries/Cabal/Cabal/build/haddock.t ...


        total time  =       25.36 secs   (25362 ticks @ 1000 us, 1 processor)
        total alloc = 59,578,959,808 bytes  (excludes profiling overheads)

COST CENTRE                    MODULE                                SRC                                                                                      %time %alloc

extract_renamed_stuff          HscMain                               compiler/main/HscMain.hs:(402,1)-(435,18)                                                 25.6   31.3
tc_rn_src_decls                TcRnDriver                            compiler/typecheck/TcRnDriver.hs:(500,4)-(572,7)                                           9.6    6.5
annotate.typ                   Haddock.Backends.Hyperlinker.Renderer utils/haddock/haddock-api/src/Haddock/Backends/Hyperlinker/Renderer.hs:153:5-56            4.3    4.0
spanToNewline                  Haddock.Backends.Hyperlinker.Parser   utils/haddock/haddock-api/src/Haddock/Backends/Hyperlinker/Parser.hs:(126,1)-(156,21)      3.9    3.5
ppHyperlinkedModuleSource.html Haddock.Backends.Hyperlinker          utils/haddock/haddock-api/src/Haddock/Backends/Hyperlinker.hs:80:5-50                      2.9    5.2
solve_loop                     TcInteract                            compiler/typecheck/TcInteract.hs:(231,9)-(235,44)                                          2.7    1.0
...
```

wz1000:
That says 25% of the time is spent in `extract_renamed_stuff`. However, now I doubt whether we need renamed source at all to generate .hie files. Initially there were some issues with some name uniques not matching due to differences introduced by the typechecker, but I believe I should have solved those with the `name_remappping` stuff. 

The traversal of the renamed source enables hyperlinking of imports/exports/annotations/foreign declarations and other stuff that is not in the TypecheckedSource. Maybe there is a better way to extract those specifically, and not extract the rest of AST that is already there in the typechecked source. 

I believe it would be worth testing to see if not including RenamedSource in the hie file generation stuff yields a significant performance improvement.

wz1000:
> Isn’t `extract_renamed_stuff` the top-level function for where HIE files get created and written out (so that entry includes not just time spent traversing the renamed source, but also time spent compacting the AST and writing it to a file)?

Oh, yes, I misremembered. However, the format will not change if we don't traverse the RenamedSource. All that needs to be changed in `mkHieFile` is

```haskell
         flat_asts = concat
           [ tasts
-          , rasts
-          , imps
-          , exps
           ]
```

and we can also avoid setting `keep_rn` to `True` in HscMain when given `-fwrite-ide-info`

harpocrates:
By "format" I don't mean the Haskell types in `HieTypes`, I mean the structures that people expect to be inside `.hie` files.

Would you like to make a GHC patch for this? Then we can post some performance numbers there, as well as check that nobody is relying on the information that is about to be removed from `.hie` files.

mpickering:
There are not that many cost centres defined in the GHC code base so I wouldn't trust the information to precisely point to what is taking up the time.

wz1000:
@harpocrates see https://github.com/wz1000/ghc/tree/hiefile-no-renamed

wz1000:
Parts of the AST that are unaccounted for without using renamed source:
* [ ] Type signatures
* [x] Imports
* [x] Exports
* [ ] Deriving Declarations
* [ ] Fixity declarations
* [ ] Default Declarations
* [x] Foreign declarations
* [ ] Warning declarations
* [ ] Ann Declarations
* [x] Rule Declarations

harpocrates:
@wz1000 Any updates on potential performance improvements from the GHC side?

I'm gonna see if I can optimize the rest of the stuff in `Haddock.Backends.Hyperlinker.Parser`. In particular, I suspect there will be significant performance gains if we transform the preprocessing to work on `StringBuffer` instead of `String`. I expect to get about a 5% improvement from that.

wz1000:
@harpocrates the approach I suggested earlier(not traversing the renamed AST) will not work. Type signatures can be buried arbitrarily deep in the AST, and to hyperlink those, we need the Renamed Source(since `LHsType GhcRn` is simply converted to `Type` by the typechecker, thus losing the structure of the original source).

The other items in my checklist above should be possible to extract without a full traversal of the renamed AST. To see the possible gains, you can benchmark the branch I linked above. If the gains are substantial, we might want to add a flag that enables this behaviour at the cost of information for type signatures.

harpocrates:
I did what I said I would in https://github.com/haskell/haddock/pull/977#issuecomment-452046425 and updated the numbers here: https://docs.google.com/spreadsheets/d/1xmUdUoVqbYtIRYdz5TgWT6rwHsVGjYEgEmMwWClU9-c/edit#gid=0. The Haddock-side gains were more sizeable than I anticipated, but still don't offset the GHC-side cost of generating the `.hie` files.

Take the `Cabal` package as an example (one of the bad cases): instead of being 46.04% slower and allocating 51.58% more, it is now "only" 33.56% slower and allocates "only" 43.41% more.

harpocrates:
Good news. By limiting (for now) which expressions have their types serialized into the HIE files, performance returns to levels that make this mergeable (2-3% percent faster but allocate 2-3% more) without sacrificing any features.

We'll now see if anyone else objects: https://gitlab.haskell.org/ghc/ghc/merge_requests/214.

wz1000:
@harpocrates 2-3% faster than the previous version with .hie files or 2-3% faster than the old way of generating hyperlinked source?

harpocrates:
@wz1000 Than the old way, otherwise it wouldn't really matter, right?

harpocrates:
Actually, with the latest tweaks to that merge request, both allocations and total time drop by a couple percentage points compared to the old way. I do love changes that are net wins in all dimensions: time, allocations, bug fixes, and features! 😀

I intend to accept the new output and merge this as soon as the GHC MR gets merged.

harpocrates:
[Final numbers](https://docs.google.com/spreadsheets/d/1xmUdUoVqbYtIRYdz5TgWT6rwHsVGjYEgEmMwWClU9-c/edit#gid=847080098), in case anyone is interested:

  - 9-10% faster on the sum of boot libs running `--html` and `--hyperlinked-sources`
  - 6-7% less allocations on the sum of boot libs running `--html` and `--hyperlinked-sources`

Thanks to everyone for being patient!

alexbiehl:
Very nice! Thank you sir!

Issue #979, opened 2018-12-13
internal error: extractDecl when selectively reexporting data instance


watashi:
## Env

* ghc and haddock built from ghc master

```
$ uname -a
Linux localhost 4.18.8-arch1-1-ARCH #1 SMP PREEMPT Sat Sep 15 20:34:48 UTC 2018 x86_64 GNU/Linux
$ cabal --version
cabal-install version 2.5.0.0
compiled using version 2.5.0.0 of the Cabal library 
```

## Repro

Here is the minimal repro I constructed, with two packages where pkg2 depends on pkg1:

```haskell
$ cat pkg1/A.hs
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE TypeFamilies #-}
module A
  ( C(..)  -- [1]
  , T(TI)
  , error
  ) where

class C a where
  data T a

instance C Int where
  data T Int = TI { ti :: Int }

$ cat pkg2/B.hs
module B
  ( module A
  ) where

import A hiding (error) -- [2]

$ cabal sandbox init
$ cabal add-source pkg1 pkg2
$ cabal install pkg1 pkg2 --with-ghc ~/ghc/inplace/bin/ghc-stage2 --with-haddock ~/ghc/inplace/bin/haddock --haddock-html --enable-documentation --reinstall
...
haddock: internal error: internal: extractDecl
CallStack (from HasCallStack):
  error, called at src/Haddock/Interface/Create.hs:1116:12 in haddock-api-2.22.0-inplace:Haddock.Interface.Create
...
```

## Issue

I looked into this a little bit and this is what the callstack looks like:

```haskell
mkExportItems _...
  Just [(module A, [C{T}, T{TI}])] -- exportList: without @C(..)@ at [1], this will be [T{T}, T{TI}]
  [T{TI}, C{T}] -- allExports; without @C(..)@ at [1], this will be [T{T, TI}] and haddock works
  _ _
-- we go to the last case of 'lookupExport' (haddock works without @hiding (error)@ at [2]
availExportItem _... T{TI}
-- the decl of 'T' is from another package: 'hiDecl' returns @data family T a :: Type@ (haddock works otherwise)
availExportDecl T{TI} (data family T a :: Type) (_, [(TI, _)])
-- @availExportsDecl avail@ is 'False', so we handle the sub one 'TI'
extractDecl [] TI (data family T a :: Type)
-- name doesn't exist in the decl returned by 
```

[...]

PR #980, opened 2018-12-13
Output better debug infromation on internal error in extractDecl

watashi:
This will make investigation of #979 easier, following the same idea as https://github.com/haskell/haddock/commit/c03dad1c5d4ab7c44234d145ba9c13f17a918201

haddock used to print

```
haddock: internal error: internal: extractDecl
  error, called at src/Haddock/Interface/Create.hs:1116:12 in haddock-api-2.22.0-inplace:Haddock.Interface.Create
```

for the repro in #979 and now prints

```
haddock: panic! (the 'impossible' happened)
  (GHC version 8.7.20181213 for x86_64-unknown-linux):
        extractDecl

Unhandled decl for TI:
    data family T a :: Type
Call stack:
      callStackDoc, called at compiler/utils/Outputable.hs:1159:37 in ghc:Outputable
      pprPanic, called at utils/haddock/haddock-api/src/Haddock/Interface/Create.hs:1116:12 in main:Haddock.Interface.Create
```

### DIFF
--- a/haddock-api/src/Haddock/Interface/Create.hs
+++ b/haddock-api/src/Haddock/Interface/Create.hs
@@ -1112,7 +1112,9 @@ extractDecl declMap name decl
             in case matches of
               [d0] -> extractDecl declMap name (noLoc . InstD noExt $ DataFamInstD noExt d0)
               _ -> error "internal: extractDecl (ClsInstD)"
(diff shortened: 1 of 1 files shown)


harpocrates:
Thanks! #979 looks pretty nasty indeed, so thanks for debugging that too.

PR #983, opened 2018-12-20
Load plugins when compiling each module

### LINKED ISSUE #900: Haddock fails to run plugins with GHC 8.6.1 alpha

harpocrates:
The previous approach in 72d82e52f2a6225686d9668790ac33c1d1743193 was not working as reported in #900.

Fixes #900.

NB: this does _not_ fix the situation for local plugins (which have not historically worked either)

### DIFF
--- a/haddock-api/src/Haddock.hs
+++ b/haddock-api/src/Haddock.hs
@@ -76,7 +76,6 @@ import Packages
 import Panic (handleGhcException)
 import Module
 import FastString
-import qualified DynamicLoading
 
 --------------------------------------------------------------------------------
 -- * Exception handling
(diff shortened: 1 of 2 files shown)


mpickering:
Looks correct to me.

isovector:
Does this exist in any releases? I'm tearing my hair out trying to get haddock to work with plugins.

PR #994, opened 2019-01-07
Add a flag to hide instance lists

Lysxia:
See #698 

Is this the right way to add such a flag? That seems like a lot of lines to touch.

### DIFF
--- a/haddock-api/src/Haddock.hs
+++ b/haddock-api/src/Haddock.hs
@@ -28,6 +28,7 @@ module Haddock (
 import Haddock.Backends.Xhtml
 import Haddock.Backends.Xhtml.Meta
 import Haddock.Backends.Xhtml.Themes (getThemes)
+import Haddock.Backends.Xhtml.Utils
 import Haddock.Backends.LaTeX
 import Haddock.Backends.Hoogle
 import Haddock.Backends.Hyperlinker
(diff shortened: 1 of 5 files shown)


harpocrates:
> Is this the right way to add such a flag? That seems like a lot of lines to touch.

I think this is the right way. I agree that the line diff is unfortunate. One refactor that will be in order (as soon as some of the other larger architectural changes are done) is to package together a bunch of the parameters that get threaded through everywhere (qual options, whether to use unicode, etc.) into one record.

Then, adding a flag like this one would be a matter of extending that record.

harpocrates:
I'm curious, what does this bring us that we don't already get with https://github.com/haskell/haddock/pull/995? Is it just whether instance lists are open or closed by default? If so, I'm a bit wary of

  * introducing yet another command line flag
  * how invasive the line change (esp. for such a relatively small functionality change)

Also, won't https://github.com/haskell/haddock/pull/1007 (once it is no longer WIP) render this obsolete?

Lysxia:
Indeed, the feature overlap is quite perplexing. The main point of this PR (compared to the other(s)) is to have instances folded statically, without relying on Javascript which may happen to be either unavailable or simply slow, nor on any pre-setting on the part of visitors (#1007 still requires a few clicks).

As the various examples from the mailing list discussion (linked in #698) show, there are some modules where instances have no business starting out unfolded, so it seems fair to be able to fix that statically.

For these reasons, I still consider this PR to be the main solution to issue #698.

About your point on line changes, do you have an idea of when the refactoring of options would be appropriate you mentioned earlier would be appropriate to do?

Lysxia:
In the end this feature might be too niche to be worth it. cc @Kleidukos

PR #995, opened 2019-01-08
Add button to toggle all instances

Lysxia:
As discussed in #698 

I'm not a typescript programmer, so tell me if anything could be improved!

---

Possible minor issues, that I wouldn't mind addressing in this PR (so this is currently WIP):

- ~~This uses a global boolean variable to decide whether to collapse/expand. In particular, it ignores whether all instances are already all collapsed or expanded.~~ (Won't fix in this PR. Too much work.)

- ~~The text doesn't reflect the state of that variable (it just says "toggle").~~ (FIXED)

### DIFF
--- a/haddock-api/resources/html/js-src/details-helper.ts
+++ b/haddock-api/resources/html/js-src/details-helper.ts
@@ -99,17 +99,36 @@ function initCollapseToggles() {
   });
 }
 
-var allInstancesToggled = false;
+var allInstancesCollapsed = false;
 
-function toggleAllInstances() {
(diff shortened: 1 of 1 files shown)


Lysxia:
Should I update the minified source too?

harpocrates:
Yes, you should update the minified source. I've tried this out and it is splendid! Look forward to merging this.

harpocrates:
Superseded by #1007, which was merged in 59843f9e3d222901421a92fff793a1e031f38f65.

PR #997, opened 2019-01-15
Make a fixture of weird parsing of lists

phadej:
Add list blocks fixtures  …
The second example is interesting.
If there's a list directly after the header, and that list has
deeper structure, the parser is confused: It finds two lists:
- One with the first nested element,
- everything after it

I'm not trying to fix this, as I'm not even sure this is a bug,
and not a feature.


harpocrates:
Thanks!

This sorts of thing is nice to document.

PR #1005, opened 2019-01-25
Fix #1004 with a pinch of dropForAlls

### LINKED ISSUE #1004: Haddock renders imported data types with polymorphic return kinds incorrectly
(Originally discovered in well-typed/generics-sop#92).

Run Haddock on this file:

```haskell
module Bug (Product(..)) where

import Data.Functor.Product
```

And you'll get this:

haddock

That return kind is entirely wrong, given that we're also pretty-printing the type variable binders for `Product` itself. I was under the false impression that I had already fixed this issue in #702, but now that I look more carefully at the code that I added:

https://github.com/haskell/haddock/blob/44169f4b1907e34fdf8ff84cf8b7509b1dfcaf55/haddock-api/src/Haddock/Convert.hs#L275-L280

It turns out my solution wasn't sufficiently general. We're using `splitFunTys` here, but that won't work in the case where the return kind is headed by a `forall` (such as in the example above). A more robust solution is to use `dropForAlls` followed by `splitFunTys`.

Patch incoming.

RyanGlScott:
#1004 occurred because `splitFunTys` wouldn't work on a data type return kind that was headed by a `forall`, so let's drop those `forall`s before attempting to split it.

### DIFF
--- a/haddock-api/src/Haddock/Convert.hs
+++ b/haddock-api/src/Haddock/Convert.hs
@@ -274,7 +274,7 @@ synifyTyCon coax tc
 -- which this function obtains.
 synifyDataTyConReturnKind :: TyCon -> Maybe (LHsKind GhcRn)
 synifyDataTyConReturnKind tc
-  = case splitFunTys (tyConKind tc) of
+  = case splitFunTys (dropForAlls (tyConKind tc)) of
       (_, ret_kind)
         | isLiftedTypeKind ret_kind -> Nothing -- Don't bother displaying :: *
         | otherwise                 -> Just (synifyKindSig ret_kind)
--- a/html-test/ref/Bug1004.html
+++ b/html-test/ref/Bug1004.html
@@ -1,0 +1,2072 @@
+<html xmlns="http://www.w3.org/1999/xhtml"
+><head
+  ><meta http-equiv="Content-Type" content="text/html; charset=UTF-8"
+     /><meta name="viewport" content="width=device-width, initial-scale=1"
+     /><title
+    >Bug1004</title
+    ><link href="#" rel="stylesheet" type="text/css" title="Linuwial"
+     /><link rel="stylesheet" type="text/css" href="#"
+     /><link rel="stylesheet" type="text/css" href="#"
+     /><script src="haddock-bundle.min.js" async="async" type="text/javascript"
+    ></script
+    ><script type="text/x-mathjax-config"
+    >MathJax.Hub.Config({ tex2jax: { processClass: &quot;mathjax&quot;, ignoreClass: &quot;.*&quot; } });</script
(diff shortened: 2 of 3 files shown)


harpocrates:
Do you mind perhaps copy-pasting the data declaration for `Product` directly into the test case (instead of re-exporting it)? Else we'll get a test case that needs to be re-accepted every time someone changes a docstring on `Product` (or any of its many instances) and we end up committing into the source tree a pretty huge output.

Thanks for the patch in any case, and nice catch! 👍

RyanGlScott:
The problem is that `synifyDataTyConReturnKind` is a code path that is only ever reached when Haddocking things that are loaded from external packages. (If you just inlined the definition of `Product`, the bug doesn't surface in the first place.) `Product` was a convenient choice since it's in `base`, which is by far the most readily available external package.

Is there another way to test something from an external package that would be more stable w.r.t. changes in Haddocks? I'm not too familiar with the Haddock test suite, so if there is already a prebuilt solution for this, I'm not aware of it.

harpocrates:
Oops! Didn't mean to close this.

You are totally right, this code path isn't currently exercised by Haddock except for stuff form external packages (and instances). It will be once Hi Haddock is merged though... Let me think some more about how to best add a test case.

harpocrates:
@RyanGlScott Handling some of the merge conflicts between  8.6 and 8.8 and I just realized that this may already have been fixed by https://github.com/haskell/haddock/commit/aa9644fc4179de044557438f2bd3003642750a97#diff-6e1cbdc9e1a19780e3e767df4a17b7bdR323. Specifically, is there any case where the change this PR added (copied here for your convenience)

```haskell
synifyDataTyConReturnKind :: TyCon -> Maybe (LHsKind GhcRn)
synifyDataTyConReturnKind tc
   = case splitFunTys (dropForAlls (tyConKind tc)) of
       (_, ret_kind)
         | isLiftedTypeKind ret_kind -> Nothing -- Don't bother displaying :: *
         | otherwise                 -> Just (synifyKindSig ret_kind)
```

does better than the change in  `ghc-8.8` (also copied here for your convenience)

```haskell
synifyDataTyConReturnKind :: TyCon -> Maybe (LHsKind GhcRn)
synifyDataTyConReturnKind tc
  | isLiftedTypeKind ret_kind = Nothing -- Don't bother displaying :: *
  | otherwise                 = Just (synifyKindSig ret_kind)
  where ret_kind = tyConResKind tc
```

?

The test case you added (`Bug1004`) does not indicate any, but I figured I'd ask...

RyanGlScott:
Ah, I didn't think about using `tyConResKind`—that's _way_ cleaner. Assuming that `Bug1004` still passes with the `tyConResKind` formulation of `synifyDataTyConReturnKind`, I'd favor using that over the formulation in this PR.

PR #1007, opened 2019-01-26
Allow users to expand or collapse all instances by default

Lysxia:
WIP!

See https://github.com/haskell/haddock/issues/698#issuecomment-453656245

While #994 allows collapsing instances in the generated HTML, it still requires opt-in on the part of package developers. This PR allows browsers to override that default as well, without having to press the button introduced in #995 every time.

Builds on top of #995.

Ping @ntc2 

---

One additional change is the use of `localStorage` instead of cookies to store client state (which get unnecessarily sent with subsequent HTTP requests). There are still remaining usages of cookies that I haven't looked into. (May be a separate issue)

---

Status

The code is functional, but still a bit messy. Some clean up is in order. I cobbled together `preferences.tsx` by copying bits of `quick-jump.tsx`, so comments and suggestions over there would be highly appreciated.

The biggest missing piece is to fix the CSS for the preferences menu. Is it sufficient to fix the Linuwial theme only, or do Classic and Ocean also need to be modified? I could also use a lot of help here, it would probably be most efficient if someone else could take this on. Otherwise I can do it however slowly; is there a particular location in the `.css` file(s) where the new rules should go?


ntc2:
This sounds great, thanks! I haven't tried building it yet, but it sounds like what I was hoping for. I don't really know anything about CSS, so I probably can't help with that, but I'll take a look if I try building this patch ...

ntc2:
@Lysxia I just tried this out and it is indeed exactly what I was hoping for, thanks!!!

I haven't made any progress on the CSS yet ...

ntc2:
@Lysxia I still haven't tried tweaking the CSS, but I did try tweaking the UI a bit. My concerns with the old UI were that

- if some instances are manually expanded/collapsed, then having both the "expand all" and "collapse all" buttons seems desirable.
- adding two entries to the menu at the top of the page seems like a lot, and if we have both buttons all the time that's tree new entries.
- the option to choose "Default"  for "Default state of instance lists" sounded confusing to me. 

I've attempted to address these concerns by replace the new "Preferences" menu with a more specific "Instances" menu, moving the expand/collapse buttons into this menu, and replacing the three "Default state of instance lists" radio choices with two mutually exclusive checkboxes; when neither box is checked you get the "Default" behavior.

Screenshots:

Old UI:

instances-old-ui

New UI:

instances-new-ui

Please let me know what you think.

I still plan to try tweaking the CSS the next time I work on this.

Lysxia:
I like your suggestions! Thanks @ntc2 !

harpocrates:
Although I haven't tried this out, the idea sounds great. Using `localStorage` seems fine.

> The biggest missing piece is to fix the CSS for the preferences menu. Is it sufficient to fix the Linuwial theme only, or do Classic and Ocean also need to be modified? I could also use a lot of help here, it would probably be most efficient if someone else could take this on. Otherwise I can do it however slowly; is there a particular location in the .css file(s) where the new rules should go?

Fixing Linuwial will be enough. The other themes are only still around for the sake of placating folks who may complain about the new theme. I don't think there is any particular location in the .css file where the new rules should go.

Lysxia:
Thanks!

If we follow @ntc2's suggestions (which seems likely), this will subsume #995.

ntc2:
@Lysxia I pushed my changes on top of this PR here: https://github.com/ntc2/haddock/tree/pr-1007-user-collapse-setting. Assuming you still like this approach, can you pull my branch into this PR?

In addition to the changes I summarized above, I did a little styling of the new menu. It's not amazing, but a little better than the default I think:

instances-styled-new-ui

Basically, I added some padding, changes the colors to match the Linuwial theme, and styled the buttons accordingly.

Lysxia:
Thanks @ntc2! This looks good :)

ntc2:
If @harpocrates is OK with this approach, then PR #995 can be closed.

harpocrates:
I'm happy with this approach! I also really like that it doesn't introduce another command line option.

Should I understand this is ready for a final review? If so, it might be one of the last commits to make it into the release accompanying GHC 8.8!

ntc2:
@harpocrates this is essentially ready for review.

However, if PR #994 (add command line flag to change default collapse/expand state per module) doesn't get merged, then a small simplification of this PR would be in order: the UI in this PR currently provides three possible choices for the default collapse/expand state of instances:

1. collapse by default: selecting by checking the "collapse by default" box.

2. expand by default: selecting by checking the "expand by default" box.

3. no default / use the underlying module default: selected by not checking either box.

If PR #994 gets dropped, then Choice 3 doesn't make much sense, since it becomes equivalent to Choice 2. if the only possibilities are Choice 1 and Choice 2, then the UI for this choice could be simplified to a single box "collapse by default", which defaults to Choice 2 by being unchecked by default.

Also, if "collapse by default" ever became the Haddock default (what Issue #698 proposed and discussion on PR #823 decided against) we would just toggle the default state of this check box. So, the simpler UI is also future proof in that sense.

harpocrates:
@ntc2 @Lysxia Let's do this, but without #994. If it turns out folks love having things collapsed by instances default, it'll be a straightforward change to make that the default. Could one of you make the changes Nathan suggests in https://github.com/haskell/haddock/pull/1007#issuecomment-466289259?

The sooner the better (perhaps either today or tomorrow?) - I'm hoping we can squeeze this in as the final change before the 8.8 release!

ntc2:
@harpocrates @Lysxia OK,  I'll make the changes I suggested in https://github.com/haskell/haddock/pull/1007#issuecomment-466289259 now ...

Lysxia:
Awesome, thanks @ntc2 !

ntc2:
@harpocrates @Lysxia I got a little carried away, but I just pushed my updates to https://github.com/ntc2/haddock/tree/pr-1007-user-collapse-setting. I don't have write access to this PR, so one of you will have to merge my new changes into this PR.

See the detailed commit messages for a description of the new commits and let me know if you have any questions.

harpocrates:
@ntc2 Thanks a lot for doing this. I've tried out your changes locally and am happy with the UI. There appear to be some CSS issues:

  * on narrow screens (such as phones or tablets) the dropdown obscures the rest of the menu
  * it would be nice to make the style switcher dropdown (enabled with ` --built-in-themes`) consistent with the instance dropdown

However, neither of these issues is a show blocker (if there's time and no one gets to it before me, I might fiddle with the CSS too).

I'm going to review the typescript changes then merge tomorrow evening at latest.

Lysxia:
@ntc2 I've added you as a collaborator so you can push to this branch!

ntc2:
@Lysxia thanks, I just added my commits from yesterday to this PR.

@harpocrates I won't have a chance to look at the CSS again before this weekend. Like @Lysxia , I don't know much about CSS, so if you do and you have time it would probably be better for you to try improving the CSS. Re the Style Switcher, I haven't tried it / wasn't aware of it. The Instances dropdown is loosely based on the Quick Jump drop down, but I moved it to the right instead of center and that could be related to small screens problem you observed. If one of the inconsistencies you're referring to is the three vertical dots in the Instances dropdown's title, that's just a unicode character in the HTML generated in details-helper.tsx.

harpocrates:
@ntc2 @Lysxia I pushed some small tweaks to the CSS and TS basically trying to keep the style switcher, quickjump, and instance preference menu items as consistent as possible (in appearance, behaviour, and code).

The problem with small screens remains, but at least it is consistent.

I'd like to merge this as soon as possible. Any objections or last minute changes?

ntc2:
@harpocrates No objections or changes here. (I haven't actually built it yet to see how you changed the UI, but I'm not worried about it.)

Lysxia:
Does that last quick hack address the narrow screen issue? Otherwise I don't have anything to add! Shall we close the other PRs?

haddock-api/resources/html/js-src/details-helper.tsx
```diff
@@ -0,0 +1,464 @@
...
+// elements themselves. Note that this covers both <details> elements
+// that list instances -- what the above explained UI and logic is
+// concerned with -- and details about individual instances themselves
+// -- which the above is not concerend with. The association includes
+// adding event listeners that change CSS classes back and forth
+// between "expander" and "collapser"; these classes determine whether
+// an element is adorned with a right arrow ("expander") or a down
+// arrow ("collapser"). I don't understand why we don't directly use
+// the the HTML <summary> element type to allow the <details> elements
+// to be directly clickable.
```
Lysxia: This is a really nice comment!

harpocrates:
>  Does that last quick hack address the narrow screen issue? Otherwise I don't have anything to add!

That solves the vertical spacing issue, but not the horizontal overflow one. Nonetheless an improvement, so thanks! 😄 

> Shall we close the other PRs?

Yeah, I'm gonna go ahead and merge this now.

Issue #859, opened 2018-06-14
Avoiding CI breakage on ghc-head?


harpocrates:
In the time I've been following Haddock development, it roughly follows a cycle

  1. there is a branch for the latest stable GHC (currently `ghc-8.4`) and a branch for GHC HEAD (`ghc-head`) and most changes go into the former (since few people actually have HEAD available)
  2. when GHC release candidates becomes available, there is some scrambling to try to get everything in the stable branch also in the head branch
  3. when that version is actually released, a new branch is created

Point 1 is really where the most time is spent. Unfortunately, stuff gets merged into `ghc-head` without tests passing all the time. This is sometimes because the test suite cannot be built, other times because matching GHC changes have not made their way into Herbert's PPA yet. That means that by the time point 2 rolls around, we are stuck accepting a bunch of test output changes all at once (and the odds of something incorrect being let in are much higher)!

The only simpler way I can think of fixing this situation is to change our CI script to install Cabal/GHC from tarballs (so we can take advantage of the nightly-built tarballs). That way, we can keep `ghc-head` mostly green even if features coming into Haddock are going to depend on a very recent GHC. I'm already doing something similar [here](https://github.com/harpocrates/inline-rust/blob/8bf7e3c811006a851f2de733442bfa669a75d20d/.travis.yml).

In any case, I'd be interesting in people's opinions (in particular @hvr).

hvr:
We're fully aware of this workflow dilemma. But ghc-head suffers from additional issues as Hackage generally can't (nor should it!) be expected to be compatible with GHC HEAD. So it's quite likely that ghc-head will fail CI due to accidental reasons.

Fwiw, the easiest way to avoid having to resort to the rather inconvenient use of nightly tarballs would be if I simply increase the build-rate of my ghc-head PPA again to its original daily frequency.

harpocrates:
> Fwiw, the easiest way to avoid having to resort to the rather inconvenient use of nightly tarballs would be if I simply increase the build-rate of my ghc-head PPA again to its original daily frequency.

At its core, that functionally equivalent to what I was proposing - minus the inconvenience of dealing with tarballs. It would be nice if we also had an Appeyor CI (I introduced a windows-only regression not too far back that could've been easily avoided). Do you know if chocolatey offers GHC HEAD anywhere?

harpocrates:
I'm hitting this again (CI is broken on ghc-head).

[...]

PR #1012, opened 2019-01-31
Fix CI on `ghc-head`

harpocrates:
This unbreaks CI on `ghc-head`. The idea follows #859. There's a couple reasons for doing this:

  - We'll always be up to date on GHC changes. Worse case scenario, we wait a day and manually restart CI, but CI should _never_ be red for more than a day.
  - It'll be trivial to point to a custom GHC artifact, so Haddock branches which need custom GHC branches (like `wip/hi-haddock`) can continue to run CI

Making CI green also forced me to fix some breakage that's snuck in since stuff was last green.


mpickering:
You could simplify the .travis script by using `language:nix` instead I think if that's a path you want to go down.

harpocrates:
> You could simplify the .travis script by using `language:nix` instead I think if that's a path you want to go down.

Is there a polished way in Nix to install the validate artifacts and add the GHC HEAD overlay?  That sort of thing would be quite nice indeed.  Could you expand more on what you had in mind?

mpickering:
I guess you would just want nix to install ghcHEAD and cabal but leave cabal to manage all haskell dependencies.

Setting up that environment can be managed in one command:

```
nix run -f https://github.com/mpickering/ghc-artefact-nix/archive/master.tar.gz ghcHEAD cabal-install
```

Perhaps you could also keep most of the current script the same and install the dependencies globally to save yourself mucking around unpacking them.

```
nix-env --dry-run -i -f https://github.com/mpickering/ghc-artefact-nix/archive/master.tar.gz ghcHEAD cabal-install
```

harpocrates:
for the record: I would prefer some clean `nix` solution here (especially if it worked uniformly for GHC HEAD and non GHC HEAD), but there are other yaks to shave right now.

PR #1014, opened 2019-02-02
Normalise LICENSE text w/ cabal's BSD2 template

hvr:
Also, correct the `.cabal` files to advertise `BSD2` instead
of the incorrect `BSD3` license.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,23 +1,26 @@
-Copyright 2002-2010, Simon Marlow.  All rights reserved.
+Copyright (c) 2002-2010, Simon Marlow
+All rights reserved.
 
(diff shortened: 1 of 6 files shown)


hvr:
@simonmar Since you're the only explicitly named (c)-holder in the previous LICENSE text, I'm hoping you could help clarify what the licensing intent was; i.e. whether it's in fact the 2-clause BSD or something else.

simonmar:
This looks fine. I don't recall why the license differs from the standard one, no objection to syncing it.

simonmar:
Looks like CI is failing for the same reason that I recently fixed in Alex - cabal has changed where it puts the results of sdist for some reason.

PR #1016, opened 2019-02-02
Fix #1015 with dataConUserTyVars

### LINKED ISSUE #1015: Redundant quantification when rendering GADT constructors from external packages
(Yet another issue noticed in well-typed/generics-sop#92.)

Run Haddock on this code:

```haskell
module Bug ((:~:)(..)) where

import Data.Type.Equality
```

And you'll get this:

haddock

Notice that the type signature for `Refl` quantifies `b`, which is completely redundant! The culprit is these lines of code:

https://github.com/haskell/haddock/blob/fd56ac4d031963ecc137a5027d800bd8ed588264/haddock-api/src/Haddock/Convert.hs#L337-L342

This is constructing a type signature for a GADT constructor by displaying all of the universally quantified type variables following by the existentially quantified type variables. But this is not always what the user has written, as shown in the example above. We really ought to be using `dataConUserTyVars` instead, which takes this information into account.

Patch incoming.

RyanGlScott:
The central trick in this patch is to use `dataConUserTyVars` instead of `univ_tvs ++ ex_tvs`, which displays the `forall`s in a GADT constructor in a way that's more faithful to how the user originally wrote it. With these changes, the example from #1015 now renders as:

haddock

Along the way, I simplified some code that tried to calculate where a data constructor should use GADT syntax in an _ad hoc_ fashion. Luckily, this information is cached as `isGadtSyntaxTyCon`, so I just used that instead.

Thankfully, there's already a test case in `html-test` that stress-tests this code path, so I didn't need to add the test case from #1015 (since it relies on code from `base` which might change in the future).


harpocrates:
So this is something already fixed in the Hi Haddock branch at a53c131e9bf95db0180efdd560f8f27caa81363e. There are a bunch of fixes to `Convert` in that branch which exist nowhere else. Rather than re-fix them in `ghc-head` and face hell when the time comes to rebase that branch, I think the right thing to do would be split them out of `wip/hi-haddock` (since that's probably not going to get merged in time for 8.8).

If @sjakobi isn't planning on working on Hi Haddock this weekend, I could try to do that now. Thoughts @sjakobi?

RyanGlScott:
Looking at a53c131e9bf95db0180efdd560f8f27caa81363e, it looks like that commit tackles the problem in a much different way. Intead of using `dataConUserTyVars`, it simply hardcodes `con_forall` to `False`! If I'm reading that correctly, that means that Haddock will _never_ render the `forall`s in a GADT constructor's type signature, which feels rather strange.

harpocrates:
You are right, this is the right way to fix things! I just want to get ahead of the Hi Haddock merge conflicts before merging this.

harpocrates:
@RyanGlScott I've split out of `wip/hi-haddock` and into `ghc-head` the changes I was worried would conflict. Do you mind rebasing this onto `ghc-head`?

RyanGlScott:
I take it, then, that you don't want me to target this PR against `ghc-8.6` (as it is currently)?

harpocrates:
Yep, that’s right. Sorry for all of the hassle :(

RyanGlScott:
That's fine—I'm just wondering if this could be cherry-picked into `ghc-8.6`, since @kosmikus was asking for this in https://github.com/well-typed/generics-sop/issues/92#issuecomment-459971779.

harpocrates:
Oh I see. Sigh.

I guess we can merge this now in `ghc-8.6` and deal with the conflicts when next merging into `ghc-head`. Do you know if there is even a new Haddock release planned for the 8.6 release? (Is there even another 8.6 release planned?) I haven't heard anything from Ben.

harpocrates:
> I was under the impression that Haddock releases occurred independently of GHC

That's not the case. Since most people get Haddock from their GHC installation (NB: the binary-dists that get released include a Haddock executable), I've been syncing releases up with GHC releases. I guess we could release more often, but I doubt people would remember to install a newer version of Haddock. And even if they did:

  1. Cabal/Stack have some magic around trying to guess the right Haddock executable for `cabal/stack haddock`, which I don't think would pick up the newer executables (unless explicitly directed to do so). I haven't tested this...
  2. The Haddock distributed with GHC is built under slightly different circumstances (for now at least, see `flag(in-ghc-tree)`), so it's nice to minimize the "flavours" of Haddock executables floating around in the wild. I hope to get rid of this once the Make system disappears and only Hadrian remains.

> ...and that the ghc-8.6 branch was the main upstream for that release cycle (while ghc-head tracks a separate development cycle that's geared specifically to GHC HEAD's needs). Moreover, I was imagining that, at periodic intervals, there would be a mass merging of commits from one branch to the other.

This part is spot on. We regularly merge branches for older release cycles into ones for newer release cycles. This sort of merge is generally not very pleasant though.

---

If you have any ideas for improving this state of affairs, I'm all ears. 😄

RyanGlScott:
It looks like the `haddock` repo disallows force-pushing (even for branches), so rebasing this as-is proves to be trickier than I thought. I think I'll just opt to close this and open a fresh PR against `ghc-head`.

RyanGlScott:
Closing in favor of #1022.

PR #1018, opened 2019-02-02
Update README instructions for Stack

ntc2:
No need to `stack install` Haddock to test it. Indeed, `stack install` changes the `haddock` on user's `PATH` if `~/.local/bin` is on user's `PATH` which may not be desirable when hacking on Haddock.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,9 +82,9 @@ cabal test
 
 ```bash
 stack init
-stack install
+stack build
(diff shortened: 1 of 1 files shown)


harpocrates:
Right, `stack install` just for the purpose of running the testsuite is definitely not a good idea.

Thanks for the patch!

PR #1019, opened 2019-02-03
Remove `Documentation.Haddock.Utf8`

harpocrates:
The circumstances under which this module appeared are completely gone.
The Hyperlinker backend no longer needs this module (it uses the more
efficient `Encoding` module from `ghc`).

Why no deprecation? Because this module really shouldn't exist!

  - It isn't used in `haddock-library`/`haddock-api` anymore
  - It was copy pasted directly from `utf8-string`
  - Folks seeking a boot-lib only solution can use `ghc`'s `Encoding`


haddock-library/haddock-library.cabal
```diff
@@ -45,7 +45,6 @@ library
     Documentation.Haddock.Markup
     Documentation.Haddock.Parser
     Documentation.Haddock.Types
-    Documentation.Haddock.Utf8
```
hvr: fyi, needs a major ver bump!

hvr:
PS:

> Folks seeking a boot-lib only solution can use ghc's Encoding

Note that `text` is a boot-lib now too :-)

PR #1029, opened 2019-02-15
Match GHC changes for T16185

harpocrates:
`FunTy` now has an `AnonArgFlag` that indicates whether the arrow is
a `t1 => t2` or `t1 -> t2`.

This commit shouldn't change any functionality in Haddock.


harpocrates:
Superseded by #1031

Issue #1032, opened 2019-02-25
New color scheme has low contrast.


phadej:
https://github.com/haskell/hackage-server/issues/811

The background of color block is too close to the default white:

![screenshot from 2019-02-24 21-07-17](https://user-images.githubusercontent.com/51087/53319673-dbe93d80-38db-11e9-9edf-aa38f44262c6.png)

The bullet arrow is too subtle, could it be filled?

![screenshot from 2019-02-24 21-15-14](https://user-images.githubusercontent.com/51087/53319686-e4417880-38db-11e9-8c46-f3b929a8a971.png)

Also the `body` (virtually everything "black") color text is `#333`, which is IMO too light, given the font-family is very light on itself.

harpocrates:
@phadej These seem like reasonable requests. Perhaps you'd like to take a stab at fixing this?

Note also that there is not very little time left until the 8.8 release!

gbaz:
Here's how it looks with the body to #111 and the background (which should be set for at least pre, .src, and #contents-list) to #eee.

![image](https://user-images.githubusercontent.com/1125849/53349385-f8d93d00-38ea-11e9-9068-4f6ac0612898.png)

Note that the bullet is a unicode literal. It will look different depending on the fonts on the system. So it comes in for me (os x) like this:

![image](https://user-images.githubusercontent.com/1125849/53349701-ae0bf500-38eb-11e9-9c98-d1444a6f7068.png)

Changing it from "WHITE RIGHT-POINTING SMALL TRIANGLE" to "BLACK RIGHT-POINTING SMALL TRIANGLE" makes it come in like this:

![image](https://user-images.githubusercontent.com/1125849/53349817-edd2dc80-38eb-11e9-83e4-d7bd53ddcb4f.png)

I actually prefer the first one, in this case.

phadej:
@gbaz note that the triangle is different on my machine/screenshot, it depends on the font.

harpocrates:
@gbaz @phadej Just a head's up: if you'd like this to be in the coming release, please open a PR! Otherwise, this is just going to slip by...

gbaz:
should be resolved by #1037

phadej:
Thanks all for discussion!

Issue #444, opened 2015-09-13
Missing instances


dmwit:
The haddocks here don't list a `Fractional` instance for `Double` or `Float`:

http://hackage.haskell.org/package/base-4.8.1.0/docs/Prelude.html#t:Fractional

However, I know there was one, because immediately below that we have `class Fractional a => Floating a` and it *does* list `Floating` instances for `Double` and `Float`.

I'm actually not sure whether this is a haddock bug or a bug elsewhere, but thought it might be worth mentioning.

Fuuzetsu:
This is just like #372 so closing as duplicate.

Issue #333, opened 2014-10-17
Strange pattern synonym issue


ekmett:
This patch

https://github.com/ekmett/sdl2/commit/f9dc8f351173e9d014ebbfc7b64e75e2f89fd5bf

took a package from building and haddocking just fine to having `cabal haddock` crash with

```
panic! (the 'impossible' happened) (GHC version 7.8.3 for x86_64-apple-darwin): 
  mkUsage 
    <<details unavailable>>
```

I tried it with 2.14.3 and with the latest version on hackage.

PR #1037, opened 2019-02-28
increase contrast of linuwal theme

gbaz:
Simplest possible patch to improve the contrast as requested in https://github.com/haskell/haddock/issues/1032

There might be better solutions, but I think this is modest, straightforward, and should help. If someone wants to propose something else or further tweaks before the cutoff I'm open, but wanted to have at least this in the pipeline.

### DIFF
--- a/haddock-api/resources/html/Linuwial.std-theme/linuwial.css
+++ b/haddock-api/resources/html/Linuwial.std-theme/linuwial.css
@@ -11,7 +11,7 @@ html {
 
 body {
   background: #fefefe;
-  color: #333;
+  color: #111;
   text-align: left;
   min-height: 100vh;
   position: relative;
--- a/haddock-api/resources/html/Linuwial.std-theme/linuwial.css
+++ b/haddock-api/resources/html/Linuwial.std-theme/linuwial.css
@@ -11,7 +11,7 @@ html {
 
 body {
   background: #fefefe;
-  color: #333;
+  color: #111;
   text-align: left;
   min-height: 100vh;
   position: relative;
@@ -388,7 +388,7 @@ summary {
 pre {
   padding: 0.5rem 1rem;
   margin: 1em 0 0 0;
-  background-color: #f7f7f7;
+  background-color: #eee;
   overflow: auto;
 }
 
@@ -408,7 +408,7 @@ blockquote {
 }
 
 .src {
-  background: #f4f4f4;
+  background: #eee;
   padding: 0.2em 0.5em;
 }
 
@@ -533,7 +533,7 @@ div#style-menu-holder {
 }
 
 #contents-list {
-  background:  #f7f7f7;
+  background:  #eee;
   padding: 1em;
   margin: 0;
 }
--- a/haddock-api/resources/html/Linuwial.std-theme/linuwial.css
+++ b/haddock-api/resources/html/Linuwial.std-theme/linuwial.css
@@ -11,7 +11,7 @@ html {
 
 body {
   background: #fefefe;
-  color: #111;
+  color: #333;
   text-align: left;
   min-height: 100vh;
   position: relative;
@@ -234,7 +234,7 @@ Display the package name on top of the menu links and center both elements:
  */
 
  body, button {
-   font: 400 15px/1.4 'PT Sans',
+   font: 400 14px/1.4 'PT Sans',
      /* Fallback Font Stack */
      -apple-system,
 	  BlinkMacSystemFont,
@@ -388,8 +388,10 @@ summary {
 pre {
   padding: 0.5rem 1rem;
   margin: 1em 0 0 0;
-  background-color: #eee;
+  background-color: #f7f7f7;
   overflow: auto;
+  border: 1px solid #ddd;
+  border-radius: 0.3em;
 }
 
 pre + p {
@@ -408,7 +410,7 @@ blockquote {
 }
 
 .src {
-  background: #eee;
+  background: #f2f2f2;
   padding: 0.2em 0.5em;
 }
 
@@ -533,7 +535,7 @@ div#style-menu-holder {
 }
 
 #contents-list {
-  background:  #eee;
+  background: #f4f4f4;
   padding: 1em;
   margin: 0;
 }


NunoAlexandre:
I am not in favour of this PR. `#333` in (pretty much) white background is more than enough, and darken `src`, `pre`, etc, brings no benefit whatsoever. I.e, even if we'd take the special background from those elements, it would make no functional difference as it's not a necessity to start with.

I'd focus on the purple color pallete to improve the contract instead.

In any case, the #eee makes the code blocks ugly imo, and decreases the contrast with its content, where the contrast is actually functionally important.

phadej:
> darken `src`, `pre`, etc, brings no benefit whatsoever

the problem is that `#f4f4f4` is too close to white. Either we remove background color, or make it clearly distinguishable from white.

There is no time to "focus". Some change have to be done. I'm :+1: to this change, it might make things uglier, but it makes things more accessible.

I hope that @harpocrates will make a decision before cutting a release for GHC-8.8.

NunoAlexandre:
@phadej 

See how you make my point all alone:

> the problem is that `#f4f4f4` is too close to white. Either we remove background colour or make it clearly distinguishable from white.

> but it makes things more accessible.

Why does it need to be "clearly distinguishable from white", when removing its background colour is also a solution?

And in which objective way does this make things more accessible? The background of those elements we enumerated makes no real difference, they are more of a cosmetic detail than anything. And so I see no reason to make it ugly.

This is bikeshedding, it's not solving anything imho. Making changes just because "some changes have to be done" will make the theme inedible again in a year. That's my 2 cents, which I mean to give constructively. Up to you, guys.

phadej:
@NunoAlexandre I don't care whether it's made darker or removed, I only ask that something is done before the release is cut. The subtly almost white background forces me to squint: "is there something or not".

phadej:
FWIW, I'd prefer clean design, without too much "cosmetic" details. I look at the haddock documentation hours per day, it should be easy for eyes. For example Stackage theme is equally terrible:
look at https://www.stackage.org/haddock/lts-13.9/sop-core-0.4.0.0/Data-SOP.html

But there, at least instances' list background is darker (than code block background) and instances are easier to read.

cartazio:
https://en.m.wikipedia.org/wiki/Color_space might help this discussion

Human perception works best with multiple dimensions of contrast. The concern here is accessibility / usability. 

The theme seems to be a grayscale of blues.  

Human vision is actually weakest at detecting subtle blue contrasts.  That’s even ignoring user color temperature settings — subtle blue differences will get destroyed by night shift mode on many platforms. 

 differences a la brightness or grayscale work better when they are also differences in a human perception weighted maximally separated points in rgb space when transformed to reflect human perception. 

I do agree that if this is done carelessly that’s garish as hell, but then everything is easy to see and it becomes a conversation about toning it down rather than “I can’t seeeeeee”

To be clear: I love that this stuff is getting some love.  We’re just concerned that some users have different flavors of cheap lcds, ocular health or night shift or tired but coding with haddoc references and will hit usability issues with the new theme. 

We want the best new themes possible and we wanna help make sure any release there of goes well

harpocrates:
@NunoAlexandre The point is not to start bikeshedding every last detail - I'm sorry if it comes off as such. I'm personally very grateful for everything you've done (especially given how long and drawn out the process was).

The original point, raised by @hvr off-thread, was: on less nice and older screens, `#f4f4f4` is almost impossible to distinguish from white. This is purely a functional problem. I don't agree with @phadej that removing the background would solve the problem either.

@NunoAlexandre from a design perspective, what do you think would be the best way forward? A darker background, a more colored background, or perhaps even a border could all fix this... I'll also be playing around with this.

----

Perhaps a darker border would help?

```css
pre {
  padding: 0.5rem 1rem;
  margin: 1em 0 0 0;
  background-color: #f7f7f7;
  overflow: auto;
  border: 1px solid #ddd;
  border-radius: 0.3em;
}
```

screen shot 2019-02-28 at 7 01 08 am

haddock-api/resources/html/Linuwial.std-theme/linuwial.css
```diff
@@ -408,7 +408,7 @@ blockquote {
 }
 
 .src {
-  background: #f4f4f4;
+  background: #eee;
```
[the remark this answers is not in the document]
gbaz: My take is that having one be #f7f7f7 and one be #f4f4f4 was a bit weird to begin with. Those sorts of subtle gradient distinctions are hard for users to pick up on anyway, and I thought it was clearer to take care of that as part of the PR. If you want to send the `contents-list` background to something a bit different, go for it! My taste isn't the best in these things -- I always go for simpler rather than subtler :-)

NunoAlexandre:
@harpocrates 

> The original point, raised by @hvr off-thread, was: on less nice and older screens, `#f4f4f4` is almost impossible to distinguish from white. This is purely a functional problem. I don't agree with @phadej that removing the background would solve the problem either.

Thanks for explaining the context!

> @NunoAlexandre from a design perspective, what do you think would be the best way forward? A darker background, a more colored background, or perhaps even a border could all fix this... I'll also be playing around with this.

I quite like your suggestion above! Much better than setting the background colour to `#eee`, imo. I'd maybe make the border colour closer to `#eee`, but it looks nice.

harpocrates:
@gbaz Do you mind if I push some changes directly to this branch? Then I'll post some screenshots and we can hopefully finish this off.

gbaz:
Go right ahead! As I said i just made a minimal fix for the problem. I’m
not attached to any particular solution.

harpocrates:
Can we agree on the following?

screen shot 2019-02-28 at 11 06 33 am

Summary:

  * darken slightly the grey behind type signatures and such
  * add a border and round the corners on code blocks
  * knock the font down by one point

gbaz:
Can we keep the body text darker? I think that the contrast issue there
isn’t resolved by other things.

harpocrates:
>  Can we keep the body text darker? I think that the contrast issue there isn’t resolved by other things.

Done.

haddock-api/resources/html/Linuwial.std-theme/linuwial.css
```diff
@@ -234,7 +234,7 @@ Display the package name on top of the menu links and center both elements:
  */
 
  body, button {
-   font: 400 15px/1.4 'PT Sans',
+   font: 400 14px/1.4 'PT Sans',
```
phadej: font size on Hackage seems to be 16, 17px check the description of e.g. https://hackage.haskell.org/package/sop-core, so 14 imho too small in comparison. (12pt seems to be 16px, and 11pt 15px) e.g. https://everythingfonts.com/font/tools/units/pt-to-px - I wouldn't go under 11pt.

haddock-api/resources/html/Linuwial.std-theme/linuwial.css
```diff
@@ -408,7 +410,7 @@ blockquote {
 }
 
 .src {
-  background: #f4f4f4;
+  background: #f2f2f2;
```
phadej: if really wonder if human can distinguish `#f2f2f2` from `#f4f4f4`, so this feels odd from development POV. But they can be different, I guess.

haddock-api/resources/html/Linuwial.std-theme/linuwial.css
```diff
@@ -390,6 +390,8 @@ pre {
   margin: 1em 0 0 0;
   background-color: #f7f7f7;
   overflow: auto;
+  border: 1px solid #ddd;
```
phadej: this looks nice.

phadej: Ah, the #444 is the same as header's bottom border. Than this makes sense. :+1:

haddock-api/resources/html/Linuwial.std-theme/linuwial.css
harpocrates: Sure, it is for the project summary., but keep in mind that we are jumping from `font: 13px/1.4 sans-serif;` for the current actual docs.

haddock-api/resources/html/Linuwial.std-theme/linuwial.css
harpocrates: Heh. I spent some time toggling CSS.

haddock-api/resources/html/Linuwial.std-theme/linuwial.css
phadej: @harpocrates fair point. Would be nice to harmonise font-size at some point though.

haddock-api/resources/html/Linuwial.std-theme/linuwial.css
harpocrates: Yeah, I was trying to match the header's bottom border. Initially I had it as `gainsboro` (to match the vertical line on the left), but that was a bit too light.

phadej: I'm quite happy with current state. It's clearly an improvement for me.

NunoAlexandre: Agree, looks nice 👍

PR #1051, opened 2019-04-02
Fix #1050 by filtering out invisible AppTy arguments

### LINKED ISSUE #1050: Haddocks incorrectly display invisible kind arguments
Unfortunately, this is an issue that's exclusive to code loaded from external packages, so it's difficult to provide a minimal example. The smallest example that I can come up with (which requires `cabal new-haddock`) is located at [this repo](https://github.com/RyanGlScott/haddock-kind-bug/tree/3c994c7dd88aa9bdb67b78bc46ab91e05f618ae4). You can observe the issue by running `cabal new-haddock haddock-kind-bug2` in that repo and them opening the generated Haddocks:

haddock

Compare that to the actual [source code](https://github.com/RyanGlScott/haddock-kind-bug/blob/3c994c7dd88aa9bdb67b78bc46ab91e05f618ae4/haddock-kind-bug1/src/HaddockKindBug1.hs#L8-L9) of `T`:

```haskell
newtype T :: (forall k. k -> Type) -> (forall k. k -> Type) where
MkT :: forall (f :: forall k. k -> Type) k (a :: k). f a -> T f a
```

Variable naming issues aside (`f` is renamed to `a`, and `a` is renamed to `b`), the immediate issue is that the field of `MkT` displays the kind variable `k` as a visible argument (as in `a k b`) which is incorrect, as it should be invisible (i.e., it should be displayed as `a b`).

The culprit is these lines in `Convert`:

https://github.com/haskell/haddock/blob/fd56ac4d031963ecc137a5027d800bd8ed588264/haddock-api/src/Haddock/Convert.hs#L521-L524

This displays _any_ argument involved in an `AppTy`, regardless of whether it is intended to be visible in the source code or not. In order to make this code correct with respect to visibility, we should do something closer to what GHC does in TH reification (as in [this code from `TcSplice`](https://gitlab.haskell.org/ghc/ghc/blob/7cf5ba3dba88356571197b470556e888581212d7/compiler/typecheck/TcSplice.hs#L1822-1838)) where we first use `splitAppTys` to gather up all `AppTy` arguments and then filter out all invisible arguments.

Patch incoming.

RyanGlScott:
This makes the `synifyType` case for `AppTy` more intelligent by taking into consideration the visibilities of each `AppTy` argument and filtering out any invisible arguments, as they aren't intended to be displayed in the source code. (See #1050 for an example of what can happen if you fail to filter these out.)

Along the way, I noticed that a special `synifyType` case for `AppTy t1 (CoercionTy {})` could be consolidated with the case below it, so I took the opportunity to tidy this up.

### DIFF
--- a/haddock-api/src/Haddock/Convert.hs
+++ b/haddock-api/src/Haddock/Convert.hs
@@ -602,11 +602,14 @@ synifyType _ vs (TyConApp tc tys)
         in noLoc $ HsKindSig noExt ty' full_kind'
       | otherwise = ty'
 
-synifyType s vs (AppTy t1 (CoercionTy {})) = synifyType s vs t1
-synifyType _ vs (AppTy t1 t2) = let
-  s1 = synifyType WithinType vs t1
-  s2 = synifyType WithinType vs t2
-  in noLoc $ HsAppTy noExt s1 s2
+synifyType _ vs ty@(AppTy {}) = let
+  (ty_head, ty_args) = splitAppTys ty
+  ty_head' = synifyType WithinType vs ty_head
+  ty_args' = map (synifyType WithinType vs) $
+             filterOut isCoercionTy $
+             filterByList (map isVisibleArgFlag $ appTyArgFlags ty_head ty_args)
+                          ty_args
+  in foldl (\t1 t2 -> noLoc $ HsAppTy noExt t1 t2) ty_head' ty_args'
 synifyType s vs funty@(FunTy InvisArg _ _) = synifyForAllType s Inferred vs funty
 synifyType _ vs       (FunTy VisArg t1 t2) = let
   s1 = synifyType WithinType vs t1


RyanGlScott:
As noted in #1050, the only way I know to actually trigger the bug that this PR fixes relies on external packages, so I have absolutely no idea how I would add a regression test for this. Is there a recommended way to do this?

harpocrates:
> As noted in #1050, the only way I know to actually trigger the bug that this PR fixes relies on external packages, so I have absolutely no idea how I would add a regression test for this. Is there a recommended way to do this?

Once Hi Haddock finally lands (hopefully in the next couple of months), every type or declaration that Haddock renders will have been produced from the `Convert` module. Then, it should be pretty easy to a regression test. In the meantime, there is still non good way to do this. Sorry 😭

RyanGlScott:
> In the meantime, there is still non good way to do this. Sorry 😭

No worries! I'm fine with parking this until Hi Haddock lands, if that will make things easier in the long run.

RyanGlScott:
Aha! I've discovered a trick to distill a test case that doesn't rely on external packages. It turns out that Haddock uses `Convert` to render type signatures for top-level functions that don't have a type signature. This knowledge led me to create this test case:

```hs
{-# LANGUAGE GADTs #-}
{-# LANGUAGE PolyKinds #-}
{-# LANGUAGE RankNTypes #-}
module Bug where

import Data.Kind

newtype T :: (forall k. k -> Type) -> (forall k. k -> Type) where
  MkT :: forall (f :: forall k. k -> Type) k (a :: k). f a -> T f a

mkT = MkT
```

With `haddock-2.24.0`, you'll get the following output:

haddock

Moreover, unlike my overwrought test case [here](https://github.com/RyanGlScott/haddock-kind-bug), the test case above should be suitable for checking into Haddock's test suite.

The `f k a` in `mkT`'s type signature should be corrected to `f a` when this patch is applied. However, it's been exactly one year since I last looked at this PR, and a lot has changed since then. Should I continue to target the `ghc-head` branch, or should I resubmit this and target the `ghc-8.10` branch instead?

harpocrates:
Wow! Very clever idea for a test case. I may add some more `Convert` tests using this trick!

If possible, could you target 8.10 instead? CI is working there. I'm also not sure what the state of `ghc-head` is right now (something may be pushed there in anticipation of Marge merging a GHC MR with matching Haddock changes).

RyanGlScott:
> If possible, could you target 8.10 instead?

Sure. I'll close this PR in favor of #1157.

PR #1058, opened 2019-05-02
Redo ParseModuleHeader

### LINKED ISSUE #1055: Module header parser requires fields in hardcoded order
... which is hard to remember.

https://github.com/haskell/haddock/blob/8964666efc4d4ab9756a83d16a02115a38744408/haddock-api/src/Haddock/Interface/ParseModuleHeader.hs#L35-L43

--- 

I made a small gist https://gist.github.com/phadej/3cabc7228891d534cef9fd571f9908fc which should be good enough to substitute current very manual approach even with slightly more declarative one

How does it looks like?

phadej:
Fixes #1055

### DIFF
--- a/haddock-api/src/Haddock/Interface/ParseModuleHeader.hs
+++ b/haddock-api/src/Haddock/Interface/ParseModuleHeader.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE DeriveFunctor #-}
 {-# OPTIONS_GHC -Wwarn #-}
 -----------------------------------------------------------------------------
 -- |
@@ -11,7 +12,8 @@
 -----------------------------------------------------------------------------
 module Haddock.Interface.ParseModuleHeader (parseModuleHeader) where
 
-import Control.Monad (mplus)
+import Control.Applicative (Alternative (..))
+import Control.Monad (ap)
 import Data.Char
 import DynFlags
 import Haddock.Parser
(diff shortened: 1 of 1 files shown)


harpocrates:
Thanks for picking this up!

Perhaps a foolish question: why not use `ReadP` from `Text.ParserCombinators.ReadP`? Looks like you anyways duplicate a bunch of that functionality.

phadej:
@harpocrates I need indentation, tracking it with `ReadP` is hard.

haddock-api/src/Haddock/Interface/ParseModuleHeader.hs
```diff
@@ -73,78 +85,98 @@ parseModuleHeader dflags pkgName str0 =
...
-
-      -- comparison is case-insensitive.
-      extractPrefix :: String -> String -> Maybe String
-      extractPrefix [] s = Just s
-      extractPrefix _ [] = Nothing
-      extractPrefix (c1:cs1) (c2:cs2)
-         | toUpper c1 == toUpper c2 = extractPrefix cs1 cs2
-         | otherwise = Nothing
+
+data C = C {-# UNPACK #-} !Int Char
```
phadej: Note: `C` is not a basic `Char`, it's a `Char` carrying it's column. Thus we can `munch` all spaces but only not-spaces which are indented.

phadej:
ping @harpocrates

I clarified why e.g. `ReadP` is not good enough for this task. Is there still something?

harpocrates:
> I clarified why e.g. ReadP is not good enough for this task. Is there still something?

Nope, that's all. Sorry for not merging sooner - I've held off with hopes that some simpler solution (that doesn't involve rolling yet another parser type) would come up. Given that it hasn't and that this is a great improvement, let's merge!

phadej:
@harpocrates for the record, I looked at how [`indentation-parsec`](http://hackage.haskell.org/package/indentation-parsec) adds indentation sensitivity, and decided that in this isolated and specific case, own new small parser is not as bad (and already an improvement on adhoc `String` is a list of `Char` manipulation)

PR #1074, opened 2019-07-29
Update parsing to strip whitespace from table cells

alinab:


alexbiehl:
Thanks Alina! Can you also add a test case to the fixture test in haddock-library?

alinab:
I just added a test case for this as well updated fixture data for tables in 6 other files that were throwing up errors because of this update.

haddock-library/fixtures/examples/table-simple.parsed
```diff
@@ -3,50 +3,44 @@ DocTable
...
-                                                  (DocAppend
-                                                     (DocMonospaced (DocString "OK"))
-                                                     (DocString "       ")),
+                                                  (DocMonospaced (DocString "OK")),
                             tableCellRowspan = 1},
                          TableCell
                            {tableCellColspan = 1,
-                            tableCellContents = DocString
-                                                  " operation successful                     ",
+                            tableCellContents = DocString " operation successful",
```
alexbiehl: I wonder if we should strip leading whitespace as well?

alexbiehl:
> I just added a test case for this as well updated fixture data for tables in 6 other files that were throwing up errors because of this update.

Awesome, thank you!

haddock-library/fixtures/examples/table-simple.parsed
alinab: I will look into that as a additional task.

alinab: It involved a trivial change but I have added code that strips leading whitespace as well. Also test fixtures and one html-test reference have been updated for this.

alexbiehl:
Thanks again Alina!

PR #1076, opened 2019-07-30
Generalize string type

alinab:


alexbiehl:
Thank you for your contribution! 
Unfortunately, in the meantime since we first spoke about the issue the Text library got part of GHCs distribution, allowing us to depend on it. I would say instead of parameterizing everything we directly switch to Text.

alinab:
Ah, I see. I guess that's cleaner (and simpler) in the long-run. I'll have a look at the other recent/triaged issues.

PR #1081, opened 2019-08-23
Leading commas are not supported in the export list

parsonsmatt:
I read this part of the docs, tried it out, and got a parse error. I don't think that GHC supports this syntax.

### DIFF
--- a/doc/markup.rst
+++ b/doc/markup.rst
@@ -508,22 +508,6 @@ on, where the number of ``*``\ s indicates the level of the heading
 If you use section headings, then Haddock will generate a table of
 contents at the top of the module documentation for you.
 
-The alternative style of placing the commas at the beginning of each
-line is also supported. e.g.: ::
-
-    module Foo (
-      -- * Classes
-      , C(..)
-      -- * Types
-      -- ** A data type
-      , T
-      -- ** A record
-      , R
-      -- * Some functions
-      , f
-      , g
-      ) where
-
 When not using an export list, you may insert section headers in the
 module body. Such section headers associate with all entities
 declaried up until the next section header. For example: ::


harpocrates:
Nice catch. However, the example is only slightly off. The problem is only the very first comma - the rest can work:

```haskell
module Foo (
      -- * Classes
        C(..)
      -- * Types
      -- ** A data type
      , T
      -- ** A record
      , R
      -- * Some functions
      , f
      , g
      ) where
```

Even better would be to patch the GHC parser to accept this leading comma; fix the compiler to match the docs, rather than the other way around. 😄

parsonsmatt:
> Even better would be to patch the GHC parser to accept this leading comma

[don't get me started](https://github.com/ghc-proposals/ghc-proposals/pull/87)

harpocrates:
So now that the GHC proposal is accepted, am I to understand that the example you are removing should start working (as soon as the proposal is implemented)?

parsonsmatt:
Well, the docs are currently wrong, so we can fix them, and whenever someone implements the extra commas, this can be fixed up then. Or the docs can be wrong until some unspecified time in the future - after all, they've been wrong up until now :smile:

harpocrates:
Fixed in 89ec5657326eefbcdcead6e8a5f28f72fd21cf94.

parsonsmatt:
I don't see how that commit addresses this. As far as I can tell, the offending section is still present.

harpocrates:
The very first offending comma has been removed: https://github.com/haskell/haddock/commit/89ec5657326eefbcdcead6e8a5f28f72fd21cf94#diff-897e80371095d1cf43491db53dae292cR516. the leading comma style is sufficiently widespread that I don't think we should remove this section outright.

parsonsmatt:
Thanks! I missed that :smile:

PR #688, opened 2017-10-04
Precise Haddock: Use Avails for export resolution

alexbiehl:
Since https://github.com/ghc/ghc/commit/7e5d4a0e8e673f79b93ff63c4a4d0cd2b71e3063 GHC holds the `Avails` for each export list entry. This patch makes use of these `Avails` so it doesn't have to calculate all the reexported names by itself. 

This resolves
- #121 
- #174 
- #225 
- #344 
- #584 
- #591

PR #1082, opened 2019-08-27
Fix the ignore-exports option

lexi-lambda:
The `ignore-exports` option has been broken since #688, as mentioned in https://github.com/haskell/haddock/pull/766#issue-172505043. This PR fixes it.


harpocrates:
Thanks, merged now. 😄

PR #1084, opened 2019-09-08
QuickJump: Don't search for every keypress.

kolmodin:
When the search box gets updated, wait 150ms before searching.


harpocrates:
I see you that `haddock-api/resources/html/package-lock.json` was removed and `haddock-api/resources/html/yarn.lock` added. What is the reason for this?

Otherwise, this seems reasonable. Ideally, the UI would not be blocked while it searches, but that's a whole other problem.

kolmodin:
I switched from npm package manager to yarn package manager (which is built on top of npm). Yarn has several advantages over plain npm so the js community as a whole is moving towards yarn. 
They consume the same input files and work in a similar way but yarn is easier to use.
They use different lock files and it's dangerous to keep both since they can have locked different package versions.

harpocrates:
Right, QuickJump could stand to have nicer API 😄. But this PR is a definite improvement, so let's get it merged!

> I switched from npm package manager to yarn package manager (which is built on top of npm). Yarn has several advantages over plain npm so the js community as a whole is moving towards yarn.
They consume the same input files and work in a similar way but yarn is easier to use.
They use different lock files and it's dangerous to keep both since they can have locked different package versions.

Seems reasonable. Shouldn't the documentation [here](https://github.com/haskell/haddock/blob/ghc-8.8/haddock-api/resources/html/README.md) be updated to talk about `yarn` though?

kolmodin:
Oh, sure, I'll update the instructions. I've got a lot going on now, but I'll get to it shorly I hope.

Kleidukos:
@mergifyio rebase

Kleidukos:
@kolmodin Do you still want to work on this?

kolmodin:
No, sorry. Can someone take over, if this PR is still relevant?

Kleidukos:
I have no idea if it's still relevant, you opened it before I took office :)

lf-:
quickjump usually lags out the JS engine instantly if I apply it to esqueleto or persistent e.g., so I think this PR is still relevant, although I am wondering if it is doing something exceedingly inefficient and which could be improved instead?

Kleidukos:

PR #1085, opened 2019-09-08
Use new GHC artifact for Travis CI.

kolmodin:
Previously validate-x86_64-linux-deb8 was used which no longer exists on
gitlab. Switch to validate-x86_64-linux-deb9-hadrian.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ matrix:
     - os: linux
       addons: {apt: {packages: [ghc-ppa-tools,cabal-install-head], sources: [hvr-ghc]}}
       env:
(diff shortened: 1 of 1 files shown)


harpocrates:
Oh bother, looks like we'll have to tweak the `./configure` command to have the right linker defaults. Thanks for picking this up though!

kolmodin:
Yes, seems like this isn't enough. But it's a step in the right direction so consider merging this first.

kolmodin:
Looks like this PR was superseeded by #1089, right? Feel free to close.

harpocrates:
Hah! I never realized this was targeting `ghc-8.8`, I thought it was for `ghc-head` (where CI is still broken for the reasons this tries to fix).

You can either close this or re-target it to `ghc-head` then. Now that GHC runs the Haddock testsuite as part of its own CI, I'm less concerned about CI on the `ghc-head` branch here.

PR #1088, opened 2019-09-19
Small change in to facilitate extended typed-holes

Tritlo:
This change has no functional effect on haddock itself, it just changes one pattern to use `_ (` rather than `_(`, so that we may use `_(` as a token for extended typed-holes later (see [!1766 ](https://gitlab.haskell.org/ghc/ghc/merge_requests/1766) on GitLab)


Tritlo:
Closed in favor of https://github.com/haskell/haddock/pull/1088

PR #1090, opened 2019-09-20
Small change in to facilitate extended typed-holes

Tritlo:
This change has no functional effect on haddock itself, it just changes one pattern to use `_ (` rather than `_(`, so that we may use `_(` as a token for extended typed-holes later.

Now based on the correct branch!


harpocrates:
Sounds good, thanks for the patch!

PR #1100, opened 2019-10-17
Define `XRec` for location information and get rid of `HasSrcSpan`

sgraf812:
In https://gitlab.haskell.org/ghc/ghc/merge_requests/1970 I propose a
simpler way to encode location information into the GHC and Haddock AST
while incurring no cost for e.g. TH which doesn't need location
information.

These are just changes that have to happen in lock step.

### DIFF
--- a/haddock-api/src/Haddock/Types.hs
+++ b/haddock-api/src/Haddock/Types.hs
@@ -668,6 +668,8 @@ instance MonadIO ErrMsgGhc where
 -- * Pass sensitive types
 -----------------------------------------------------------------------------
 
+type instance XRec DocNameI f = Located (f DocNameI)
+
 type instance XForAllTy        DocNameI = NoExtField
 type instance XQualTy          DocNameI = NoExtField
 type instance XTyVar           DocNameI = NoExtField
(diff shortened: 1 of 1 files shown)


sgraf812:
We should merge this ASAP as https://gitlab.haskell.org/ghc/ghc/merge_requests/1970 has already hit `master`.

sgraf812:
@bgamari Can you merge this? I'm a little uneasy with a stale submodule commit having hit `master`...

harpocrates:
@sgraf812 In the future don't worry about getting approval for this sort of PR - especially if it passed GHC's CI.

Also, it would be better to merge the Haddock PR _before_ the GHC one, adjust the GHC PR to point to a new commit on Haddock's `master`, and only then merge the GHC MR. Otherwise, we end up with commits in GHC's history whose Haddock submodule is not on a protected branch.

sgraf812:
Yes, I would've merged earlier, but I lack Collaborator rights.

harpocrates:
That’s inconvenient - I’d give it to you, but I think only Herbert has the ability to grant that. I’ll try to be more on top of PR statuses in the future. 

An additional ping to the effect of “GHC CI passed and the GHC MR is approved” would help me out too.

PR #1105, opened 2019-10-25
Refactor for OutputableBndrId changes

sgraf812:
See https://gitlab.haskell.org/ghc/ghc/merge_requests/2010.


sgraf812:
https://gitlab.haskell.org/ghc/ghc/merge_requests/2010 is green now.

PR #1108, opened 2019-11-10
Fix #783 Don't show button if --quickjump not present

### LINKED ISSUE #783: Don't show 'QuickJump' button in header if not present

mpilgrem:
In the TypeScript `init.ts`, makes `quickJump.init()` (which creates the button) conditional on the presence of meta tag `hasQuickJump` with content `"true"`.

In the Haskell, threads the `withQuickJump` flag result through to `headHtml` to create the meta tag, accordingly.


mpilgrem:
The CI build fails because of the existing commit e6ca100973c496cd98da3385594fa9a81320f7cb. Without that commit, it builds under Windows 10 using `stack`.

mpilgrem:
I have thought of a different approach to this, which I think is better, and will provide a revised pull request. The creation of the dummy `li` element is moved from the TypeScript to the Haskell and identified, and its existence (or not) replaces the use of the meta tag.

haddock-api/src/Haddock/Backends/Xhtml.hs
```diff
@@ -142,6 +145,16 @@ headHtml docTitle themes mathjax_url =
                      ,   "}"
                      , "});" ]
 
+quickJumpButtonLi :: Bool  -- ^ With Quick Jump?
+                  -> Maybe Html
+-- The TrueScript should replace this <li> element, given its id. However, in
```
Kleidukos, line 150: I think you meant `TypeScript`?

haddock-api/src/Haddock/Backends/Xhtml.hs
```diff
@@ -142,6 +145,16 @@ headHtml docTitle themes mathjax_url =
...
+-- case it does not, the element is given content here too.
+quickJumpButtonLi True = Just $ li ! [identifier "quick-jump-button"]
+  << anchor ! [href "#"] << "Quick Jump"
+
+quickJumpButtonLi False = Nothing
+
+
```
Kleidukos, line 157: Extraneous blank line

Kleidukos (changes requested):
@mpilgrem Hi, and thank you very much for this PR! I left you a couple of comments and it should be good.
That being said, would you be able to include a screenshot of the updated interface when `--quickjump` isn't passed as an argument?

mpilgrem:
Below is a screenshot of the updated interface when `--quickjump` is not passed as a command line option:

image

mpilgrem:
@Kleidukos, I have updated the PR to reflect your two comments and posted a screenshot.

Kleidukos:
@mpilgrem Brilliant! We are currently in a merge phase to backport PRs to the ghc-9.0 branch, so while I won't be merging your PR as it is now (since it's targeting the now-defunct ghc-8.8 branch), it will be ported to the next main branch.

Thanks again :sparkles:

Kleidukos:
@mergifyio rebase

Kleidukos:
@mergifyio rebase

mpilgrem:
@Kleidukos I saw your request for a rebase and the bot's failed attempt. I have rebased manually against the `ghc-9.2` branch and tested the rebased code locally by: (1) building `haddock.exe` locally with `stack` and `nightly-2022-06-12` (GHC 9.2.3); and (2) creating Haddock documentation for a simple test project (`stack new test`) with `haddock -o docs -h app/Main.hs src/Lib.hs` and `haddock -o docs -h --quickjump app/Main.hs src/Lib.hs` and reviewing the results using the `Live Server` extension for Visual Studio Code.

I experienced some initial difficulties with rebuilding the minified JS files in `haddock-api\resources\html`. I found I had to `npm install gulp@4.0.2` (previously it was local version 3.9.1) to get the `gulp` command to work without errors.

Kleidukos:
One more reason to reshape our frontend stacK

Kleidukos (approved):

mpickering:
This PR breaks the quickjump feature on `hackage.haskell.org`.

See https://gitlab.haskell.org/ghc/ghc/-/issues/21984

We will need to revert it for the 9.4.2 release.

mpilgrem:
@mpickering, sorry! I only ever tested locally. Is there a recommended way to test that pull requests for `haddock` will not break the Haddock functionality on Hackage?

mpickering:
I tested the changes by using a local hackage server (which is quite easy to setup, it's documented in the README).

I also noticed that quick jump was not supported for candidates, so I added that (https://github.com/haskell/hackage-server/pull/1122) so perhaps another way to test in future is upload a candidate for one of your packages to check everything is in order.

PR #1112, opened 2019-11-12
Fix typo in comma-first export list

wygulmage:
As far as I can tell, putting a comma before the first item in an export list  (`module M ( , a, b, c) where`) is not legal Haskell. And Haddock chokes on headings before the opening parenthesis of the export list.


harpocrates:
Fixed in 89ec5657326eefbcdcead6e8a5f28f72fd21cf94.

PR #1114, opened 2019-11-16
Fix #1113 If no Signatures, no section of index.html

### LINKED ISSUE #1113: Empty 'Signatures' section in index.html (no backpack)
For a project that does not use `backpack`, `haddock` creates an `index.html` contents page with an empty `Signatures` section (see below). I do not know if that is the intended behaviour. As a user, I would expect that if no 'module signatures' are being used, the `Signatures` section would be absent.

**Steps to reproduce**
1. Build `haddock.exe` from source (excluding the last commit e6ca100973c496cd98da3385594fa9a81320f7cb, which causes the build to fail). `haddock --version`:
````
Haddock version 2.23.0, (c) Simon Marlow 2006
````
2. Create a simple project with `stack new testHaddock` (uses resolver `lts-14.14` (GHC 8.6.5)).

3. Run `haddock -odocs -h app/Main.hs src/Lib.hs`. This results in the following:

output

mpilgrem:
After this proposed commit, `ppSignatureTree` is like the existing `ppModuleTree`.

### DIFF
--- a/haddock-api/src/Haddock/Backends/Xhtml.hs
+++ b/haddock-api/src/Haddock/Backends/Xhtml.hs
@@ -303,6 +303,7 @@ ppPrologue pkg qual title (Just doc) =
 
 
 ppSignatureTree :: Maybe Package -> Qualification -> [ModuleTree] -> Html
+ppSignatureTree _ _ [] = mempty
 ppSignatureTree pkg qual ts =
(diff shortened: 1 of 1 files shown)


mpilgrem:
The CI build fails because of the existing commit e6ca100973c496cd98da3385594fa9a81320f7cb. Without that commit, it builds under Windows 10 using `stack`.

Issue #1071, opened 2019-07-01
could not find link destination for data family


guibou:
The following code:

```
{-# LANGUAGE TypeFamilies #-}
module NewTypeExport where

class Foo t where
  data Bar t

data Biz

instance Foo Biz where
  data Bar Biz = Bar Biz
```

Generates the following warning: 

```
Warning: NewTypeExport: could not find link destinations for:
    Type D:R:BarBiz0 Bar
```

When actually `Bar Biz` is visible, hence I don't understand the warning.

harpocrates:
Haddock doesn't do a great job in general with data families and data instances. Some thoughts:

  * `Type` is something added by Haddock (since all instances are reified from GHC internals) so we should always be able to link to it (perhaps it needs to be manually added to the renaming environment).
  *  `D:R:BarBiz0` is a mess to look at. I don't think we ever provide anchors to derived names... Perhaps the warning should start by filtering out names which satisfy `isDerivedOccName`?
  * `Bar` really should have a link destination. One day, when #897 gets merged, there will be a reasonable place to place such anchors.

guibou:
> `Type` is something added by Haddock (since all instances are reified from GHC internals) so we should always be able to link to it (perhaps it needs to be manually added to the renaming environment).

My mistake for `Type`, I had not the `-i ...../base.haddock` list of flags.

Kleidukos:
@guibou is this still reproducible?

jcmartin:
I see this issue a great deal when dealing with Generics. Specifically, when I use GeneralizedNewtypeDeriving and derive a Generic, it complains that it cannot find the link destination for the associated type family Rep instance.

Kleidukos:
Thanks for the confirmation. :)

Issue #1070, opened 2019-07-01
Do not warn about private link to private symbols


guibou:
The following:

```haskell
module Private (
  bar,
  Foo(a),
  ) where

-- | 'foo' is really great, and uses 'bar'. This reference to 'foo'
-- will triggers a warning when it is actually never used in the final
-- document, because this documentation is private.
foo = undefined

bar = undefined

class Foo t where
  a :: t
  b :: t
```

Leads to the following warnings when using `haddock Private.hs`:

```
Warning: Private: could not find link destinations for:
    b foo
```

- The link to `foo` cannot be resolved because the documentation for `foo` won't be exported. However the link to `foo` is inside an unexported piece of documentation, so this is not an issue. I then consider that this warning can be discarded.
- The link to `b` will appear in the minimal complete definition exported for `Foo`. I agree that `b` appears on the minimal complete definition (see #330), but a warning should not be generated. I also think that the documentation must explicitly say that `b` is not exported, with something such as `Minimal complete definition: a, b (private)`.

These warnings are especially annoying in continuous integration where we want to fail if haddock returns any warning. See for example the work of #927.

This was tested with ghc 8.8.1 (currently in rc status). Similar results appears with ghc 8.6.5.

guibou:
cc @ghorn

guibou:
Note that due to these bugs, it is close to impossible to build a warning free documentation because a few `base` elements triggers theses warnings.

harpocrates:
Thanks for the bug report! I agree with everything you pointed out.

guibou:
Note that an easy workaround is to:

- export the missing symbols
- Do not make any link to the hidden symbol in hidden documentation (or do not write hidden haddock). This does not work for the `class` functions.

harpocrates:
I've been thinking about this a bit more.

  * The minimal complete definition is a bit of an edge case. We can fix it with some custom code and maybe we should.

  * The link to the hidden symbol is an interesting manifestation of another problem: Haddock doesn't have a good way of keeping track of which docs will somehow get used, so it processes _every_ docstring (and during renaming, warns about there not being a link destination).

    I think we should figure out how to filter out docstrings for "private" names earlier. We shouldn't have to do any work on docstrings (including the link resolution that triggers the warning from this issue) if they aren't even used. Fun fact: although the hidden docstrings don't show up in the docs, they show up in the `*.haddock` interface files!

harpocrates:
[...]

PR #1118, opened 2019-11-26
Ghc more detailled missing symbols

### LINKED ISSUE #1072: Add a flag to ignore warning based on symbol names
In the context of making haddock fail if there is any warning #927, there are still a lot of warnings which cannot be easily fixed because they depends on code for which the user has no control or exists just because of haddock bug or behavior #854 #1071 #1070, ...

I propose to add a new command line argument to set symbols which can be ignored instead of displaying the `could not find link destinations for XXX` with the following semantics:

`--ignore_link_destination_warnings=X,Y,Z` where `X,Y,Z` is a coma separated list of symbol which won't trigger a warning if haddock could not find link destination for.

`X,Y,Z` can either be a list of fully qualified symbol (e.g. `Data.List.sort`) or unqualified names (`list`). I did not tried to implement it yet, so I don't really know what will be the most convenient.

To clarify, I'm just proposing the not display the warning if the specified identifier is in the `--ignore_link_destination_warnings` list. This won't have any other impact on the haddock process.

What are your thoughts about this? If the idea is appreciated, I volunteer to implement it.

guibou:
This pull request is related to #969, but is independent, and closes #1072.

We are using #969 to force haddock to fail when there is missing link on some ghc symbols. We have this in our CI process to ensure a bit of quality on our haddock documentation.

However, sometimes we cannot directly fix missing links. For example the link may be to an hidden module or some symbols not yet handled by haddock, such as #1070 and #1071.

This pull request changes haddock in a few ways:

- missing link now display fully qualified symbol names. Instead of displaying missing link about `Word8`, it says `Data.Word.Word8`.
- missing link are now displayed one per line. This results to a longer haddock output, but this is easier to read.
- it introduces a flag `--ignore-link-symbol` which will ignore a specified missing link. This helps reducing output garbage and will ensure that haddock won't fail once #969 is merged.

### DIFF
--- a/haddock-api/src/Haddock/Interface.hs
+++ b/haddock-api/src/Haddock/Interface.hs
@@ -110,7 +110,7 @@ processModules verbosity modules flags extIfaces = do
   let warnings = Flag_NoWarnings `notElem` flags
   dflags <- getDynFlags
   let (interfaces'', msgs) =
-         runWriter $ mapM (renameInterface dflags links warnings) interfaces'
+         runWriter $ mapM (renameInterface dflags (ignoredSymbols flags) links warnings) interfaces'
   liftIO $ mapM_ putStrLn msgs
 
   return (interfaces'', homeLinks)
--- a/haddock-api/src/Haddock/Interface/Rename.hs
+++ b/haddock-api/src/Haddock/Interface/Rename.hs
@@ -31,10 +31,11 @@ import Control.Arrow ( first )
 import Control.Monad hiding (mapM)
 import Data.List
 import qualified Data.Map as Map hiding ( Map )
+import qualified Data.Set as Set
 import Prelude hiding (mapM)
 
-renameInterface :: DynFlags -> LinkEnv -> Bool -> Interface -> ErrMsgM Interface
-renameInterface dflags renamingEnv warnings iface =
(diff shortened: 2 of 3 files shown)


guibou:
The CI issue seems irrelevant to the change I've done.

guibou:
@harpocrates hello,

I'd really like to have feedback on this change, at least know if I have a chance to get it merged or not.

guibou:
Thank you!

treeowl:
I can't seem to get this option recognized in GHC 9.0.1. Am I missing something?

guibou:
> I can't seem to get this option recognized in GHC 9.0.1. Am I missing something?

That's an haddock option.

treeowl:
@guibou, okay, but when I stick it in `{-# OPTIONS_HADDOCK #-}` I'm told it's not recognized.

guibou:
> @guibou, okay, but when I stick it in `{-# OPTIONS_HADDOCK #-}` I'm told it's not recognized.

Haa, I tested, yes you are right. `OPTIONS_HADDOCK` seems to only recognize a list of settings, but it does not works as `OPTIONS_GHC` is, that's not command line options directly passed to haddock.

guibou:
I opened #1424 to track this discussion.

PR #1120, opened 2019-12-29
Support BoxedRep

andrewthad:


harpocrates:
I believe this is for https://gitlab.haskell.org/ghc/ghc/-/merge_requests/2249.

andrewthad:
Correct. That is what it is for.

Kleidukos:

PR #1129, opened 2020-02-07
Fix build of haddock in stage1

cgibbard:
We have to use the correct version of the GHC API, but the version of the compiler itself doesn't matter.

### DIFF
--- a/haddock.cabal
+++ b/haddock.cabal
@@ -33,7 +33,7 @@ bug-reports:          https://github.com/haskell/haddock/issues
 copyright:            (c) Simon Marlow, David Waern
 category:             Documentation
 build-type:           Simple
-tested-with:          GHC==8.6.*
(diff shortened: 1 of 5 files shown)


haddock.cabal
```diff
@@ -64,7 +64,7 @@ executable haddock
 
   -- haddock typically only supports a single GHC major version
   build-depends:
-    base ^>= 4.14.0.0
+    base ^>= 4.12.0.0
```
bgamari: Wouldn't this also need to admit 4.13 since this is what we use when we build with stage1?

Ericson2314: ah good catch, the peril of just trying with what `ghc.nix` gives you!

cgibbard:
It should also certainly work with 4.14, oops!

Ericson2314:
@phadej any idea how to add released GHCs alongside the existing unreleased one here? I gather the travis yaml is a hand-modified version of some haskell-ci version's output.

PR #1141, opened 2020-03-22
Fix 8.10 Travis CI

harpocrates:


harpocrates:
Despite looking green, CI actually error-ed here. Superseded by #1143, which will actually work.

PR #1151, opened 2020-03-29
Forbid spaces in anchors (#1148)

### LINKED ISSUE #1148: Parse pragmas in @@ blocks
The following comment

```haskell
-- @
-- {-# LANGUAGE DeriveGeneric #-}
-- @
```

Gets interpreted as containing an anchor `# LANGUAGE DeriveGeneric #`, `{-<a id=" LANGUAGE DeriveGeneric "></a>-}`, so people will just see this in the rendered doc:

```
{--}
```

The current workaround is to escape the `#`:

```
-- @
-- {-\# LANGUAGE DeriveGeneric \#-}
-- @
```

The problem is that it is very easy to miss.

It would be nice to detect this situation automatically and not parse these as anchors, so we wouldn't have to escape them. Some possible criteria:

- Don't recognize anchors `#...#` containing/starting with/ending with spaces
- Don't recognize anchors `#...#` surrounded by `{-`, `-}`
- Don't recognize anchors in the middle of code blocks

Lysxia:
Closes #1148

### DIFF
--- a/haddock-library/src/Documentation/Haddock/Parser.hs
+++ b/haddock-library/src/Documentation/Haddock/Parser.hs
@@ -227,7 +227,7 @@ takeWhile1_ = mfilter (not . T.null) . takeWhile_
 -- DocAName "Hello world"
 anchor :: Parser (DocH mod a)
 anchor = DocAName . T.unpack <$>
-         disallowNewline ("#" *> takeWhile1_ (/= '#') <* "#")
+         ("#" *> takeWhile1_ (\x -> x /= '#' && not (isSpace x)) <* "#")
 
 -- | Monospaced strings.
 --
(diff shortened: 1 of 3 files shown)


harpocrates:
There appears to be another test for the wrong behaviour which is causing CI to fail. You'll have to update that test so CI passes. I propose: rename the test to `"doesn't parse a multi word anchor"` and include a link to the HTML spec in a comment right above the test. 😄

Lysxia:
Thanks for the suggestion :) Done!

harpocrates:
Thanks for your contribution! :+1:

PS. Travis CI has been flaky at updating statuses after finishing. Looking directly on the Travis CI website, this PR passed CI.

PR #1161, opened 2020-04-15
Remove some bad link warnings

### LINKED ISSUE #1070: Do not warn about private link to private symbols

harpocrates:
Fixes #1070, using the approach in https://github.com/haskell/haddock/issues/1070#issuecomment-605628782.

  * fb91c0fa56bcfe44097f65fef7fc86a700ac0fe5: Haddock has no business renaming (or doing anything) with docstrings that will never make it to any output, so we prune them early. No more warnings about anything in private docstrings!

  * 4e28b146174ca8dd5e5f8c9c8db55cb4baaa63fb: Minimal signatures no longer cause renaming warnings. Since showing the un-exported symbol is intentional, we shouldn't warn on it.

/cc @guibou


guibou:
Thank you!

Issue #1164, opened 2020-04-18
Add some predictable way of labeling section headers


garetxe:
In working in updating https://github.com/haskell/haddock/pull/1078 to the latest version of haddock in git HEAD, I noticed that https://github.com/haskell/haddock/commit/db3725fe9d1aaa424356f290654fb04213ef6c4e broke the capability of adding custom anchors for sections.

I do see the reasoning for the change (particularly in light of https://github.com/haskell/haddock/issues/1054), but I was wondering if there was any way to restore some equivalent functionality: that is, having a way of generating a predictable label that points to a section in the documentation.

Perhaps one way would be, instead of ignoring anchors in section headers, use the given anchor as the section id, instead of generating some numerical id automatically. I have not attempted to implement this, so I don't know how hard it is, but I could give it a try.

(The old possibility of adding anchors to section headers is used extensively in `gi-gtk`, for instance. Having the code generator keep track of the numbering of the sections across module boundaries is difficult and very brittle.)

garetxe:
> Perhaps one way would be, instead of ignoring anchors in section headers, use the given anchor as the section id, instead of generating some numerical id automatically. I have not attempted to implement this, so I don't know how hard it is, but I could give it a try.

I just tried this, it turned out to be fairly easy:

```diff
diff --git a/haddock-api/src/Haddock/Backends/Xhtml.hs b/haddock-api/src/Haddock/Backends/Xhtml.hs
index e3d4e8ca..4e87d0be 100644
--- a/haddock-api/src/Haddock/Backends/Xhtml.hs
+++ b/haddock-api/src/Haddock/Backends/Xhtml.hs
@@ -672,10 +672,16 @@ numberSectionHeadings = go 1
   where go :: Int -> [ExportItem DocNameI] -> [ExportItem DocNameI]
         go _ [] = []
         go n (ExportGroup lev _ doc : es)
-          = ExportGroup lev (show n) doc : go (n+1) es
+          = case collectAnchors doc of
+              [] -> ExportGroup lev (show n) doc : go (n+1) es
+              (a:_) -> ExportGroup lev a doc : go (n+1) es
         go n (other:es)
           = other : go n es
 
+        collectAnchors :: DocH (Wrap (ModuleName, OccName)) (Wrap DocName) -> [String]
+        collectAnchors (DocAppend a b) = collectAnchors a ++ collectAnchors b
+        collectAnchors (DocAName a) = [a]
+        collectAnchors _ = []
 
[...]

PR #1179, opened 2020-04-25
Add support for custom section anchors

garetxe:
As discussed in https://github.com/haskell/haddock/issues/1164

### DIFF
--- a/doc/markup.rst
+++ b/doc/markup.rst
@@ -508,6 +508,19 @@ on, where the number of ``*``\ s indicates the level of the heading
 If you use section headings, then Haddock will generate a table of
 contents at the top of the module documentation for you.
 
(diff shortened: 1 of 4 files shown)


harpocrates:
Thank you putting this together, it looks great. Before merging, do you think you could add something about this in the Haddock manual around here: https://github.com/haskell/haddock/blob/ghc-8.10/doc/markup.rst#section-headings?

garetxe:
Thanks for taking a look! I just added some documentation for this in ed6e7b2. I would be happy to change it if you think of a better way to document the change.

garetxe:
Sorry, noticed a typo, should be better now.

harpocrates:
Looks great, thank you for putting this all together. 👍

garetxe:
Great, thanks a lot!

PR #1182, opened 2020-04-26
Add dependency on exceptions because GHC.Exception was boiled down (ghc #18075)

ulysses4ever:
As per [GHC ticket #18075](https://gitlab.haskell.org/ghc/ghc/issues/18075) (corresponding [MR: 3155](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3155/)), the `Exception` module in GHC was boiled down, and functions like `Exception.gcatch` should be replaced with their `Control.Monad.Catch` equivalents from the `exceptions` package. This PR implements this replacement for Haddock as well as adding the `exceptions` dependency.

### DIFF
--- a/haddock-api/src/Haddock/GhcUtils.hs
+++ b/haddock-api/src/Haddock/GhcUtils.hs
@@ -446,11 +446,6 @@ modifySessionDynFlags f = do
   return ()
 
 
--- | A variant of 'gbracket' where the return value from the first computation
--- is not required.
-gbracket_ :: ExceptionMonad m => m a -> m b -> m c -> m c
-gbracket_ before_ after thing = gbracket before_ (const after) (const thing)
-
 -- Extract the minimal complete definition of a Name, if one exists
 minimalDef :: GhcMonad m => Name -> m (Maybe ClassMinimalDef)
 minimalDef n = do
--- a/haddock-api/haddock-api.cabal
+++ b/haddock-api/haddock-api.cabal
@@ -58,6 +58,7 @@ library
                , containers
                , deepseq
                , directory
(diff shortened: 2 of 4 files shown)


ulysses4ever:
@harpocrates I'm not sure about the CI failure…

haddock-api/src/Haddock/GhcUtils.hs
```diff
@@ -449,7 +450,7 @@ modifySessionDynFlags f = do
 -- | A variant of 'gbracket' where the return value from the first computation
 -- is not required.
 gbracket_ :: ExceptionMonad m => m a -> m b -> m c -> m c
-gbracket_ before_ after thing = gbracket before_ (const after) (const thing)
+gbracket_ = MC.bracket_
```
harpocrates, line 453: `gbracket_` was defined just to work around the annoying type of `gbracket`. Now that we have `MC.bracket_` with the right type, I think it would be better to just replace all occurences of `gbracket_` with `MC.bracket_`. Even better, I think there is only once such occurrence.

https://github.com/haskell/haddock/blob/7c34f89d801bafba3162bc091dd486cbb9e9fd5b/haddock-api/src/Haddock/Utils.hs#L407

ulysses4ever:
@harpocrates thanks for your feedback! I tried to address it.

harpocrates:
FYI: GHC HQ usually handles merging in these changes (and really most things related to the `ghc-head`), but let me know if they want this merged first.

ulysses4ever:
Not sure. @mpickering suggested I create the PR, so maybe he has an idea…

PR #1184, opened 2020-04-30
Fix test results

hsyl20:
Fix test results after https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3138

### DIFF
--- a/html-test/ref/Hash.html
+++ b/html-test/ref/Hash.html
@@ -307,7 +307,7 @@
 		      ></span
 		      > <a href="#" title="Hash"
 		      >Hash</a
-		      > <a href="#" title="Prelude"
+		      > <a href="#" title="GHC.Exts"
 		      >Float</a
 		      ></span
 		    > <a href="#" class="selflink"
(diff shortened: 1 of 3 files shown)


harpocrates:
This is a significant user change. Does this mean all links to `Float` and `Double`  will now point to `GHC.Exts` instead of `Prelude`? That's not a good thing - `GHC.Exts` is not a beginner friendly location.

hsyl20:
I have no idea why haddock now considers `GHC.Exts` as the home of `Float`, `Double`, etc. 

https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3138 shouldn't change anything but yet it does.

hsyl20:
Note that in `Hash` example we already have the same behavior for `Int`: it comes from `Data.Int` and not from `Prelude`. Haddock is free to select any topologically lowest unhidden module as home.

In `haddock-api/src/haddock/Interface.hs`:
```haskell
-- | Build a mapping which for each original name, points to the "best"
-- place to link to in the documentation.  For the definition of
-- "best", we use "the module nearest the bottom of the dependency
-- graph which exports this name", not including hidden modules.  When
-- there are multiple choices, we pick a random one.
```

`GHC.Exts` doesn't use `-not-home` and it reexports `Float` et al explicitly to expose their internals. I'll try to make it `-not-home`.

hsyl20:
Making `GHC.Exts` `not-home` avoids the issue. Sorry for the noise!

harpocrates:
Thanks for investigating!

PR #1185, opened 2020-05-01
isBootSummary now produces a result of type IsBootInterface

Ericson2314:
Companion to https://gitlab.haskell.org/ghc/ghc/-/merge_requests/1798 . It's a much smaller change ignoring whitespace.


Ericson2314:
https://gitlab.haskell.org/ghc/ghc/-/merge_requests/1798 is now approved, so if this is merged then that can be.

Issue #1194, opened 2020-05-25
Rounding down the percentage.


KommuSoft:
Right now if we compile documentation where only a small fraction is undocumented, we get `100%` of the items being covered, for example:

```
 100% (446 /448) in 'Data.Char.Combining'
  Missing documentation for:
    composeCombiningSequence (src/Data/Char/Combining.hs:4016)
    composeCombiningSequence' (src/Data/Char/Combining.hs:4019)
```

It might be better to round *down*, so show `99%`, since then it is clear if scanning through the output that there are items missing.

Furthermore, it is useful if people develop some tooling to search for percentages that are not 100%, for example `\s[0-9][0-9]?[%]` and for example let the build fail.

harpocrates:
This seems like a good idea. 100% is misleading when documentation is not actually complete.

Care to make a PR? This should be a small change:

https://github.com/haskell/haddock/blob/3c9e8081228ffcc38c760a6d9501a626071a5105/haddock-api/src/Haddock/Interface.hs#L185

PR #1195, opened 2020-05-25
use floor over round to calculate the percentage

KommuSoft:
Based on issue #1194 a change to `floor` the percentage over `round`ing.


harpocrates:
Thank you!

PR #1197, opened 2020-05-26
Add links to instance definition in hyperlinker tooltips

wz1000:
Now that https://gitlab.haskell.org/ghc/ghc/-/merge_requests/1286 has landed, we can take advantage of it in haddock by displaying links to instance definitions on hover.

demo

You can browse through all the docs of ghc+boot libs here: http://78.47.113.11/docs/html/libraries/

/ping @harpocrates since I think this will require more of a review than the usual PR to `ghc-head`

### DIFF
--- a/haddock-api/resources/html/solarized.css
+++ b/haddock-api/resources/html/solarized.css
@@ -90,8 +90,13 @@ span.annot:hover span.annottext{
 span.annot span.annottext:before{
   content: "";
   position: absolute;
-  left: -1em; top: -1em;
+  left: -1.5em; top: -1.1em;
   background: #FFFFFF00;
   z-index:-1;
-  padding: 2em 2em;
+  padding: 2em 3em;
+}
+
+:target {
+  border: 2px solid #D4D4D4;
+  background-color: #e5eecc;
 }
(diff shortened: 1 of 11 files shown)


mpickering:
Did you fix this so it took less than 42mb for one specific file?

wz1000:
@mpickering now that file "only" takes 19mb. I am working on a patch to simplify that file, but in general this problem is unavoidable on files that abuse GHCs constraint solver hard enough without significant changes to the hyperlinker. The javascript skills required for that are far beyond me.

wz1000:
The size of the html file generated for `GHC.Iface.Ext.Ast` has been reduced from 19mb to 1.77 mb now that https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3376 has been merged

mpickering:
Sounds good but there must be modules in the wild which will produce similarly big files. So either you need to fix the root cause or add an escape hatch which puts an upper bound on the size of the generated file imo.

wz1000:
Would a flag to control this be good enough?

Kleidukos:

PR #1204, opened 2020-06-10
Use functions exported from GHC.HsToCore.Docs

wz1000:
See https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3475

This seems like low hanging fruit from https://github.com/haskell/haddock/pull/906 that should be easy to adopt, and a clear win since it reduces duplication and synchronization bugs like https://gitlab.haskell.org/ghc/ghc/issues/18241.

/cc @sjakobi


wz1000:
@harpocrates thoughts on this?

sjakobi:
Thanks @wz1000! This looks like a nice improvement to me. :)

harpocrates:
Yes, thank you! Did you make sure the deleted code really matches the Haddock one (in other words - has there been any accidental divergence?). If so, LGTM. 👍

wz1000:
I had quickly gone through it earlier, but I just went through it line by line. I found one missing case, one missing bang pattern and one missing comment. I've updated the ghc MR with these now.

harpocrates:
Thanks! That’s my only concern so, all good on the Haddock side - merge with the GHC MR I guess.

PR #1220, opened 2020-07-27
Allow more characters in anchor following module reference

Lysxia:
Currently the only allowed characters in a `"module#anchor`" anchor are alphanums plus `[.#\\]`. In particular it rejects hyphens `-`, which are pretty common in anchors.

This PR refines the parsing of anchors in modules to allow more characters, and also reject spaces (because they're technically [not allowed by HTML](https://www.w3.org/TR/html51/dom.html#the-id-attribute)).


alexbiehl (approved):
Thanks! This is looking good.

PR #1227, opened 2020-08-19
Move CI to Github Actions

alexbiehl:


alexbiehl:
Oh, wow the old travis CI is very complicated. Let's not stop a running system for now :)

Issue #1231, opened 2020-08-26
Search results should be scrollable


MaxGabriel:
When you press `s` on a haddock page, you get a search box to search the package. When you have many search results, the search box overflows the page. This has two issues:

1. You can't scroll the box, so you can't see all the search results. I have confirmed additional search results do exist by seeing there are additional DOM elements with more results
2. The box taking up the entire page is a little awkward

To repro:

1. Visit https://hackage.haskell.org/package/amazonka-1.6.1
2. Press `s`
3. Search "Auth" (case sensitive—auth doesn't have many results)

I think a good fix would be to add:

```
overflow: scroll;
max-height: 80%; 
```

to the `#search-results` `div`. Any max-height will do, not specifically 80%. I've tested this works by making edits in web inspector.

PR #1235, opened 2020-09-20
Allow scrolling search results

MaxGabriel:
Closes https://github.com/haskell/haddock/issues/1231

I'm having a little trouble building Haddock locally, so I've only been able to test this change live in web inspector, not from actually building Haddock.


alexbiehl:
Cool, thanks! This is great!

PR #1238, opened 2020-10-02
Display linear/multiplicity-polymorphic arrows correctly

monoidal:
Haddock was not showing linear and multiplicity-polymorphic arrows. I think this was a shortcut taken during implementation of linear types.

### DIFF
--- a/haddock-api/src/Haddock/Backends/LaTeX.hs
+++ b/haddock-api/src/Haddock/Backends/LaTeX.hs
@@ -1073,9 +1073,13 @@ ppr_mono_ty (HsForAllTy _ tele ty) unicode
 ppr_mono_ty (HsQualTy _ ctxt ty) unicode
   = sep [ ppLContext ctxt unicode
         , ppr_mono_lty ty unicode ]
-ppr_mono_ty (HsFunTy _ _ ty1 ty2)   u
+ppr_mono_ty (HsFunTy _ mult ty1 ty2)   u
   = sep [ ppr_mono_lty ty1 u
-        , arrow u <+> ppr_mono_lty ty2 u ]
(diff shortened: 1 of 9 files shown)


haddock-api/src/Haddock/Backends/LaTeX.hs
```diff
@@ -1367,14 +1371,18 @@ quote :: LaTeX -> LaTeX
...
 dcolon unicode = text (if unicode then "∷" else "::")
 arrow  unicode = text (if unicode then "→" else "->")
+lollipop unicode = text (if unicode then "⊸" else "%1 ->")
 darrow unicode = text (if unicode then "⇒" else "=>")
 forallSymbol unicode = text (if unicode then "∀" else "forall")
 starSymbol unicode = text (if unicode then "★" else "*")
 atSign unicode = text (if unicode then "@" else "@")
 
+multAnnotation :: LaTeX
+multAnnotation = text "%"
```
goldfirere, line 1384: This looks suspect. Maybe `text "\\%"`?

alexbiehl:
```suggestion
multAnnotation = text "\\%"
```

alexbiehl:
This looks reasonable so far but could use a test.

haddock-api/src/Haddock/Backends/LaTeX.hs
monoidal: No, escaping already happens here:
```
latexMunge '%'  s = "{\\char '45}" ++ s
```

monoidal:
I added a test - this is ready for another review.

alexbiehl-gc:
Cool! Looking good!

alexbiehl (approved):

PR #1253, opened 2020-11-16
Fix after Plugins moved into HscEnv

hsyl20:
Patch for https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4401


alexbiehl-gc (approved):

alexbiehl (approved):
Oops my bad, wrong account.

PR #1260, opened 2020-12-02
Use gender neutral word in docs

vrom911:


jycr753:
So the guy who wrote the license does not decide what he is called lol makes sense

alexbiehl (approved):
Thanks! It does look reasonable.

PR #1266, opened 2020-12-08
Move to GitHub CI

alexbiehl:
Right now this is a straight copy of @chshersh's blog post at https://kodimensional.dev/github-actions.

Will adapt to haddock in successive commits.


chshersh:
@alexbiehl I'm happy to see this config used in more places 🤗 I've been using it for than a year and it works perfectly for all my projects.

I'm currently in the process of updating the blog post to bring several improvements to the config. Nothing critical, only the following:

* Move `cabal configure` before `freeze` for more reliable results
* Build dependencies in a separate step
* Update versions of all actions
* Use the `main` branch by default

I haven't updated the blog post yet, but you can find the latest config in @kowainik GitHub Actions workflow templates:

* https://github.com/kowainik/.github/blob/main/workflow-templates/ci.yml

alexbiehl:
Thanks again for your efforts. It's working great so far!

I guess it's unlikely, but do you happen to have an example in which you track GHC HEAD and head.hackage? The ghc-head branch requires us to use a) an unreleased GHC and b) head.hackage overlay to account for incompatible changes. Currently we heavily patch global cabal configuration to insert head.hackage. I want to streamline the process and it'd be great if there was prior art.

chshersh:
@alexbiehl I'm afraid, I don't have an example of this 😞 

I know, that the current `actions/setup-haskell` is going to be deprecated

* https://github.com/actions/setup-haskell/pull/56

It was temporary moved to `haskell-ci/setup` and I already see an issue there about GHC HEAD support:

* https://github.com/haskell-CI/setup/issues/6

But, AFAIU, the final official home for this action will be here:

* https://github.com/haskell/actions

I guess, you can comment under the GHC HEAD issue, tell that you would be interested as well, and ask it to move to `haskell/actions` to remember about it 🙂

PR #1288, opened 2021-01-08
Fix #1201: Add link to the readthedocs in cabal description to show on hackage.

### LINKED ISSUE #1201: Link to the manual in the description
The package description on Hackage currently contains only the cheatsheet. It would be useful if it also [linked to the manual](https://haskell-haddock.readthedocs.io/en/latest/index.html), which is otherwise a bit hard to discover if you don't already know about it.

emiflake:
This PR adds a link to the haddock binary readthedocs in the package description, which shows up on hackage.

I agree with #1201 that it's a good idea to add a link to the readthedocs, I hadn't encountered it until I saw this issue. Even though the docs are quite outdated, the base usage has stayed pretty much the same, so I think it's still fruitful to know about them.

Perhaps it could be useful to also include this link in the README. (Or only in the README, as it would be visible in both of the places.)

### DIFF
--- a/haddock.cabal
+++ b/haddock.cabal
@@ -23,6 +23,8 @@ description:
   without any documentation annotations, Haddock can generate useful documentation
   from your source code.
   .
+  Documentation for the haddock binary is available at [readthedocs](https://haskell-haddock.readthedocs.io/en/latest/).
+  .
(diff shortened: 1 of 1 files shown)


alexbiehl-gc (approved):
Thanks!

alexbiehl-gc:
Ah f it, I have to use the other account for approvals

alexbiehl (approved):

PR #1289, opened 2021-01-12
Cherry Pick PR 1286 - Fix #1206 by passing instance name as anchor

### LINKED ISSUE #1206: Can't un-collapse collapsible example on instances

soupi:
Chery picking PR #1286 as requested.

(cherry picked from commit 88106cbe72b6a444a9b335ee53089d8eac503bd7)


alexbiehl (approved):
Sweet, thanks!

alexbiehl:
Unfortunately there is no nightly GHC release so that CI always fails for this branch :( 

Wether the test suite runs through can only be seen once we bump the haddock submodule in GHC.

Issue #1291, opened 2021-01-15
CI failure for hoogle-test's Bug873


Kleidukos:
The `hoogle-test` part of the test suite fails in CI with

`hoogle-test: hoogle-test/out/Bug873: removeDirectoryRecursive:getSymbolicLinkStatus: does not exist (No such file or directory)`

This is blocking https://github.com/haskell/haddock/pull/1181

cc @alexbiehl-gc Do you have any idea why?

alexbiehl-gc:
Uh, annoying. I haven't seen that one. Have you tried to re-trigger CI? Might be just a blip.

PR #1292, opened 2021-01-15
Disable Bug873 test

Kleidukos:
While we are migrating to a new binary interface with #1181, this bug needs to be disabled.


alexbiehl-gc (approved):

Kleidukos:
closes #1291

PR #1295, opened 2021-01-21
Prepare haddock code for new Language

nomeata:
With the introduction of GHC2021, the `Languages` data type in GHC will grow. In preparation of that (and to avoid changing haddock with each new language), this change makes the code handle extensions to that data type gracefully.

### DIFF
--- a/haddock-api/src/Haddock/Backends/Xhtml.hs
+++ b/haddock-api/src/Haddock/Backends/Xhtml.hs
@@ -222,10 +222,7 @@ moduleInfo iface =
           ("Language", lg)
           ] ++ extsForm
         where
-          lg inf = case hmi_language inf of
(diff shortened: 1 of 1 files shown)


Kleidukos:
@nomeata Hi! Thank you very much for making Haddock future-proof. :) 

Since you've branched from `ghc-head`, the Github Action CI will not be triggered, and instead we will be trying to run the Travis CI one (which will most certainly fail).
Could you re-open it from the ghc-9.0 branch? Or ghc-8.10?

nomeata:
Rebased and changed base

Kleidukos (approved):
Thank you very much :sparkles:

nomeata:
Will this find its way to `ghc-head`, so that my patch against GHC can build on it?

PR #1296, opened 2021-01-21
Merge the late additions to ghc-8.10 into ghc-9.0

Kleidukos:
Last merge request to synchronise ghc-9.0 and ghc-8.10, ghc-9.0 is now the main branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ name: CI
 on:
   pull_request:
   push:
-    branches: ["ghc-8.10"]
+    branches: ["ghc-9.0"]
 
 jobs:
   cabal:
@@ -13,31 +13,36 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        cabal: ["3.2"]
+        cabal: ["3.4"]
         ghc:
-          - "8.10.1"
-          - "8.10.2"
+          - "9.0.1"
 
     steps:
     - uses: actions/checkout@v2
-      if: github.event.action == 'opened' || github.event.action == 'synchronize' || github.event.ref == 'refs/heads/ghc-8.10'
+      if: github.event.action == 'opened' || github.event.action == 'synchronize' || github.event.ref == 'refs/heads/ghc-9.0'
 
-    - uses: actions/setup-haskell@v1.1.4
+    - uses: haskell/actions/setup@v1
       id: setup-haskell-cabal
       name: Setup Haskell
       with:
         ghc-version: ${{ matrix.ghc }}
         cabal-version: ${{ matrix.cabal }}
 
+    - name: Prepare environment
+      run: echo "$HOME/.ghcup/bin" >> $GITHUB_PATH
+
+
     - name: Freeze
       run: |
         cabal freeze
-
-    - uses: actions/cache@v1
-      name: Cache ~/.cabal/store
+ 
+    - uses: actions/cache@v2
+      name: Cache ~/.cabal/store and .ghcup
       with:
-        path: ${{ steps.setup-haskell-cabal.outputs.cabal-store }}
-        key: ${{ runner.os }}-${{ matrix.ghc }}-${{ hashFiles('cabal.project.freeze') }}
+        path: |
+          ${{ steps.setup-haskell-cabal.outputs.cabal-store }}
+          .ghcup
+        key: ${{ runner.os }}
 
     - name: Build
       run: |
@@ -45,5 +50,4 @@ jobs:
         cabal build all
 
     - name: Test
-      run: |
-        cabal test all
+      run: cabal test all
--- a/haddock-api/haddock-api.cabal
+++ b/haddock-api/haddock-api.cabal
@@ -189,6 +189,7 @@ test-suite spec
                , containers
                , deepseq
                , directory
+               , exceptions
                , filepath
                , ghc-boot
                , transformers
--- a/README.md
+++ b/README.md
@@ -41,11 +41,13 @@ Note: before building `haddock`, you need to build the subprojects
 `haddock-library` and `haddock-api`, in this order!
 The `cabal v2-build` takes care of this automatically.
 
-#### Using [`cabal v2-build`][cabal v2]
+#### Using `cabal`
+
+Requires cabal `>= 3.4` and GHC `== 9.0`:
 
 ```bash
-cabal v2-build -w ghc-8.10.1
-cabal v2-test -w ghc-8.10.1 all
+cabal build all --enable-tests
+cabal test all
 ```
 
 #### Using `stack`
@@ -55,22 +57,6 @@ stack init
 stack build
 export HADDOCK_PATH="$(stack exec which haddock)"
 stack test
-```
-
-#### Using Cabal sandboxes (deprecated)
-
-```bash
-cabal sandbox init
-cabal sandbox add-source haddock-library
-cabal sandbox add-source haddock-api
-cabal sandbox add-source haddock-test
-# adjust -j to the number of cores you want to use
-cabal install -j4 --dependencies-only --enable-tests
-cabal configure --enable-tests
-cabal build -j4
-# run the test suite
-export HADDOCK_PATH="dist/build/haddock/haddock"
-cabal test
 ```
 
 ### Git Branches
--- a/cabal.project
+++ b/cabal.project
@@ -3,5 +3,12 @@ packages: ./
           ./haddock-library
           ./haddock-test
 
+with-compiler: ghc-9.0
+
+allow-newer:
+  ghc-paths:Cabal,
+  *:base,
+  *:ghc-prim
+
 -- Pinning the index-state helps to make reasonably CI deterministic
-index-state: 2020-12-08T20:13:44Z
+index-state: 2021-01-24T12:09:34Z


.github/workflows/ci.yml
```diff
@@ -13,37 +13,40 @@ jobs:
...
+    - name: Prepare environment
       run: |
-        cabal freeze
+        echo "$HOME/.ghcup/bin" >> $GITHUB_PATH
+        sudo apt-get install libtinfo-dev
 
-    - uses: actions/cache@v1
+    - name: Freeze
+      run: |
+        cabal freeze --allow-newer
```
int-index, line 38: It’s probably a better idea to update `cabal.project` accordingly so that local builds work too, see #1305 1305

.github/workflows/ci.yml
```diff
@@ -13,37 +13,40 @@ jobs:
...
       with:
         ghc-version: ${{ matrix.ghc }}
         cabal-version: ${{ matrix.cabal }}
 
-    - name: Freeze
+    - name: Prepare environment
       run: |
-        cabal freeze
+        echo "$HOME/.ghcup/bin" >> $GITHUB_PATH
+        sudo apt-get install libtinfo-dev
```
int-index, line 34: Why is this needed? In the build log I see:

```
libtinfo-dev is already the newest version (6.1-1ubuntu1.18.04).
libtinfo-dev set to manually installed.
```

.github/workflows/ci.yml
```diff
@@ -13,37 +13,40 @@ jobs:
...
+        cabal: ["3.4"]
         ghc:
-          - "8.10.1"
-          - "8.10.2"
+          - "9.0.0.20201227"
 
     steps:
     - uses: actions/checkout@v2
-      if: github.event.action == 'opened' || github.event.action == 'synchronize' || github.event.ref == 'refs/heads/ghc-8.10'
+      if: github.event.action == 'opened' || github.event.action == 'synchronize' || github.event.ref == 'refs/heads/ghc-9.0'
```
int-index, line 22: What would go wrong without this line (i.e. if it is removed entirely)? cc @chshersh

.github/workflows/ci.yml
```diff
@@ -13,37 +13,40 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        cabal: ["3.2"]
+        cabal: ["3.4"]
         ghc:
-          - "8.10.1"
-          - "8.10.2"
+          - "9.0.0.20201227"
```
int-index, line 18: Since we have a separate branch for every release, I don’t think there’s much point to specifying the ghc version in the matrix – could be moved directly into the `Setup Haskell` step

.github/workflows/ci.yml
```diff
@@ -13,37 +13,40 @@ jobs:
...
       with:
         path: ${{ steps.setup-haskell-cabal.outputs.cabal-store }}
-        key: ${{ runner.os }}-${{ matrix.ghc }}-${{ hashFiles('cabal.project.freeze') }}
+        key: ${{ runner.os }}
 
     - name: Build
       run: |
-        cabal configure --enable-tests --enable-benchmarks --test-show-details=direct
-        cabal build all
+        cabal configure --enable-tests --enable-benchmarks --test-show-details=direct --allow-newer -w ghc-9.0
```
int-index, line 48: Likewise, instead of `--allow-newer` and `-w ghc-9.0` here, it’d be more convenient for building locally if this was specified in `cabal.project` (#1305)

haddock-api/haddock-api.cabal
```diff
@@ -189,6 +189,7 @@ test-suite spec
                , containers
                , deepseq
                , directory
+               , exceptions
```
int-index, line 192: The previous lack of CI shows! 😄

hoogle-test/src/Bug873/Bug873.hs
```diff
@@ -1,5 +0,0 @@
-module Bug873 (($), ($$)) where
```
int-index, line 1: Is this test useless or something? Why is it removed? (I don’t object, just curious)

haddock-api/haddock-api.cabal
Kleidukos: You don't say…  
I'm really playing cat and mouse here :P

hoogle-test/src/Bug873/Bug873.hs
Kleidukos: Undecipherable failure, it's planned to reinstate it but we're short on contributors to properly take care of that.

.github/workflows/ci.yml
Kleidukos: Okay great :)

Kleidukos:
@harpocrates I'm having weird binary serialisation/deserialisation errors in CI. Do you have any idea about a recent commit that could have introduced it?

hoogle-test/src/Bug873/Bug873.hs
int-index: Looking into it

int-index: By “undecipherable failure” you probably mean this:

```
Failed to run Haddock on test package 'Bug873'
hoogle-test: hoogle-test/out/Bug873: removeDirectoryRecursive:getSymbolicLinkStatus: does not exist (No such file or directory)
```

Turns out this error message is produced by the use of `removeDirectoryRecursive` in `haddock-test/src/Test/Haddock.hs`, in the `runHaddock` function:

```
        unless succeeded $ removeDirectoryRecursive (outDir cfgDirConfig tpkg)
```

This code is weird. It runs when the test failed (`unless succeeded`) and assumes that the `outDir` exists (`removeDirectoryRecursive` expects the directory to exist). However, it is precisely when the test fails that we don’t get an output directory. So we would get this undecipherable failure for any other failing test, not `Bug873` specifically. The test driver needs fixing.

By commenting out this line I got a more helpful error message:

```
Failed to run Haddock on test package 'Bug873'
Haddock output is at '/dev/null'. This file can be set with `--haddock-stdout`.
```

OK, so `haddock` fails on this file, but its output is hidden. I haven’t figured out how to use the `--haddock-stdout` option though, so I hacked it around by specifying `let haddockStdOut = stdout` in `runHaddock` and got this output:

```
haddock: While creating Haddock interface for Bug310:

While creating Haddock interface for Bug310:

get Wrap: Bad h
  error, called at src/Haddock/InterfaceFile.hs:729:12 in haddock-api-2.24.0-inplace:Haddock.InterfaceFile
```

Now we’re talking! This error is apparently produced by this code in `haddock-api/src/Haddock/InterfaceFile.hs`:

```
instance Binary DocName where
  put_ bh (Documented name modu) = do
    putByte bh 0
    put_ bh name
    put_ bh modu
  put_ bh (Undocumented name) = do
    putByte bh 1
    put_ bh name

  get bh = do
    h <- getByte bh
    case h of
      0 -> do
        name <- get bh
        modu <- get bh
        return (Documented name modu)
      1 -> do
        name <- get bh
        return (Undocumented name)
      _ -> error "get DocName: Bad h"
```

This is as far as I got. And since you mention binary serialization issues in https://github.com/haskell/haddock/pull/1296#issuecomment-766374759, I suspect it might be related.

harpocrates: > OK, so haddock fails on this file, but its output is hidden. I haven’t figured out how to use the --haddock-stdout option though

I think you can pass it with `--test-options` from Cabal, or else if you `cabal run` the test suite, you can pass it to `--run-options`. The `removeDirectoryRecursive` is an issue I remember coming up against too (and I think I similarly would just change the code.

The binary serialization issue (for which I got a ping, but can't find the original comment), is probably due to the fact that the version of GHC used for CI comes with `.haddock` interface files built using the version of Haddock that was in the submodule at the time. However, this PR introduces breaking changes to that binary format. In order for this test to pass, you end up in a chicken and egg problem: you need GHC to bump the Haddock submodule so that the generated `base.haddock` matches the binary format that this PR has, but you can only bump the GHC submodule once this PR is merged.

The circular dependency problem is part of why at some point I embedded (most of) Haddock's test suite inside GHC's testsuite (at least there we would catch regressions). Over time, I've come more and more to believe that we should just try to move Haddock into GHC's tree (perhaps with the one exception being the `haddock-library` library, which is designed to build on older GHC versions, eg. so it can be used in Pandoc). Of course, the downside is that this would create lots of developer friction (in order to contribute to Haddock, you'd need to go through the whole GHC merge/build process).

Kleidukos: @harpocrates Hi! Thanks for chiming in, and sorry for the ghost ping, I had a bad case of "hands doing weird stuff with the GitHub interface".

.github/workflows/ci.yml
chshersh: Nothing should go wrong. This is just an optimization. According to docs, possible events that trigger GitHub Actions include a lot of things, and by default CI is running a lot of times for different type events:

* https://docs.github.com/en/developers/webhooks-and-events/github-event-types

My understanding of this, is this is configured that way mostly because GitHub Actions can do a lot of things, not just build the code, but also review, add labels, write messages, etc. However, I usually only want to build on code changes for pull requests and commits to the main branches, and don't want to spend GitHub Actions resources on other events.

int-index: OK, but doesn’t this `if` only apply to the checkout step? The other steps would run but without cloning the sources, and then fail

chshersh: Good point. Feel free to remove this line. I guess, nothing should become broken, and the config can be simplified.

I reread the documentation again, and it seems that the correct way to filter by events is something like this:

* https://github.com/Exheredrey/Exheredrey/blob/5a11c102af246b144ced9674dfb28ef235fd8d9f/.github/workflows/clang.yml#L5-L7

alexbiehl:
There we go, it’s green!

Kleidukos:
yep', sounds like it. :)

PR #1309, opened 2021-01-26
CI for ghc-9.0

alexbiehl-gc:


Kleidukos:
You should also include the cabal.project [from this commit](https://github.com/haskell/haddock/pull/1296/commits/a09ffde4254846e76a280555aed43a951ea428e0)

PR #1318, opened 2021-02-07
Fix search div not scrolling

MorrowM:
Issue: When using the search feature on a haddock page, if the search results overflow the search div scrolling will scroll the page rather than the search div.

Expected behavior: Scrolling inside the scroll div will cause it to scroll
Actual behavior: Scrolling inside the scroll div causes the page to scroll

Tested on Google Chrome, Firefox and Chromium Edge.


alexbiehl-gc:
Sweet, thank you!

alexbiehl-gc (approved):

PR #1324, opened 2021-02-07
Fix #1206 by passing instance name as anchor (ghc-9.0)

### LINKED ISSUE #1206: Can't un-collapse collapsible example on instances

soupi:
cherry pick of #1286 as requested.

I did not try to compile this - I hope the CI works. Let me know if more work is required on this.


alexbiehl (approved):
🚢

PR #1326, opened 2021-02-08
Release 

### LINKED ISSUE #1325: Remnants of merge conflict in `ghc-9.0` branch
I see this in the `ghc-9.0` branch, which looks to me like leftover from some merge/rebase:

https://github.com/haskell/haddock/blob/6b7fe26d9f592d7fcc8e77fbeb39f591fdff3386/haddock-api/src/Haddock/Interface/Create.hs#L48-L62

We should remove it.

alexbiehl-gc:
Also fixes #1325


alexbiehl-gc:
Let's wait with this one until we can properly build testsuite and all :(

Kleidukos:
@alexbiehl-gc Phadej has made a revision of tree-diff yesterday, could you update the hackage state index to the most recent? :)

alexbiehl:
@Kleidukos Cool! Still not workable though: We need a new release of ghc-paths first to get rid of the allow-newers or else people will not be able to build Haddock from Hackage as is.

PR #1327, opened 2021-02-09
Fix to build with HEAD

hsyl20:
Also remove some warnings

### DIFF
--- a/haddock-api/src/Haddock/Backends/Hyperlinker.hs
+++ b/haddock-api/src/Haddock/Backends/Hyperlinker.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PatternSynonyms #-}
 module Haddock.Backends.Hyperlinker
     ( ppHyperlinkedSource
     , module Haddock.Backends.Hyperlinker.Types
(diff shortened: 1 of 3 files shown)


haddock-api/src/Haddock/Backends/Hyperlinker.hs
```diff
@@ -18,7 +18,7 @@ import Data.Maybe
 import System.Directory
 import System.FilePath
 
-import GHC.Iface.Ext.Types  ( HieFile(..), HieASTs(..), HieAST(..), SourcedNodeInfo(..) )
+import GHC.Iface.Ext.Types
```
Kleidukos, line 21: Is the amount of stuff imported from Ext.Types too large to be explicitly listed?

hsyl20: I couldn't manage to import HiePath. It's just me not knowing how to import a pattern synonym ^^

Kleidukos: If I'm not mistaken: 

> Pattern synonyms can be exported and imported by prefixing the conid with the keyword pattern:

(_from https://gitlab.haskell.org/ghc/ghc/-/wikis/pattern-synonyms_)

Could you try this:

```suggestion
import GHC.Iface.Ext.Types (pattern HiePath, HieFile(..), HieASTs(..), HieAST(..), SourcedNodeInfo(..)) 
```

?

hsyl20: Ok now I know! Fixed

Kleidukos (approved):
LGTM, Godspeed!

hsyl20:
Thanks @Kleidukos !

PR #1328, opened 2021-02-09
Stable sort for (data/newtype) instances

alexbiehl:
@hsyl20 Found non determinism in Haddock output. Obviously we were comparing instance heads by Unique. The effect is probably aggravated by the Parallel Haddock work. 

https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4951#note_329145

### DIFF
--- a/haddock-api/src/Haddock/Interface/AttachInstances.hs
+++ b/haddock-api/src/Haddock/Interface/AttachInstances.hs
@@ -179,19 +179,28 @@ findFixity iface ifaceMap instIfaceMap = \name ->
 -- Collecting and sorting instances
 --------------------------------------------------------------------------------
 
+-- | Stable name for stable comparisons. GHC's `Name` uses unstable
+-- ordering based on their `Unique`'s.
+newtype SName = SName Name
+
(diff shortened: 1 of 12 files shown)


Kleidukos (approved):
LGTM

haddock-api/src/Haddock/Interface/AttachInstances.hs
```diff
@@ -194,7 +195,8 @@ instance Ord SName where
 -- For the benefit of the user (looks nice and predictable) and the
 -- tests (which prefer output to be deterministic).
 data SimpleType = SimpleType SName [SimpleType]
```
Kleidukos, line 197: Any reason why you're not using `!` and `{-# UNPACK #-}` for the constructors?

alexbiehl: No reason, it's been non-strict before so I thought I'd just leave it.

Kleidukos: Alright. :)

PR #1330, opened 2021-02-13
Clarify build instructions are for ghc-9.0 branch

toelli-msft:
Tried to build `ghc-head` using these instructions before realizing that the "GHC `== 9.0`" requirement applies specifically to the `ghc-9.0` branch. (I presume that building `ghc-head` actually requires a GHC that one has built from the latest GitLab source.)

I think this augmentation of the instructions would help clarify the situation for newcomers.


Kleidukos:
Hi Tom, thanks for the contribution! :sparkles:

PR #1332, opened 2021-02-14
Add -Wcompat-unqualified-imports flag

phadej:


phadej:
It seems to be part of https://downloads.haskell.org/~ghc/9.0.1/docs/html/users_guide/using-warnings.html#ghc-flag--Wcompat but I wonder if warnings are simply ignored...

PR #1335, opened 2021-02-19
a set of eta-reductions

KommuSoft:


Kleidukos:
Hi @KommuSoft,
while I understand the motivation behind the set of PRs you opened, I would prefer that we have a discussion about the usage of HLint in the codebase. I am not against it, but I would rather have a fixed set of lints to follow before reviewing linting PRs.

KommuSoft:
That's why these are split up per type, to discard the ones that are perhaps not wanted :).

Kleidukos:
Having done that with GHC I can tell you that this is the kind of things where we need to have a proper consensus about, because it obviously affects everyone who does a PR, and the maintainers for the years to come. :)

PR #1337, opened 2021-02-19
use (:) over (++)

KommuSoft:


Kleidukos:
See https://github.com/haskell/haddock/pull/1335#issuecomment-782268765

PR #1338, opened 2021-02-19
Remove the use of pure, liftM, and not . elem

KommuSoft:


Kleidukos:
https://github.com/haskell/haddock/pull/1335#issuecomment-782268765

PR #1341, opened 2021-02-22
html-test: Always set language

nomeata:
from ghc-9.2 on, the “default” langauge of GHC is expected to change
more wildly. To prepare for that (and unblock
https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4853), this sets the
language for all the test files to `Haskell2010`. This should insolate
this test suite against changes to the default.


Kleidukos (approved):

nomeata:
That was quick :-)

Kleidukos:
I need to port to ghc-head before you can use it in GHC, @nomeata ;)

PR #1360, opened 2021-03-07
Add support for labeled module references

Kleidukos:
Support a markdown-style way of annotating module references. For instance

```
```

reference "Module.Name#anchor" but the text displayed on the link will
be "label".


alexbiehl (approved):
Awesome!

PR #1364, opened 2021-03-17
Allow optparse-applicative 0.16 and tree-diff 0.2

felixonmars:
Builds fine and all tests pass here.


Kleidukos:
Hi @felixonmars, and thanks for this PR!

Could we have a little bit of context here? Has it prevented you from building haddock somewhere?

felixonmars:
Yes, we are currently removing the bounds to build with new libraries for Arch packaging.

Kleidukos:
Ah, doing the Lord's work. :)  
Let's go then! :rocket:

PR #1365, opened 2021-03-23
NameCache refactoring

hsyl20:
Companion MR for https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4951

### DIFF
--- a/hypsrc-test/ref/src/Classes.html
+++ b/hypsrc-test/ref/src/Classes.html
@@ -194,7 +194,7 @@
 	><span class="annottext"
 	  >bar :: Int -&gt; Int
 </span
-	  ><a href="#"
+	  ><a href="Classes.html#bar"
 	  ><span class="hs-identifier hs-var hs-var hs-var hs-var"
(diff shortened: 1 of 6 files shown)


Kleidukos:
oooh I like this kind of diff! :)

Kleidukos (approved):

alexbiehl:
Thank you sir! 🚢

alexbiehl (approved):

hypsrc-test/ref/src/Classes.html
```diff
@@ -194,7 +194,7 @@
 	><span class="annottext"
 	  >bar :: Int -&gt; Int
 </span
-	  ><a href="#"
+	  ><a href="Classes.html#bar"
```
alexbiehl, line 197: I like it, but I don't know why your patch could cause it.

hsyl20: I have no idea either. It appeared recently after a rebase

PR #1311, opened 2021-02-04
Add UnitId to Target record

fendor:
Companion to https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4937#note_328434

PR #1368, opened 2021-03-27
Add UnitId to Target record

fendor:
This way we always know to which home-unit a given target belongs to.
So far, there only exists a single home-unit at a time, but it
enables having multiple home-units at the same time.

Add the changes back from https://github.com/haskell/haddock/pull/1311 to merge https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4937 again.

CI is not green yet, afaict, I will ping when it is.

cc @Ericson2314, @Kleidukos, @hsyl20


fendor:
CI is green: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4937/pipelines

Is that enough to get merged?

Kleidukos (approved):
LGTM

PR #1400, opened 2021-05-31
Rename "levity" to "representation" in Bug873 reference output

sheaf:
I'm adjusting documentation and error messages throughout GHC to mention representation polymorphism instead of levity polymorphism where appropriate, see [GHC merge request 5860](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/5860).    
This is the only change needed to the haddock submodule to account for that.

### DIFF
--- a/hoogle-test/ref/Bug873/test.txt
+++ b/hoogle-test/ref/Bug873/test.txt
@@ -18,9 +18,9 @@ module Bug873
 --   It is also useful in higher-order situations, such as <tt><a>map</a>
 --   (<a>$</a> 0) xs</tt>, or <tt><a>zipWith</a> (<a>$</a>) fs xs</tt>.
 --   
---   Note that <tt>(<a>$</a>)</tt> is levity-polymorphic in its result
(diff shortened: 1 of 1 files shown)


Kleidukos:
Fantastic, ping me when you've finished the GHC MR :)

sheaf:
@Kleidukos Thank you. Do you know if I can refer to this commit hash in the GHC MR?

Kleidukos:
I don't know enough about how submodules, but since these are two different trees (GHC and Haddock), I would say that you cannot (or at least it won't be resolved by GitLab)

Kleidukos:
@sheaf Is this something that will have to be backported to one of the release branches (like ghc-9.2) ?

sheaf:
I don't really have much of an opinion, but I'm already finding it difficult enough to coordinate the merge across repositories.

The GHC MR is ready now; I think if this commit gets merged with this specific hash then it will be OK (unless someone else changes the haddock submodule in the meantime).

sheaf:
I think this'll have to wait until some GHC MRs come through, such as the one that corresponds to 1ceb34b.

PR #1407, opened 2021-06-09
Keep track of ordered list indexes and render them

### LINKED ISSUE #1406: Ordered lists always start at 1.
If one writes an ordered list in the documentation, then regardless of the (offset) numbers that are used in the ordered list, it starts with one.

A simple example can be found here:

https://github.com/jtdaugherty/vty/blob/fa74b0ae1ca2883e9c4fbfe55abc9253dc7b9841/src/Graphics/Vty/Attributes/Color.hs#L61-L75

here the items start with `0.`, because that is the value for the `Word8` that will be used, but the [rendered documentation](https://hackage.haskell.org/package/vty-5.33/docs/Graphics-Vty-Attributes-Color.html#t:Color) starts at `1.`.

simmsb:
Currently, when ordered lists are parsed the sequence/index numbers of the list
items are discarded at the parser.

This PR changes `DocOrderedList` to also keep track of the sequence number, and
updates the HTML and JSON (though I'm unsure on how changes to the JSON backend
should be dealt with) backends to include these. The hoogle backend remains
unchanged.

Resolves #1406


Kleidukos (approved):

Kleidukos (changes requested):
Thank you very much @simmsb. :) Could you please indicate each time the index appears in a tuple/function parameter that the variable is indeed an index? Such as replacing `(x, a)` with `(index, a)`? :)

mpickering:
Commonmark only pays attention to the first index in the list. So for example:

```
1. foo
3. foo
```

Renders as 

```
1. foo
2. foo
```

https://spec.commonmark.org/dingus/?text=1.%20asdsad%0A3.%20asdasd%0A

simmsb:
Seems so, would that be the preferred behavior?

Lysxia:
I wonder what the reason for that spec is. I once read that it reduces diffs when you remove an element in the middle of a large list, but that seems a rather niche concern. For a tool like haddock it seems preferable to give more control to the user.

Kleidukos (approved):

sol:
1. This is a breaking change and **this breaks existing documentation**!

    Specifically, it is a common pattern to introduce each list item with `1.`, say:

    ```
    1. foo
    1. bar
    1. baz
    ```

1. There is prior art for the original behavior.  Other formats, like Markdown (CommonMark, GFM, ...), MediaWiki and RST don't require you to micromanage list indices.  For a user who is familiar with those formats, **the new behavior is surprising**.

4. In software development, if I have two alternative options on how to proceed, if I can make an equally good case for both options, but one of the options is backwards compatible and the other option introduces a breaking change, then I **proceeded with the backwards compatible option**.

From my perspective it would have been better if this PR would never have landed, but that ship has sailed.  I think the best bet for damage control is to do what CommonMark and GFM do and 

> only pay~s~ attention to the first index in the list

as @mpickering pointed out.

As for the motivating example from https://github.com/haskell/haddock/issues/1406, I think this is what [definition lists](https://haskell-haddock.readthedocs.io/latest/markup.html#definition-lists) are meant for.

Issue #1404, opened 2021-06-03
Justified Paragraphs with limited line length


Topsii:
I find that Haddock produces way too long lines that are not easy to read. The fix for #405 still produces too long lines in my opinion.
<details>
  <summary>Screenshots</summary>

  ![travers_orig](https://user-images.githubusercontent.com/12656465/120662914-d222ed00-c489-11eb-84e1-54698e93ffe6.png)
  ![overview_orig](https://user-images.githubusercontent.com/12656465/120662931-d6e7a100-c489-11eb-9556-06be42101462.png)
</details>
I would prefer justified paragraphs with limited line width, which increase readability.
<details>
  <summary>Screenshots</summary>

![travers_mod](https://user-images.githubusercontent.com/12656465/120664412-31353180-c48b-11eb-969b-088565142f6c.png)
![overview_mod](https://user-images.githubusercontent.com/12656465/120664448-37c3a900-c48b-11eb-9b0a-7c20b68be1e4.png)
</details>
I have read that some people dislike so much unused white space. Further some function signatures could get too long and there may be other elements than text paragraphs such as images in the documentation. So, this could be a compromise:
<details>
  <summary>Screenshots</summary>

![travers_modsamewidth](https://user-images.githubusercontent.com/12656465/120665360-fd0e4080-c48b-11eb-91bc-1ddfa5a97040.png)
![overview_modsamewidth](https://user-images.githubusercontent.com/12656465/120665373-00093100-c48c-11eb-8752-ac091231b0ac.png)
</details>

Also for the screenshots I removed the 0.2em top and bottom padding of the cells in the instances table and I personally prefer it .<details>
  <summary>Screenshots</summary>

![mapM_orig](https://user-images.githubusercontent.com/12656465/120668684-f92fed80-c48e-11eb-96b3-a96f6b101008.png)
![mapM_modsamewidth](https://user-images.githubusercontent.com/12656465/120668690-faf9b100-c48e-11eb-864e-f4d4f66a5d98.png)
</details>
That is because most of the time I scroll instead of navigating via the table of contents.  The instances list is expanded by default and can be huge for some classes such as <code>Traversable</code>. In <code>base-4.12.0.0</code> the layout of the instances list did not take as much vertical space.

gbaz:
[...]

PR #1410, opened 2021-06-19
Limit paragraph width to 70em.

ddssff:
Studies show extremely long lines are difficult to read.   Before:

before

After:

after

### DIFF
--- a/haddock-api/resources/html/Linuwial.std-theme/linuwial.css
+++ b/haddock-api/resources/html/Linuwial.std-theme/linuwial.css
@@ -29,6 +29,7 @@ body {
 
 p {
   margin: 0.8em 0;
+  max-width: 70em;
 }
 
 ul, ol {
--- a/haddock-api/resources/html/Linuwial.std-theme/linuwial.css
+++ b/haddock-api/resources/html/Linuwial.std-theme/linuwial.css
@@ -29,7 +29,6 @@ body {
 
 p {
   margin: 0.8em 0;
-  max-width: 70em;
 }
 
 ul, ol {
--- a/haddock-api/resources/html/Ocean.theme/ocean.css
+++ b/haddock-api/resources/html/Ocean.theme/ocean.css
@@ -19,7 +19,6 @@ body {
 
 p {
   margin: 0.8em 0;
-  max-width: 70em;
 }
 
 ul, ol {
@@ -218,6 +217,7 @@ pre {
 #content {
   margin: 0 auto;
   padding: 0 2em 6em;
+  max-width: 70em;
 }
 
 #package-header {
--- a/haddock-api/resources/html/Ocean.theme/ocean.css
+++ b/haddock-api/resources/html/Ocean.theme/ocean.css
@@ -19,6 +19,7 @@ body {
 
 p {
   margin: 0.8em 0;
+  max-width: 70em;
 }
 
 ul, ol {


Kleidukos:
@ddssff Hi! Could you show in the description a "before" and an "after" screnshot? :)

ddssff:
Sure!   Before:
before

After:

after

ddssff:
Done.

Kleidukos:
@ddssff Do you have an example with function definitions? Sorry to ask all this from you but since it is a UI matter, I'd rather take an enlightened decision. :)

ddssff:
New and improved.   Apparently this problem only affects the ocean.css style, and not the linuwial.css style. It is not clear  to me how or why packages are meant to choose between these two, but most package on hackage are using ocean.css.   Before:
Screenshot_20210620_093941
After:
Screenshot_20210620_094031
It is pretty much exactly what you get if you actually resize your window.

gbaz:
>  Apparently this problem only affects the ocean.css style, and not the linuwial.css style. It is not clear to me how or why packages are meant to choose between these two, but most package on hackage are using ocean.css

The current default theme is linuwal. The user can pick a theme other than the default (such as `ocean` or `classic`) by passing the `--theme` flag on invoking haddock. Most of the docs on hackage are ocean because they were built before the linuwal theme became standard -- new docs are all built using linuwal. So improvements to ocean are certainly welcome imho, but its worth being aware that this will affect a narrow section of users.

ddssff:
It seems to me that the current default theme is ocean, that is what I get when I run cabal new-haddock on my nix system.  Maybe I'm out of date? I don't know how to get the linuwial theme to be used.

Kleidukos:
> It seems to me that the current default theme is ocean, that is what I get when I run cabal new-haddock on my nix system. Maybe I'm out of date? I don't know how to get the linuwial theme to be used.

What Haddock version are you running?

ddssff:
2.22.0

gbaz:
Well as per the Fine Changelog the theme was introduced and made default in 2.23 https://hackage.haskell.org/package/haddock-2.25.0/changelog

ddssff:
Missed it by this much!

Kleidukos:
@ddssff No worries, Haddock can be confusing :)

Topsii:
> Apparently this problem only affects the ocean.css style, and not the linuwial.css style.

No, as I understand it the linuwial style uses the following:
```
@media only screen and (min-width: 1280px) {
  #content {
    width: 63vw;
    max-width: 1450px;
  }
  ...
}
```
`1450px` admits long lines that are difficult to read. This does happen on sufficiently large monitors, see #1404.

I wonder whether large code blocks (or images/formulas?) in already existing documentation will be truncated if we change `max-width` as you do in your commit.

> The studies actually say that 70 characters, which is between 30em and 50em, may be ideal. But 70em seems quite alright.

Is there a reason why `70em`?

For reference [Pursuit the documentation tool for Purescript](https://pursuit.purescript.org/packages/purescript-prelude/5.0.1/docs/Data.Functor#v:map) gives the corresponding `content` section a width of `42em` on my screen, which matches the recommendation 'between `30em` and `50em`. So why don't we follow suit?

It seems to me that `70em` could be an acceptable compromise to sidestep the truncating issue for a lot of already existing documentation.
However it may be more appropriate to change the width of `p` elements only?
```
.doc p {
    width: 40em;
}
```
On the other hand setting the `max-width` for `#content` looks more aesthetic to me due to the alignment of paragraphs and code blocks on the right. I guess the `min-width` of code blocks could be set to `40em` too to achieve this?

Alas I am not experienced with CSS and don't know what else needs to be accounted for or can be done.

gbaz:
These comments don't belong on this PR. This was a PR to change a theme, and given that the theme is no longer default, I guess its semi-abandoned? Discussing changing a _different_ theme in a _different_ way should be done elsewhere.

ddssff:
Well that means we have to discuss where that would be so the discussion
can continue.

gbaz:
Well I would suggest the already linked ticket on the topic: #1404

PR #1428, opened 2021-10-11
Enable Haddock tests in GHC windows CI

wz1000:
Goes with https://gitlab.haskell.org/ghc/ghc/-/merge_requests/6295


wz1000:
@Kleidukos can we merge it now?

PR #1444, opened 2021-12-14
Add hint about inline link issue

malteneuss:
Quickfix for https://github.com/haskell/haddock/issues/774. Feel free to edit the phrasing.

### DIFF
--- a/doc/markup.rst
+++ b/doc/markup.rst
@@ -1126,8 +1126,8 @@ that any inline link at the beginning of a line within a multi-line comment
 isn't rendered correctly: ::
    
     {-| Some multi-line comment that has a
-    [link](https://not-rendering.org) and a
-    [reference link]: https://not-rendering.org
+    [link](https://example.com) and a
+    [reference link]: https://example.com
     -}
     
 Adding a space or a word in front of such a link can be used as a workaround.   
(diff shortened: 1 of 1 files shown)


doc/markup.rst
```diff
@@ -1121,6 +1121,17 @@ followed by the URL enclosed in regular parentheses, for example: ::
 The link text is used as a description for the URL if the output
 format supports it.
 
+Hint: There's a `known issue <https://github.com/haskell/haddock/issues/774>`_ 
+that any inline link at the beginning of a line within a multi-line comment 
+isn't rendered correctly: ::
+   
+    {-| Some multi-line comment that has a
+    [link](https://not-rendering.org) and a
+    [reference link]: https://not-rendering.org
```
Kleidukos, line 1130: The domain `not-rendering.org` should not be used, it risks being resolved and queries by automated tools, and bring unwanted traffic to the website, even if it is not registered right now.

Prefer `example.com` instead.

Kleidukos (changes requested):
Hi, thank you for the PR. I've left a change request and it should be good!

doc/markup.rst
malteneuss: Fixed. Thanks for the info.

Kleidukos (approved):
LGTM!

PR #1454, opened 2022-02-25
IDoc link has bitrotted; replaced with web.archive.org cache.

mengwong:
Or perhaps this paragraph should be reformatted to reflect that IDoc is obsolete.


Kleidukos (approved):
Wonderful, thanks!

PR #1458, opened 2022-03-02
Make links in Synopsis functional again (Fixes #1457)

### LINKED ISSUE #1457: Clicking a link in the synopsis does not navigate to the linked section
When clicking a link in the Synopsis, the Synopsis panel closes, but the browser does not navigate to the linked section. I tested this on the [`Data.Text` module page for the text-2.0 package](https://hackage.haskell.org/package/text-2.0/docs/Data-Text.html) with Chromium 97.0.4692.99 and Firefox 96.0.2.

I plan to do some more investigation on this issue and will update accordingly.

stevehartdata:
Commit e41c1cbe9f0476997eac7b4a3f17cbc6b2262faf added a call to
e.preventDefault() when handling click events that reach a toggle
element. This prevents the browser from following hyperlinks within the
Synopsis section when they are clicked by a user. This commit restores
functioning hyperlinks within the Synopsis section by removing the call
to e.preventDefault(), as it does not appear to be necessary, and
removing it increases the flexibility of the details-helper code.


Kleidukos:
Thank you very much. :)

PR #1459, opened 2022-03-03
Allow text-2.0 in haddock-library

Bodigrim:
As a Hackage trustee, I made a relevant revision https://hackage.haskell.org/package/haddock-library-1.10.0/revisions/


Kleidukos (approved):
Alright, thanks Bodigrim!

PR #1467, opened 2022-03-25
Fix CI

stevehartdata:
This should fix CI for Haddock. I've gotten it to succeed in my forked repo, so any additional changes needed to get it to succeed here should be minimal.

Let me know if there is any reason we need to keep all the constraints in `cabal.project`. If there is, then we'll probably need to go through and update them by hand to get the build to work. I'm not familiar with how haddock is integrated into GHC packaging, so I'd appreciate any feedback in that regard.

I'm also hoping to add other versions of GHC to the workflow so that we can back up our `tested-with` claims. I'll create a separate PR for those changes later.


Kleidukos:
Wonderful, thank you very much. :) 

> Let me know if there is any reason we need to keep all the constraints in cabal.project. If there is, then we'll probably need to go through and update them by hand to get the build to work. I'm not familiar with how haddock is integrated into GHC packaging, so I'd appreciate any feedback in that regard.

That can be done at a later time. :)

> I'm also hoping to add other versions of GHC to the workflow so that we can back up our tested-with claims. I'll create a separate PR for those changes later.

The whole development cycle has to be revised, so maybe we'll have to do this indeed. :)

Issue #1474, opened 2022-04-26
Linting in the codebase


philderbeast:
about the usage of HLint in the codebase. I am not against it, but I would rather have a fixed set of lints to follow before reviewing linting PRs.

_Originally posted by @Kleidukos in https://github.com/haskell/haddock/issues/1335#issuecomment-782268765_

Kleidukos:
This is excellent, thank you very much for this!

Kleidukos:
Clearly the unused language pragma are the most pressing stuff to remove from the codebase. :) 
Redundant bang patterns as well.

philderbeast:
@Kleidukos once we have the hlint action passing, we could then chip away at the hints as we please, starting as you suggest with:

```diff 
  - ignore: {name: "Avoid lambda"} # 3 hints
  - ignore: {name: "Avoid lambda using `infix`"} # 5 hints
  - ignore: {name: "Eta reduce"} # 18 hints
  - ignore: {name: "Functor law"} # 3 hints
  - ignore: {name: "Fuse concatMap/map"} # 1 hint
  - ignore: {name: "Fuse foldr/map"} # 2 hints
  - ignore: {name: "Fuse mapM/map"} # 2 hints
  - ignore: {name: "Hoist not"} # 1 hint
  - ignore: {name: "Move brackets to avoid $"} # 7 hints
  - ignore: {name: "Move guards forward"} # 1 hint
  - ignore: {name: "Move map inside list comprehension"} # 2 hints
  - ignore: {name: "Redundant $"} # 11 hints
  - ignore: {name: "Redundant <$>"} # 3 hints
- - ignore: {name: "Redundant bang pattern"} # 1 hint
  - ignore: {name: "Redundant bracket"} # 44 hints
  - ignore: {name: "Redundant id"} # 1 hint
  - ignore: {name: "Redundant if"} # 3 hints
  - ignore: {name: "Redundant lambda"} # 2 hints
  - ignore: {name: "Redundant map"} # 3 hints
  - ignore: {name: "Redundant return"} # 1 hint
  - ignore: {name: "Replace case with fromMaybe"} # 2 hints
  - ignore: {name: "Replace case with maybe"} # 4 hints
- - ignore: {name: "Unused LANGUAGE pragma"} # 21 hints
  - ignore: {name: "Use $>"} # 4 hints
  - ignore: {name: "Use ++"} # 3 hints
  - ignore: {name: "Use :"} # 3 hints
[...]

PR #1475, opened 2022-04-26
Add hlint action .hlint.yaml with ignores & CPP.

philderbeast:
Before following any hlint suggestions, set up an hlint CI action with ignores to achieve "No hints", see #1474.


philderbeast:
@Kleidukos it looks like the CI that has failed was failing before this pull request.

Kleidukos:
Ok I'll take a look, thanks

PR #1482, opened 2022-05-06
Fix hyperlinks to external items and modules

### LINKED ISSUE #1481: Hyperlinked sources as generated by the Nix toolchain prepend `../` in absolute paths
Sample bit of bad source from the Persistent `Database.Persist.TH` hyperlinked docs generated by haddock 2.25.1:

```html
<span class="annot"><span class="annottext">Exp -&gt; Stmt
</span><a href="../file:///nix/store/3bwbsy0llxxn1pixx3ll02alln56ivxy-ghc-9.0.2-doc/share/doc/ghc/html/libraries/template-haskell-2.17.0.0/src" class=""><span class="hs-identifier hs-var">NoBindS</span></a></span>
```

It's also the case in other packages such as `aeson` linking to functions in packages included with ghc. I don't know/remember if it affects links to packages sourced from hackage.

I am guessing that the following code, or code near it, is the cause of the leading `../`:

https://github.com/haskell/haddock/blob/bc0f2d6c1bfd331a9426fa9a92288444178267b1/haddock-api/src/Haddock/Backends/Hyperlinker/Renderer.hs#L265-L270

However, setting up a development environment such that I can test it with the Nix builder is running into some bugs in nixpkgs, so I can't proceed any further on figuring this out myself.

lf-:
Fixes #1481.

There were two bugs in this:
* We were assuming that we were always getting a relative path to the
  module in question, while Nix gives us file:// URLs sometimes. This
  change checks for those and stops prepending `..` to them.
* We were not linking to the file under the module. This seems
  to have been a regression introduced by #977. That is, the URLs were
  going to something like
  file:///nix/store/3bwbsy0llxxn1pixx3ll02alln56ivxy-ghc-9.0.2-doc/share/doc/ghc/html/libraries/base-4.15.1.0/src
  which does not have the appropriate HTML file or fragment for the item
  in question at the end.

There is a remaining instance of the latter bug, but not in the
hyperlinker: the source links to items reexported from other modules are
also not including the correct file name. e.g. the reexport of Entity in
esqueleto, from persistent has a source link `file:///nix/store/jwrrwxbkr2pcpf4kjgd3zyalf6jakxkz-persistent-2.13.3.5-doc/share/doc/persistent-2.13.3.5/html/src`. This bug is probably somewhere around here: https://github.com/haskell/haddock/blob/ghc-9.2/haddock-api/src/Haddock/Backends/Xhtml/Layout.hs#L289-L293

NOTE: This needs to get tested with relative-path located modules. It seems
correct for Nix based on my testing.

Testing strategy:

```
nix-shell '<nixpkgs>' --pure -A haskell.packages.ghc922.aeson
mkdir /tmp/aesonbuild && cd /tmp/aesonbuild
export out=/tmp/aesonbuild/out
genericBuild

ln -sf $HOME/co/haddock/haddock-api/resources .
./Setup haddock --with-haddock=$HOME/path/to/haddock/exec --hyperlink-source
```

### DIFF
--- a/haddock-api/src/Haddock/Backends/Hyperlinker/Renderer.hs
+++ b/haddock-api/src/Haddock/Backends/Hyperlinker/Renderer.hs
@@ -23,6 +23,7 @@ import System.FilePath.Posix ((</>))
 
 import qualified Data.Map as Map
 import qualified Data.Set as Set
+import qualified Data.List as List
 
 import Text.XHtml (Html, HtmlAttr, (!))
 import qualified Text.XHtml as Html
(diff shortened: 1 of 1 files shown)


Kleidukos (approved):

lf-:
@Kleidukos does anything need to be done on my end to get this forward-ported to ghc-9.4?

Kleidukos:
@lf- There will be no problem for ghc-9.4.

sternenseemann:
@Kleidukos This doesn't seem to be in GHC 9.4.1-rc1, since the patch we add in nixpkgs still applies?! Is that accurate or is something else wrong?

Kleidukos:
@sternenseemann From my interactions with the GHC release team, there will be a release of Haddock bundled with GHC 9.4.1

Kleidukos:
(note that this probably will also mark the move of Haddock to the GHC tree)

sternenseemann:
Maybe it's wise to follow up? Currently `utils/haddock` is a submodule pointing to 2368e9329e6600b46000abd24ec00b7e27bcae75 on GHC's `ghc-9.4` branch and arguably haddock is a pretty big thing to change after the first release candidate has been published…

Issue #1408, opened 2021-06-16
`foo'` links to foo


zliu41:
While `'foo''` and ``` `foo'' ``` both correctly link to `foo'`, ``` `foo'` ``` links to `foo`, which seems like a bug.

Kleidukos:
Alright, I can reproduce it with the following code:

```haskell
module Module where

-- | This is the `foo'` function.
foo' :: ()
foo' = ()
```

```bash
❯ haddock Module.hs 
Haddock coverage:
Warning: Package name is not available.
Warning: 'foo' is out of scope.
    If you qualify the identifier, haddock can try to link it anyway.
  50% (  1 /  2) in 'Module'
  Missing documentation for:
    Module header
```

Kleidukos:
closed by #1483

PR #1483, opened 2022-05-08
fix parsing trailing quotes in backticked identifiers (#1408)

ulysses4ever:
I assume this needs a test… It works on the example in #1408 provided by @Kleidukos 

```haskell
module Module where

-- | This is the `foo'` function.
foo' :: ()
foo' = ()
```


Kleidukos:
@ulysses4ever This is wonderful! :)
Yes a test would be most welcome.

ulysses4ever:
@Kleidukos done. Tests pass on my end.

Kleidukos (approved):

PR #1486, opened 2022-05-20
Hide synopsis from search when hidden

### LINKED ISSUE #1451: Odd in-page search behavior with recent Chromium and Chrome
I don't know if this is a Chromium bug or a Haddock bug or some weird interaction, but since I updated Chrome and Chromium recently, I've been having a problem. Specifically, if I hit Ctrl-F to find a word in some module documentation, and the word appears some ways down the page, it ends up opening up the Synopsis tab and searching there before continuing with the main page. UGH. Note: I don't have this problem in Firefox.

Lysxia:
Fix #1451


Kleidukos (approved):

parsonsmatt:
thank you so much 🙌🏻

PR #1490, opened 2022-05-25
Use visibility to decide which interfaces are included in quickjump

coot:
This is also consistent with how html index is build.  See
haskell/cabal#7669 for rationale behind this decision.


Kleidukos (approved):

Kleidukos:
:+1:

coot:
@Kleidukos thanks for merging my small PRs so quickly, I appreciate that - thanks a lot.

btw, I fixed `--with-haddock` for `cabal haddock[s]` so I can go around using version of haddock which I want with cabal.  That's part of haskell/cabal#8162 right now although I might submit it on its own.

Kleidukos:
@coot I'll review your PR on cabal when it's ready. :)

coot:
Thanks @Kleidukos

PR #1503, opened 2022-07-16
Bump hlint version to 3.4.1, the version with counts.

philderbeast:
Redo the counts available with the --default option.


Kleidukos (approved):
Thanks!

PR #1508, opened 2022-07-19
Haddock 2.27.0

Kleidukos:


haddock-api/src/Haddock.hs
```diff
@@ -382,42 +337,9 @@ render log' dflags unit_state flags sinceQual qual ifaces installedIfaces extSrc
...
-    allInstalledIfaces = concatMap (piInstalledInterfaces . snd) packages
-||||||| parent of 1b63771d (coot/multiple package (ghc-head) (#1419))
-    -- /All/ visible interfaces including external package modules.
-    allIfaces        = map toInstalledIface ifaces ++ installedIfaces
-    allVisibleIfaces = [ i | i <- allIfaces, OptHide `notElem` instOptions i ]
-=======
     -- /All/ visible interfaces including external package modules.
     allIfaces        = map toInstalledIface ifaces ++ map snd installedIfaces
     allVisibleIfaces = [ i | i <- allIfaces, OptHide `notElem` instOptions i ]
->>>>>>> 1b63771d (coot/multiple package (ghc-head) (#1419))
```
coot: Looking at this merge conflict, it seems you picked the wrong one.  It might be a sign that your branch does not contain all the commits that where already merged and by accident you'll remove some of the work that was committed to master.

coot (changes requested):

haddock-api/src/Haddock.hs
Kleidukos: Ah, thanks. It's a shitshow indeed.

PR #1525, opened 2022-08-30
Fix line wrapping in instances list

lf-:
Previously, we set the white-space to nowrap on the instances
section. This is incorrect, because it means that on particularly long
instances, the text will overflow the page and cause horizontal
scrolling.

I chose pre-wrap here because this is code. This setting will, per MDN,
ensure that:

    Sequences of white space are preserved. Lines are broken at newline
    characters, at <br>, and as necessary to fill line boxes.

before:
image

after:
image

### DIFF
--- a/haddock-api/resources/html/Linuwial.std-theme/linuwial.css
+++ b/haddock-api/resources/html/Linuwial.std-theme/linuwial.css
@@ -568,7 +568,7 @@ table.info {
 }
 
 #interface td.src {
-  white-space: nowrap;
+  white-space: pre-wrap;
 }
 
 /* @end */


lf-:
rebased in case that's the CI fix

lf-:
@Kleidukos in case you didn't get notified :)

No worries if you're busy and can't review; we patched it internally but would like to contribute the fix back.

Kleidukos:
If the fix is already used somewhere that's fantastic. I'll take care of your PR. Thanks a lot for your patience.

Kleidukos:
Thank you :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

Issue #1531, opened 2022-10-25
haddock  2.26.0 ignores both `--ignore-all-exports` and `{-# OPTIONS_HADDOCK ignore-exports #-}`


nicuveo:
Hi y'all; here's a regression in 2.26.0. In short: both the `ignore-exports` pragma and the `--ignore-all-exports` option are ignored.

### Small repro case

Given a file `Foo.hs` containing the following:
```haskell
{-# OPTIONS_HADDOCK ignore-exports #-}
module Foo (foo) where

foo :: Int
foo = 42
bar :: Int
bar = 42
```

invoking `haddock` yields the following:
```
$ haddock-9.2.4 -V
Haddock version 2.26.0, (c) Simon Marlow 2006
Ported to use the GHC API by David Waern 2006-2008

$ haddock-9.2.4 --ignore-all-exports --html Foo.hs
   0% (  0 /  2) in 'Foo'
  Missing documentation for:
    Module header
    foo (Foo.hs:5)

$ grep -q bar Foo.html && echo found || echo not found
not found
```

and the generated HTML file does not contain `bar`.

### Expected behaviour

Running the same command on the same file with haddock version `2.25.1` yields the following:

```
$ haddock-9.0 -V
Haddock version 2.25.1, (c) Simon Marlow 2006
Ported to use the GHC API by David Waern 2006-2008

$ haddock-9.0 --ignore-all-exports --html Foo.hs
   0% (  0 /  2) in 'Foo'
  Missing documentation for:
    Module header
    foo (Foo.hs:5)
    bar (Foo.hs:8)

$ grep -q bar Foo.html && echo found || echo not found
found
```

and the generated file correctly contains `bar`.

nicuveo:
I have just tried with haddock 2.27.0, installed with ghc 9.4.2, and the bug is still there, FWIW. I have cloned the repo, I'll test to see if it is also broken on `main`.

Kleidukos:
Hi @nicuveo, and thank you very much for this report.

nicuveo:
NP! I have just checked, that bug is also present on `main`. I'll have a quick look to see if I can figure it out myself.

nicuveo:
[...]

PR #1533, opened 2022-10-25
Fix `ignore-exports` again on ghc-9.4

nicuveo:
### Description

As part of the `ghc-9.0` branch, a fix was introduced to keep the `ignore-exports` option working, but this fix was mistakenly dropped in later branches. Furthermore, the test introduced in #1082 to prevent regressions was also broken at a later point. This PR reintroduces the fix, fixes the old HTML test, and introduces a new hoogle test.

This PR targets 9.4; #1532 is a similar PR for 9.2. I don't know, however, how to ensure that this fixes also lands on `main` (or `ghc-head`?), and how to prevent the tests from just being ignored or "fixed" in the future.

### Implementation

The `fullModuleContents` function iterates on all declarations of the module, but only keeps declarations for which it finds an `AvailInfo`, obtained from GHC's type-checker's output (see [`tcg_exports`](https://hackage.haskell.org/package/ghc-9.4.2/docs/GHC-Tc-Types.html#t:TcGblEnv)). However, that list only contains the `AvailInfo` information for declarations that are exported out of the module. To prevent non-exported declarations from being ignored when we actually request them, this PR takes the same approach as b93c6351afdfaf452ad9830a5f8536a353320a84 and crafts a simple `AvailInfo` for each declaration to be used in lieu of the raw `tcg_exports` list.

### Steps to verify

See #1531 for a repro case.

### Caveats

This PR bumps the dependency version of `haddock-api` to 2.27.1, as that is the version in that branch; without that, this branch doesn't compile.


alexbiehl (approved):
Thanks!

PR #1534, opened 2022-11-01
Remove outdated footnote about module re-exports

tomsmeding:
The subsection about re-exporting an entire module described that in principle, importing only some of the definitions from `B` and then re-exporting `module B` should result in those definitions being listed individually in the haddocks for the main module. It did have a footnote, however, stating that this was currently unimplemented.

This footnote is outdated at this point, because the behaviour _is_ implemented. Reproducer: running `cabal haddock` with [this project](https://paste.tomsmeding.com/Ah9DJfIf) results in the behaviour described in the main text.

### DIFF
--- a/doc/markup.rst
+++ b/doc/markup.rst
@@ -591,14 +591,7 @@ However, if the modules are not *completely* re-exported, for example:
     import C (a, b)
 
 then Haddock behaves as if the set of entities re-exported from ``B``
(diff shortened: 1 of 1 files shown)


Kleidukos (approved):
Thank you :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

PR #1535, opened 2022-11-20
Support mtl-2.3

Bodigrim:
Starting from `mtl-2.3` `liftM` is no longer re-exported from `Control.Monad` by `Control.Monad.Writer.Strict`. A simple backward-compatible fix is to replace `liftM` with `fmap`.


Kleidukos:
Thanks!

Bodigrim:
Did I target the wrong branch? Should it be `ghc-head` instead of `main`?

PR #1536, opened 2022-11-25
Track small API change in TyCon.hs 

AndreasPK:
GHC MR [9351](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/9351)


AndreasPK:
Based on the other MR's here/on gitlab and the missing commits on the github/ghc-head branch I guess the process is to open an MR against ghc-head, and those commits will be merged when this repo get's updated to a new ghc version.

So I will close this MR.

PR #1539, opened 2022-12-07
changed haddock behavior for link syntax

olafklinke:
The old link syntax `[link text](url)` does not work anymore: the slashes become italic text and the brackets and parentheses appear verbatim.


Kleidukos:

PR #1552, opened 2022-12-22
Refactor `renameInterface` to avoid space leaks

parsonsmatt:


haddock-api/src/Haddock/Interface.hs
```diff
@@ -123,11 +124,12 @@ processModules verbosity modules flags extIfaces = do
   out verbosity verbose "Renaming interfaces..."
   let warnings = Flag_NoWarnings `notElem` flags
   dflags <- getDynFlags
-  interfaces'' <- withTimingM "renameAllInterfaces" (const ()) $ do
-    let (interfaces'', msgs) =
-           runErrMsgM $ mapM (renameInterface dflags (ignoredSymbols flags) links warnings) interfaces'
```
parsonsmatt, line 128: Through some "watching verbose log output" sleuthing, I determined that generating Haddock for the work app was failing for OOM reasons on my 64GB laptop at this point.

The problem is that we're doing `runErrMsgM`, which accumulates everything into the `Writer ErrorMessages` before exiting. On the work app, this is around 4,000 modules, and some of these modules report a *huge* number of messages.

The `CPS` `Writer` *did* improve things considerably, but the overall behavior here is still bad.

haddock-api/src/Haddock/Interface.hs
```diff
@@ -123,11 +124,12 @@ processModules verbosity modules flags extIfaces = do
...
-    liftIO $ mapM_ (BSL.putStrLn . toLazyByteString) (errorMessagesToList msgs)
-    pure interfaces''
+  let !ignoredSymbolSet = Set.fromList (ignoredSymbols flags)
+  interfaces'' <-
+    withTimingM "renameAllInterfaces" (const ()) $
+      for interfaces' $ \i -> do
+        withTimingM ("renaming interface: " <+> pprModuleName (moduleName (ifaceMod i))) (const ()) $
+          renameInterface dflags ignoredSymbolSet links warnings i
```
parsonsmatt: The rewrite and timing info here is from a previous incarnation, where I was doing something like:

```haskell
for interfaces' $ \i -> 
  (i', msgs) <- runErrMsgM $ renameInterface dflags ignoredSymbols links warnings i
  mapM_ (BSL.putStrLn . toLazyByteString) (errorMessagesToList msgs)
  pure i'
```

However, a *further* refactor dropped the `ErrMsgM` entirely in favor of making the function polymorphic over `ReportErrorMessage`. An instance is defined for `Ghc` that prints each item out as it comes.

haddock-api/src/Haddock/Interface/Rename.hs
```diff
@@ -46,69 +45,60 @@ import GHC.Types.Basic ( TopLevelFlag(..) )
 --
 -- The renamed output gets written into fields in the Haddock interface record
 -- that were previously left empty.
-renameInterface :: DynFlags -> [String] -> LinkEnv -> Bool -> Interface -> ErrMsgM Interface
-renameInterface _dflags ignoredSymbols renamingEnv warnings iface =
+renameInterface :: ReportErrorMessage m => DynFlags -> Set String -> LinkEnv -> Bool -> Interface -> m Interface
```
parsonsmatt, line 48: The `[String]` was being converted into a `Set String` at each iteration. I pushed that out so it would be faster.

The code is also now generic over the `ReportErrorMessage` class. In practice, it is only instantiated at `Ghc`.

haddock-api/src/Haddock/Interface/Rename.hs
```diff
@@ -46,69 +45,60 @@ import GHC.Types.Basic ( TopLevelFlag(..) )
...
+renameInterface _dflags ignoredSymbols renamingEnv warnings iface = do
 
   -- first create the local env, where every name exported by this module
   -- is mapped to itself, and everything else comes from the global renaming
   -- env
-  let localEnv = foldl fn renamingEnv (ifaceVisibleExports iface)
-        where fn env name = Map.insert name (ifaceMod iface) env
```
parsonsmatt, line 56: `foldl` is pessimal on lists, building up a huge thunk before anything really gets done.

haddock-api/src/Haddock/Interface/Rename.hs
```diff
@@ -46,69 +45,60 @@ import GHC.Types.Basic ( TopLevelFlag(..) )
...
-
-      -- combine the missing names and filter out the built-ins, which would
-      -- otherwise always be missing.
-      missingNames = nubByName id $ filter isExternalName  -- XXX: isExternalName filters out too much
-                    (missingNames1 ++ missingNames2 ++ missingNames3
-                     ++ missingNames4 ++ missingNames5)
+  let localEnv =
+        Map.union
+          (Map.fromList (map (\name -> (name, ifaceMod iface)) (ifaceVisibleExports iface)))
+          renamingEnv
```
parsonsmatt, line 57: `Map.union` here does the same thing - we prefer things in the first `Map`, and when that doesn't work, we use the `renamingEnv` instead.

haddock-api/src/Haddock/Interface/Rename.hs
```diff
@@ -46,69 +45,60 @@ import GHC.Types.Basic ( TopLevelFlag(..) )
...
       -- @HasOpTy t1 eqTyCon_RDR t2@ (and _not_ as @HsEqTy t1 t2@), we need to
       -- manually filter out 'eqTyCon_RDR' (aka @~@).
+      filterName name
+        | not warnings
+          = False
+        | otherwise
+          = isExternalName name
+          && not (isSystemName name)
+          && not (isBuiltInSyntax name)
+          && Exact name /= eqTyCon_RDR
```
parsonsmatt, line 72: This function is ~most of the logic from the `strings` list comprehension. The main difference is that we don't do `qualifiedName` here, instead saving that for a later step.

We pass that in to `runRnFM`, which can be used to avoid collecting names in the first place.

haddock-api/src/Haddock/Interface/Rename.hs
```diff
@@ -46,69 +45,60 @@ import GHC.Types.Basic ( TopLevelFlag(..) )
...
-                | n <- missingNames
-                , not (qualifiedName n `Set.member` ignoreSet)
-                , not (isSystemName n)
-                , not (isBuiltInSyntax n)
-                , Exact n /= eqTyCon_RDR
-                ]
+      -- rename names in the exported declarations to point to things that
+      -- are closer to, or maybe even exported by, the current module.
+      (renamedIface, missingNames) =
+        runRnFM localEnv filterName $ renameInterfaceRn iface
```
parsonsmatt, line 77: `runRnFM` still uses `WriterT` under the hood, which I'm not thrilled about. However, the overall allocations locally aren't really concerning at this point.

haddock-api/src/Haddock/Interface/Rename.hs
```diff
@@ -46,69 +45,60 @@ import GHC.Types.Basic ( TopLevelFlag(..) )
...
-
-      (rnDocMap, missingNames2) = runRnFM localEnv (mapM renameDoc (ifaceDocMap iface))
-
-      (rnArgMap, missingNames3) = runRnFM localEnv (mapM (mapM renameDoc) (ifaceArgMap iface))
-
-      (renamedOrphanInstances, missingNames4)
-        = runRnFM localEnv (mapM renameDocInstance (ifaceOrphanInstances iface))
-
-      (finalModuleDoc, missingNames5)
-        = runRnFM localEnv (renameDocumentation (ifaceDoc iface))
```
parsonsmatt, line 71: This was pulled into `renameInterfaceRn`.

haddock-api/src/Haddock/Interface/Rename.hs
```diff
@@ -46,69 +45,60 @@ import GHC.Types.Basic ( TopLevelFlag(..) )
...
-
-      -- combine the missing names and filter out the built-ins, which would
-      -- otherwise always be missing.
-      missingNames = nubByName id $ filter isExternalName  -- XXX: isExternalName filters out too much
-                    (missingNames1 ++ missingNames2 ++ missingNames3
-                     ++ missingNames4 ++ missingNames5)
```
parsonsmatt, line 77: So this does a `filter isExternalName`, and then `nubByName` to only report a name once.

The refactored version tracks a `Set Name` to avoid the `nubByName` call, and we use `filterName` to filter *before recording them*.

haddock-api/src/Haddock/Interface/Rename.hs
```diff
@@ -117,60 +107,74 @@ renameInterface _dflags ignoredSymbols renamingEnv warnings iface =
...
-  pure a = RnM (const (a, id))
-  mf <*> mx = RnM $ \lkp -> let (f, out1) = unRn mf lkp
-                                (x, out2) = unRn mx lkp
-                            in (f x, out1 . out2)
+--
+-- However, storing a @'Set' 'Name'@ can result in a huge memory blowup. To
+-- avoid this problem, we also want to have a filtering function that will
+-- prevent names from being recorded if they don't pass the filter.
+newtype RnM a = RnM { unRn :: ReaderT RnMEnv (State (Set Name)) a }
+  deriving newtype (Functor, Applicative, Monad, MonadReader RnMEnv, MonadState (Set Name))
```
parsonsmatt, line 117: I suspect that this is potentially much faster, since the implementations in `mtl` have `INLINE` annotations. 

The `State (Set Name)` *could* be a CPS `WriterT` instead - we only do `modify' (Set.insert name)`, which would be equivalent to `tell (Set.singleton name)`.

cabal.project
```diff
@@ -5,8 +5,6 @@ packages: ./
           ./haddock-library
           ./haddock-test
 
-with-compiler: ghc-9.4
```
parsonsmatt, line 8: oops, probably my fault from a bad merge resolve. `with-compiler` is specified above

haddock-api/src/Haddock/Backends/Xhtml.hs
```diff
@@ -559,6 +561,7 @@ ppHtmlIndex odir doctitle _maybe_package themes
   writeUtf8File (joinPath [odir, indexHtmlFile]) (renderToString debug html)
 
   where
+    timed = withTiming logger (fromString "ppHtmlIndex") (const ())
```
parsonsmatt, line 564: woops, forgot to put a few more of these in

haddock-api/src/Haddock/Backends/Xhtml.hs
```diff
@@ -596,7 +599,7 @@ ppHtmlIndex odir doctitle _maybe_package themes
     initialChars = [ 'A'..'Z' ] ++ ":!#$%&*+./<=>?@\\^|-~" ++ "_"
 
     do_sub_index this_ix c
-      = unless (null index_part) $
+      = withTiming logger (fromString ("do_sub_index" <> [c])) (const ()) $ unless (null index_part) $
```
parsonsmatt, line 602: index generation is taking a wild amount of time

parsonsmatt (commented):
With this patch, Haddock reliably builds on the work app, even when I have a few Chrome windows open! 🎉 

I can get some before/after numbers from `persistent` if that feels valuable

Kleidukos (approved):

Kleidukos:
Excellent

Kleidukos:
@mergifyio rebase

PR #1571, opened 2023-01-10
Port the remains of Hi-Haddock

Kleidukos:


haddock-api/src/Haddock/Interface.hs
```diff
@@ -131,202 +140,113 @@ processModules verbosity modules flags extIfaces = do
...
-  -> TcM (Interface, ModuleSet)
-processModule1 verbosity flags ifaces inst_ifaces hsc_env mod_summary tc_gbl_env = do
-  out verbosity verbose "Creating interface..."
+  (_errs, modGraph) <- depanalE [] False
+
+  liftIO $ traceMarkerIO "Load started...."
+  -- Create (if necessary) and load .hi-files.
+  success <- withTimingM "load'" (const ()) $ do
+               -- TODO: Determine how to use `load'` since GHC 9.4
+               load' noIfaceCache LoadAllTargets Nothing modGraph
```
Kleidukos, line 154: @bgamari It's unclear to me how `load'` is supposed to be used since GHC 9.4. Could you perhaps help shed some light on its usage? The patch was originally written when `load'` had different arguments and I'm not sure the `Nothing` is a good value.

bgamari: Yes, `Nothing` should be fine here. This argument is intended for use by clients like GHCi which call `load` multiple times in succession.

Kleidukos: Okay!

On my end I tracked the Data.Ix exception that occurs when running the tests to something happening inside of the GHC API (but I can't really afford to insert putStrLn statements all over the GHC API.

Is this something we could pair on? I'm quite certain there must be a better way to track the origin of such an exception. :)

bgamari: For the record, this issue was due to the fact that the Haddock interface file version was not bumped. This meant that Haddock attempted to deserialize interfaces from the bootstrap compiler, which were incompatible with those expected by this branch.

bgamari:
@Kleidukos and I spent a few hours last week looking at this and nailed most of the correctness issues. What remains is to do some performance measurements to ensure that nothing has regressed horribly.

haddock-api/src/Haddock/Interface/LexParseRn.hs
```diff
@@ -176,12 +151,12 @@ rename dflags gre = rn
 -- 'GlobalReaderEnv' during 'rename') in an appropriate doc. Currently
 -- we simply monospace the identifier in most cases except when the
 -- identifier is qualified: if the identifier is qualified then we can
--- still try to guess and generate anchors across modules but the
+-- still try to guess and generate anchors accross modules but the
```
int-index, line 154: Typo?

PR #1591, opened 2023-04-05
testsuite: Print haddock stdout/stderr on failure

mpickering:
This greatly helps when there are problems running haddock tests on GHC's CI because otherwise we get the very unhelpful error message:

```
Failed to run Haddock on test package ''
```

Now instead we also get the actual problem in the log:

```
haddock: internal error: /private/var/lib/gitlab-runner/builds/7KQsmDei/0/ghc/ghc/_build/install/lib/ghc-9.7.20230404/lib/../../../share/doc/ghc-9.7.20230404/html/libraries/array-0.5.5.0-inplace/array-0.5.5.0-inplace.conf.haddock: withBinaryFile: does not exist (No such file or directory)
```

### DIFF
--- a/haddock-test/src/Test/Haddock.hs
+++ b/haddock-test/src/Test/Haddock.hs
@@ -94,7 +94,9 @@ runHaddock cfg@(Config { .. }) = do
                     , pcStdErr = Just haddockStdOut
                     }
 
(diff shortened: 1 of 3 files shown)


Kleidukos:
Let me know if you feel it is still needed, and I'll migrate it to the GHC repo. :)

PR #1594, opened 2023-05-15
Reduce memory usage

FinleyMcIlwaine:
This PR does several things to drastically reduce Haddock's memory usage. The high-level changes are described in the commit messages.

I have been testing these changes on the Agda codebase with a version of Haddock that is compiled with a GHC version that has been patched to further reduce Haddock's memory usage. With a non-patched GHC and haddock 9.6, the maximum residency is about 1.1GB with the following profile:
eventlog2html screen shot

With these patches and the patched GHC, the maximum residency is about 580MB with the following profile:
Screen Shot 2023-05-15 at 2 30 44 PM

Note the reduction in running time as well.

On the Agda codebase, there are no diffs in the generated HTML documentation.

I'll do my best to annotate this PR with comments describing the changes.


Kleidukos:
That's some serious work, thanks a bunch @FinleyMcIlwaine

PR #1561, opened 2022-12-29
Using `lucid` for HTML generation

parsonsmatt:
This PR uses `lucid` for HTML generation, mostly as a test.

# Baseline

The baseline is the `ghc-9.4` branch , not my `xhtml` builder patch.

```
!!! ppHtml: finished in 229.42 milliseconds, allocated 1087.125 megabytes

  16,064,701,512 bytes allocated in the heap
   2,265,714,656 bytes copied during GC
     191,524,344 bytes maximum residency (14 sample(s))
       5,218,824 bytes maximum slop
             493 MiB total memory in use (0 MB lost due to fragmentation)

                                     Tot time (elapsed)  Avg pause  Max pause
  Gen  0      3820 colls,     0 par    1.220s   1.222s     0.0003s    0.0035s
  Gen  1        14 colls,     0 par    0.703s   0.704s     0.0503s    0.1284s

  TASKS: 5 (1 bound, 4 peak workers (4 total), using -N1)

  SPARKS: 0 (0 converted, 0 overflowed, 0 dud, 0 GC'd, 0 fizzled)

  INIT    time    0.001s  (  0.000s elapsed)
  MUT     time    4.838s  (  5.365s elapsed)
  GC      time    1.923s  (  1.925s elapsed)
  EXIT    time    0.001s  (  0.009s elapsed)
  Total   time    6.762s  (  7.300s elapsed)

  Alloc rate    3,320,501,272 bytes per MUT second

  Productivity  71.5% of total user, 73.5% of total elapsed
```

# Lucid

```
!!! ppHtml: finished in 445.94 milliseconds, allocated 2262.033 megabytes

  17,296,688,776 bytes allocated in the heap
   2,251,719,184 bytes copied during GC
     191,537,256 bytes maximum residency (14 sample(s))
       5,267,352 bytes maximum slop
             493 MiB total memory in use (0 MB lost due to fragmentation)
                                     Tot time (elapsed)  Avg pause  Max pause
  Gen  0      3969 colls,     0 par    1.232s   1.234s     0.0003s    0.0034s
  Gen  1        14 colls,     0 par    0.711s   0.711s     0.0508s    0.1275s

  TASKS: 5 (1 bound, 4 peak workers (4 total), using -N1)

  SPARKS: 0 (0 converted, 0 overflowed, 0 dud, 0 GC'd, 0 fizzled)

  INIT    time    0.001s  (  0.000s elapsed)
  MUT     time    5.068s  (  5.588s elapsed)
[...]

PR #1598, opened 2023-06-06
Port HTML backend to Lucid

AriFordsham:
Haddock generates HTML using xhtml, which is both [unmaintained](https://github.com/haskell/xhtml/issues/16#issuecomment-1344544478) and [slow](https://www.parsonsmatt.org/2022/12/21/haddock_performance.html).

This ports to Lucid, which is a maintained, performance-oriented replacement. 

Lucid (transitively) contains one out-of-boot dependency, blaze-markup, but will allow removal of xhtml from boot (assuming Haddock is the only user of boot xhtml?). This increases the total number of boot libraries by one, but allows removing an unmaintained one.

I assume there's some process to go through to get the change in boot libraries to happen.

According to the test suite it seems to be correct, but may be missing further testing, and has not been benchmarked.

Still to be done:
- [x] Carefully compare new and old outputs to check correctness
- [ ] Benchmark to ascertain performance benefit
- [ ] Clean up PR
- [ ] Work out getting neccesary ghc-boot changes approved


AriFordsham:
As far as I can see, the output now of the golden tests now matches the originals.

AriFordsham:
Didn't realise @parsonsmatt was already on to this: https://github.com/haskell/haddock/pull/1561. Should this be closed?

FinleyMcIlwaine:
@AriFordsham I don't think this patch is worth pursuing further, for the time being. Adding lucid as a boot library, and transitively blaze-markup, would require a disproportionate amount of maintenance overhead for the potential improvements (based on those reported in the [existing patch](https://github.com/haskell/haddock/pull/1561#issue-1513292314)). A simpler and more incremental win appears to be optimizing xhtml directly, and I think that's the way to go for now.

Kleidukos:
I'd like to approach this again after GHC 9.8 is release, which will mark the return of Haddock in the GHC tree. Indeed while the question of dependencies is critical, `lucid2` is dropping the non-boot dependency on `blaze-builder` https://github.com/chrisdone/lucid/issues/143.

This will allow us to add lucid2 as a dependency and remove `xhtml` from GHC's boot dependencies.

Kleidukos:
@AriFordsham Thanks for starting this btw, I'm closing this PR but the effort will continue on the Haskell GitLab once Haddock moves there. :)

AriFordsham:
@Kleidukos bumping this PR, as Haddock is now in GHC and Lucid's `blaze-builder` dependency is gone. I unfortunately don't have the time for this anymore, but I wanted to bring it back to your attention.

PR #1600, opened 2023-06-15
Scrap SYB traversals, add GHC progress reporting during load

FinleyMcIlwaine:
The SYB traversals were a huge contributor to memory usage for Haddock. This MR includes the work done by @wz1000 [here](https://gitlab.haskell.org/ghc/haddock/-/commit/8030115d873d5ee69560f562fe6b6d83fc290eba) to get rid of them. This brings maximum residency down about 10% in the testing I've done.

Also, with the current implementation of hi-haddock, the load operation can create an awkward silence if recompilation occurs as no progress reporting is being done by GHC. This MR changes the `load'` call to use the default `batchMsg` progress reporting and changes the default GHC verbosity level to 1 so that it is obvious when recompilation is occurring.


Kleidukos (approved):
LGTM, thanks a lot for this work.

Issue #1602, opened 2023-06-27
MathJax 3 support


voidbert:
### The proposal

Currently, haddock uses MathJax 2.7. Due to new features, new packages, etc., I believe some users would be interested in the option of using MathJax 3.

I opened this issue to see if someone would be interested in this feature, and to discuss how it could be implemented.

### Inherent problem and fix

MathJax 3 introduces some [API breaking changes](https://docs.mathjax.org/en/v3.2-latest/upgrading/v2.html). More than breaking configuration files (which could easily be fixed from haddock's end), this breaks some existing formulas. Therefore, it may be a good idea to let users choose between the older and newer version.

I am aware that a custom MathJax URL can be provided (`--mathjax [url]`). However, due to the different configuration files needed, I suggest introducing a new command-line option (like `--mathjax3 [url]`).

### In the future

If people are interested in this feature, and it ends up being implemented, a way to integrate it with cabal would be needed.

PR #1603, opened 2023-06-28
MathJax 3 support

voidbert:
Addressed issue #1602.

A new option, `--mathjax-version=VERSION` was added. This allows the user to specify what major MathJax version they want to use. This information can be used to choose the correct file (from the Cloudflare defaults), and to initialize the library correctly (2 and 3 have different configuration methods), even if the user provides their own files.

This is my first PR in this project, so please tell me if my code may break anything and how it can be improved. The HTML tests don't pass, but I tested that that was already the case in the commit before mine.

### DIFF
--- a/doc/invoking.rst
+++ b/doc/invoking.rst
@@ -130,9 +130,15 @@ The following options are available:
 
 .. option:: --mathjax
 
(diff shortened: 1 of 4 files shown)


Kleidukos:

PR #1635, opened 2024-04-17
Add grey background to inline code

jasagredo:
In line with what other programming languages show, monospaced text will now have a slightly grey background:

image

as without it, specially non-linked monospaced words were very difficult to spot:

image

This follows the same style as [Rust does](https://doc.rust-lang.org/std/), which in my opinion looks very readable 

image

-------------

This is probably a matter of preference and esthetics so I think it would be perfectly understandable if the PR is not accepted.


Kleidukos:

jasagredo:
Ported https://gitlab.haskell.org/ghc/ghc/-/merge_requests/12672 @Kleidukos

